PR #220, opened 2022-07-25
refactor: inline the increment* functions to cut down dependencies

felipesere:
`cargo-edit` brought a lot of dependencies for 3 fairly easy functions.
Their inlined versions are still easy to read and extend if necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1861,7 +1861,6 @@ dependencies = [
 name = "next_version"
 version = "0.1.6"
 dependencies = [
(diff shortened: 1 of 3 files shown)


MarcoIeni:
This looks nice, thanks! If you don't mind, can you add a comment pointing to the cargo-edit original source code?
something like `copy pasted from ...`  just to recognize their work :)

MarcoIeni:
I merged by mistake :joy: Can you do another pr?

PR #349, opened 2022-11-18
Add tests for shell completions

### LINKED ISSUE #330: Test code completions
## Motivations

When we updated clap to version 4, the `generate-completions` subcommand broke. See https://github.com/MarcoIeni/release-plz/pull/328. We should add a test.

## Solution

Use [assert_cmd](https://crates.io/crates/assert_cmd) to add a test where we run this subcommand and we check that the outcome is successful.

## Alternatives

If you want to use an alternative to assert_cmd, leave a comment.

orhun:
An attempt to fix #330

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,6 +64,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_cmd"
+version = "2.0.6"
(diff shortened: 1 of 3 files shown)


MarcoIeni:
Thanks ❤️

PR #397, opened 2022-12-27
feat(release): create GitHub release

gautamprikshit1:
Added `gh-release` argument to release sub-command.


crates/release_plz/src/args/release.rs
```diff
@@ -24,6 +24,9 @@ pub struct Release {
     /// Perform all checks without uploading.
     #[clap(long)]
     pub dry_run: bool,
+    ///Publish github release.
```
MarcoIeni, line 27:
```suggestion
    /// Publish GitHub release.
```

crates/release_plz_core/src/release.rs
```diff
@@ -26,6 +26,8 @@ pub struct ReleaseRequest {
     pub token: Option<String>,
     /// Perform all checks without uploading.
     pub dry_run: bool,
+    ///Publishes github release.
```
MarcoIeni, line 29:
```suggestion
    /// Publishes GitHub release.
```

crates/release_plz_core/src/release.rs
```diff
@@ -106,6 +108,9 @@ fn release_package(
     if input.dry_run {
         args.push("--dry-run");
     }
+    if input.gh_release {
+        args.push("--gh-release");
+    }
```
MarcoIeni, line 113: These args are passed to `cargo`. `cargo` doesn't have the `--gh-release` flag. So this won't work I think. 🤔

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +136,24 @@ fn release_package(
...
+    if input.gh_release {
+        publish_release(package);
+    }
+
     Ok(())
 }
 
+async fn publish_release(package: &Package) {
+    let page = octocrab::Octocrab::default()
+        .repos(package.authors.join(", "), package.clone().name)
```
MarcoIeni, line 148: This should be the info contained in the GitHub url of the repository.
For example, for this repository it's `.repos("MarcoIeni", "release-plz")`.
This info doesn't always correspond to the cargo.toml file.

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +136,24 @@ fn release_package(
...
+        .releases()
+        .create(&package.version.to_string())
```
MarcoIeni, line 150: This should be the tag name. It's different from the package version.
You can pass the `git_tag` variable above.

MarcoIeni (commented):
Thanks for working on this!
Here are my comments :)

crates/release_plz_core/src/release.rs
MarcoIeni: Why are you passing `-- --gh-release` to `cargo publish`?
This command invokes cargo to publish to crates.io, it doesn't invoke `release-plz`.
So you don't need to edit these args.

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +137,26 @@ fn release_package(
...
+    if input.gh_release {
+        publish_release(git_tag(&package.name, &package.version.to_string()));
+    }
+
     Ok(())
 }
 
+async fn publish_release(git_tag: String) {
+    let owner = RepoUrl::new("https://www.github.com").unwrap().owner;
+    let repo = RepoUrl::new("https://www.github.com").unwrap().name;
```
MarcoIeni, line 149: The `unwrap` here will panic, because the string `"https://www.github.com"` doesn't have owner and name.
We need to find owner and name similarly to how we do it for the update command [here](https://github.com/MarcoIeni/release-plz/blob/36616bbb470e429c4686c41234f9a6b4b7dee936/crates/release_plz/src/args/update.rs#L86).
So we need to add `repo_url` to the input struct of the release [command](https://github.com/MarcoIeni/release-plz/blob/36616bbb470e429c4686c41234f9a6b4b7dee936/crates/release_plz/src/args/release.rs#L9)

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +137,26 @@ fn release_package(
...
+    let page = octocrab::Octocrab::default()
```
MarcoIeni, line 150: we need to construct the GitHub client by using [this](https://github.com/MarcoIeni/release-plz/blob/36616bbb470e429c4686c41234f9a6b4b7dee936/crates/release_plz_core/src/github_client.rs#L15) method.
In this way, we can pass the personal access token.
Without the token, we can't create the release.
You need to read the github personal access token in input, similarly to how we do for the release-pr command [here](https://github.com/MarcoIeni/release-plz/blob/36616bbb470e429c4686c41234f9a6b4b7dee936/crates/release_plz/src/args/release_pr.rs#L17).

MarcoIeni:
- [ ] Also, we need to make sure that we are not in a GitHub repository, before creating the release. You can use [this](https://github.com/MarcoIeni/release-plz/blob/36616bbb470e429c4686c41234f9a6b4b7dee936/crates/release_plz_core/src/repo_url.rs#L35) method.

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +140,41 @@ fn release_package(
         info!("published {}", package.name);
     }
 
+    if input.gh_release {
+        publish_release(git_tag(&package.name, &package.version.to_string()), input);
```
MarcoIeni, line 144:
```suggestion
        publish_release(git_tag(&package.name, &package.version.to_string()), input).await;
```
You need to `await` this function because it's async.
Since this is the first await of the `release_package` function, you need to change the signature of the `release_package` function and add the `async` keyword, too.

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +140,41 @@ fn release_package(
...
+    }
```
MarcoIeni, line 145: can you move this block inside the `else` above?
We don't want to do a GitHub release if the user passed the `dry_run` flag.

This is how it should look:

```rust
    } else {
        wait_until_published(index, package)?;

        let git_tag = git_tag(&package.name, &package.version.to_string());
        repo.tag(&git_tag)?;
        repo.push(&git_tag)?;

        info!("published {}", package.name);

        if input.gh_release {
            publish_release(git_tag, input).await;
        }
    }
```
Notice that I reused the `git_tag` variable instead of calculating the git tag again :)

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +140,41 @@ fn release_package(
...
+    let github = GitHub::new(repo_url.clone().owner, repo_url.clone().name, secret_string);
+    let github_client = GitHubClient::new(&github);
+    if repo_url.is_on_github() {
+        let page = Octocrab::default()
+            .repos(repo_url.clone().owner, repo_url.name)
+            .releases()
+            .create(&git_tag)
+            .send()
+            .await
+            .unwrap();
```
MarcoIeni, line 173:
```suggestion
    let github_client = GitHubClient::new(&github).unwrap();
    if repo_url.is_on_github() {
        let page = github_client.create_release(&git_tag).await.unwrap();
```
Now that we have the github_client, we can use it to create the release!

I have extracted the release functionality into the `create_release` function. You can add this function [here](https://github.com/MarcoIeni/release-plz/blob/f03c598c38e8cf5ceeae542add5d54fae3915159/crates/release_plz_core/src/github_client.rs#L109):
```rust
    pub async fn create_release(&self, tag: &str) -> anyhow::Result<repos::Release> {
        self.client
            .repos(&self.github.owner, &self.github.repo)
            .releases()
            .create(tag)
            .send()
            .await
            .context("Failed to create release")
    }
```

crates/release_plz_core/src/release.rs
```diff
@@ -131,9 +140,41 @@ fn release_package(
...
+    }
+
     Ok(())
 }
 
+async fn publish_release(git_tag: String, input: &ReleaseRequest) {
+    let repo_url = match &input.repo_url {
+        Some(url) => RepoUrl::new(url.as_str()),
+        None => {
+            let repo = Repo::new(input.workspace_root().unwrap()).unwrap();
```
MarcoIeni, line 154: We already have created the `repo` on line 120.
We could move this piece of code in that function and reuse that variable, instead of creating it here

MarcoIeni: Also, we could extract these lines because they are repeated [here](https://github.com/MarcoIeni/release-plz/blob/f03c598c38e8cf5ceeae542add5d54fae3915159/crates/release_plz/src/args/update.rs#L92-L94).

I wrote this function you can add [here](https://github.com/MarcoIeni/release-plz/blob/f03c598c38e8cf5ceeae542add5d54fae3915159/crates/release_plz_core/src/repo_url.rs#L34):

```rust
    pub fn from_repo(repo: &Repo) -> Result<Self, anyhow::Error> {
        let url = repo.origin_url().context("cannot determine origin url")?;
        RepoUrl::new(&url)
    }
```

crates/release_plz_core/src/release.rs
```diff
@@ -59,7 +68,7 @@ pub async fn release(input: &ReleaseRequest) -> anyhow::Result<()> {
                 info!("{} {}: already published", package.name, package.version);
                 return Ok(());
             }
-            release_package(&mut index, package, input)?;
+            release_package(&mut index, package, input).await.unwrap();
```
MarcoIeni, line 71:
```suggestion
            release_package(&mut index, package, input).await?;
```
We can return an error here instead of panicking. Can you also do the same in the rest of your code?
You can replace the `unwrap`s with `?` and return `anyhow::Result` from the function.

MarcoIeni:
The CI looks green, congrats!
I think this could actually work at this stage.
Have you tried to test it by running `release-plz` locally to see if it creates the github release?

crates/release_plz_core/src/release.rs
```diff
@@ -129,11 +137,34 @@ fn release_package(
...
+async fn publish_release(
+    git_tag: String,
+    input: &ReleaseRequest,
+    repo: Repo,
+) -> anyhow::Result<()> {
+    let repo_url = match &input.repo_url {
+        Some(url) => RepoUrl::new(url.as_str()),
+        None => RepoUrl::from_repo(&repo),
+    }
+    .unwrap();
```
MarcoIeni, line 158:
```suggestion
    }?;
```

MarcoIeni:
It worked! Look [here](https://github.com/MarcoIeni/rust-workspace-example/releases/tag/marco-test-three-v0.1.4)! I will add the release name, otherwise it looks bad. See [this](https://github.com/MarcoIeni/rust-workspace-example/releases/tag/marco-test-three-v0.1.3).

PR #421, opened 2023-01-08
chore: add tests for gitea

MarcoIeni:
Close https://github.com/MarcoIeni/release-plz/issues/388
WIP
- [x] create a user token with [this](https://github.com/XAMPPRocky/octocrab/pull/291) endpoint
# useful links
- https://discourse.gitea.io/t/unattended-gitea-installation-from-the-cli/3373/5


MarcoIeni:
Currently, I'm trying to create a Gitea user from the admin user. However, the basic auth with the "root" user is not working. I don't know what the default root username and password are.

era:
I will take a look at the process for this since it seems like we may need to set up the admin user manually as per:  https://discourse.gitea.io/t/unattended-gitea-installation-from-the-cli/3373/12 and https://github.com/go-gitea/gitea/issues/4120

MarcoIeni:
This was in the PR:

> the first user registered to Gitea becomes the admin.

Now it's clearer, thanks.

However there's not a way to run an `exec` from docker compose.
I guess there's no way to create the first user via api, right?

era:
It does not look like we can use the API in order to create it (I may be wrong, I just couldn't find how to). But if we remove the `GITEA__security__INSTALL_LOCK=true` [flag](https://docs.gitea.io/en-us/config-cheat-sheet/#security-security), we could use the installation page in order to create it (it does feel a hacky solution tho): https://discourse.gitea.io/t/unattended-gitea-installation-from-the-cli/3373

MarcoIeni:
> we could use the installation page in order to create it

My goal is to run the tests in CI, so no manual steps should be required :/

era:
> My goal is to run the tests in CI, so no manual steps should be required :/

@MarcoIeni that would not be a manual step, we would invoke the installation page using our Rust code (for example, the link I gave on the last comment uses `curl`). The only difference is that since it's not meant for machines to access, if in some newer version, they change any of the form input names, we may have to update it (in other words, it's less stable than a proper API).

MarcoIeni:
oh ok. Yes, this can work, too :)

MarcoIeni:
Getting back to this today. Options:
1. run `gitea admin user create --username marco --password psw --email a@b.com` when running tests at the beginning.

PR #425, opened 2023-01-09
feat: handle pre-releases

MarcoIeni:
Closes https://github.com/MarcoIeni/release-plz/issues/174


crates/next_version/src/next_version.rs
```diff
@@ -14,6 +14,8 @@ pub trait NextVersion {
     fn increment_minor(&self) -> Self;
     /// Increments the patch version number for this Version.
     fn increment_patch(&self) -> Self;
+    /// Increments the patch version number for this Version.
```
sassman, line 17: nit: the comment says patch version while the actual subject is the pre-release version.

sassman (approved):
Very nice. If I get it right the semver version struct already increments the pre-release version according to the expectations right?

🚀

PR #424, opened 2023-01-09
Added gitea release functionality

gautamprikshit1:
Fixed #418

PR #428, opened 2023-01-10
Gitea release

gautamprikshit1:
Continuing #424

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .DS_Store
 .vscode
 .idea
+/vendor
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -43,9 +43,9 @@ pub struct UpdateRequest {
     /// Allow dirty working directories to be updated.
     /// The uncommitted changes will be part of the update.
     allow_dirty: bool,
-    /// If present, the new changelog entry contains a link to the diff between the old and new version.
+    /// Repository Url. If present, the new changelog entry contains a link to the diff between the old and new version.
     /// Format: `https://{repo_host}/{repo_owner}/{repo_name}/compare/{old_tag}...{new_tag}`.
-    github_repo: Option<RepoUrl>,
+    repo_url: Option<RepoUrl>,
 }
 
 #[derive(Debug, Clone)]
@@ -73,7 +73,7 @@ impl UpdateRequest {
             registry: None,
             update_dependencies: false,
             allow_dirty: false,
-            github_repo: None,
+            repo_url: None,
         })
     }
 
@@ -115,7 +115,7 @@ impl UpdateRequest {
 
     pub fn with_repo_url(self, repo_url: RepoUrl) -> Self {
         Self {
-            github_repo: Some(repo_url),
+            repo_url: Some(repo_url),
             ..self
         }
     }
@@ -368,7 +368,7 @@ impl Updater<'_> {
                 .git_tag(&package.name, &package.version.to_string());
             let next_tag = self.project.git_tag(&package.name, &version.to_string());
             self.req
-                .github_repo
+                .repo_url
                 .as_ref()
                 .map(|r| r.gh_release_link(&prev_tag, &next_tag))
         };


crates/release_plz/src/args/release.rs
```diff
@@ -26,15 +29,26 @@ pub struct Release {
...
+    #[clap(long, value_enum, default_value_t = GitBackendKind::GitHub)]
+    pub git_backend: GitBackendKind,
+}
+
+#[derive(ValueEnum, Clone, Copy, Debug, Eq, PartialEq)]
+pub enum GitBackendKind {
+    #[value(name = "github")]
+    GitHub,
+    #[value(name = "gitea")]
+    Gitea,
```
MarcoIeni, line 51: Can't you reuse the same type we have in the project already?

MarcoIeni (commented):
A comment from my smartphone. I will do a full review in the next days.

crates/release_plz/src/args/release.rs
```diff
@@ -52,13 +55,25 @@ impl TryFrom<Release> for ReleaseRequest {
...
+            token: token.clone(),
             dry_run: r.dry_run,
             git_release,
-            repo_url: r.repo_url,
+            repo_url: r.repo_url.clone(),
+            backend: match r.git_backend {
+                GitBackendKind::Github => GitBackend::Github(GitHub::new(
+                    RepoUrl::new(r.repo_url.clone().unwrap().as_str())?.name,
+                    RepoUrl::new(r.repo_url.unwrap().as_str())?.owner,
+                    token.unwrap(),
```
MarcoIeni, line 70: this is the token used for the cargo registry. You need to use the `git_token` instead.

crates/release_plz_core/src/gitea_client.rs
```diff
@@ -79,6 +85,31 @@ impl<'a> GiteaClient<'a> {
         Ok(())
     }
 
+    pub async fn create_gitea_release(
+        &self,
+        git_tag: String,
+        release_body: &str,
```
MarcoIeni, line 91: from the [API](https://try.gitea.io/api/swagger#/repository/repoCreateRelease) I see this fiels is called `body`, not `release_body`.
Was this a typo or maybe there's a reason behind this name? 🤔 
image

crates/release_plz_core/src/gitea_client.rs
```diff
@@ -13,6 +13,12 @@ pub struct GiteaClient<'a> {
     client: reqwest::Client,
 }
 
+#[derive(Serialize)]
+pub struct CreateReleaseOption<'a> {
+    tag_name: String,
+    release_body: &'a str,
```
MarcoIeni, line 19:
```suggestion
    release_body: &'a str,
```
from the [API](https://try.gitea.io/api/swagger#/repository/repoCreateRelease) I see this fiels is called `body`, not `release_body`.
Was this a typo or maybe there's a reason behind this name? 🤔 
image

crates/release_plz_core/src/gitea_client.rs
```diff
@@ -13,6 +13,12 @@ pub struct GiteaClient<'a> {
     client: reqwest::Client,
 }
 
+#[derive(Serialize)]
+pub struct CreateReleaseOption<'a> {
+    tag_name: String,
```
MarcoIeni, line 18: why this field is a string, while the other is an &str?
Why aren't them of the same type? 🤔

crates/release_plz_core/src/release.rs
```diff
@@ -30,10 +31,12 @@ pub struct ReleaseRequest {
...
     /// Perform all checks without uploading.
     pub dry_run: bool,
-    /// Publishes GitHub release.
+    /// Publishes GitHub/Gitea release.
     pub git_release: Option<GitRelease>,
-    /// GitHub repo URL.
+    /// GitHub/Gitea repo URL.
     pub repo_url: Option<String>,
+    /// Kind of git host where your project is hosted
+    pub backend: GitBackend,
```
MarcoIeni, line 39: I think it's easier to move this field into the `GitRelease` field. They are connected.

crates/release_plz/src/args/release.rs
```diff
@@ -26,32 +26,48 @@ pub struct Release {
...
+    /// Git Backend (GitHub or Gitea).
+    #[clap(long, value_enum, default_value_t = GitBackendKind::Github)]
+    pub git_backend: GitBackendKind,
 }
 
 impl TryFrom<Release> for ReleaseRequest {
     type Error = anyhow::Error;
 
     fn try_from(r: Release) -> Result<Self, Self::Error> {
+        let token = r.git_token.clone().map(SecretString::from).unwrap();
```
MarcoIeni, line 47:
```suggestion
        let token = r.git_token.clone().map(SecretString::from).context("git_token is required for git_release")?;
```
It's better to avoid unwrapping when the error depends on user input

crates/release_plz/src/args/release.rs
```diff
@@ -26,32 +26,48 @@ pub struct Release {
...
+                    GitBackendKind::Github => GitBackend::Github(GitHub::new(
+                        RepoUrl::new(r.repo_url.clone().unwrap().as_str())?.name,
+                        RepoUrl::new(r.repo_url.clone().unwrap().as_str())?.owner,
+                        token,
+                    )),
+                    GitBackendKind::Gitea => GitBackend::Gitea(Gitea::new(
+                        RepoUrl::new(r.repo_url.clone().unwrap().as_str())?,
+                        token,
+                    )?),
+                },
```
MarcoIeni, line 64: can you remove the `git_token` field from the GitRelease struct? (line 50)
The same token is already included in the `backend` field.

crates/release_plz/src/args/release.rs
```diff
@@ -26,32 +26,48 @@ pub struct Release {
...
+        let token = r.git_token.clone().map(SecretString::from).unwrap();
         let git_release = if r.git_release {
             let release = release_plz_core::GitRelease {
                 git_token: SecretString::from(
                     r.git_token
                         .context("git_token is required for git_release")?,
                 ),
+                backend: match r.git_backend {
+                    GitBackendKind::Github => GitBackend::Github(GitHub::new(
+                        RepoUrl::new(r.repo_url.clone().unwrap().as_str())?.name,
```
MarcoIeni, line 56: Here you are panicking if the user doesn't include the `repo_url` parameter.
However, [here](https://github.com/MarcoIeni/release-plz/blob/93fd96688c9c1e291844c52561d4caf7689ec018/crates/release_plz/src/args/update.rs#L86) we have a mechanism to determine automatically the `repo_url`.
It would be nice to extract that method and use it here, too.

crates/release_plz_core/src/release.rs
```diff
@@ -184,15 +188,25 @@ async fn publish_release(
...
     let repo_url = match repo_url {
         Some(url) => RepoUrl::new(url),
         None => RepoUrl::from_repo(&repo),
     }?;
-    if repo_url.is_on_github() {
-        let github = GitHub::new(repo_url.clone().owner, repo_url.clone().name, git_token);
-        let github_client = GitHubClient::new(&github)?;
-        let _page = github_client.create_release(&git_tag, release_body).await?;
+    match backend {
+        GitBackend::Github(_) => {
```
MarcoIeni, line 198:
```suggestion
        GitBackend::Github(github) => {
```
Here you already have the GitHub object, you don't need to rebuild it.
Same for gitea on line 203.

PR #429, opened 2023-01-11
feat: warn on old versions

gautamprikshit1:


MarcoIeni:
have you tested this?

gautamprikshit1:
image
Here's a screenshot of update checker in action.

MarcoIeni:
Have you tried running it with the latest version?
My doubt is that this warning will always be printed, because the `env!("CARGO_PKG_VERSION");` is not in the same format of the tag.

PR #469, opened 2023-01-18
fix: don't mark releases as latest if multiple public packages

MarcoIeni:


MarcoIeni:
I tested this. Apparently github needs a "latest" release. So this won't work.

PR #471, opened 2023-01-19
Feat: warn on old version

gautamprikshit1:
Added Update checker continuing #429
image
image

### DIFF
--- a/crates/release_plz_core/src/update_checker.rs
+++ b/crates/release_plz_core/src/update_checker.rs
@@ -21,7 +21,7 @@ pub async fn get_latest_version() -> String {
             .ok()
             .unwrap()
             .tag_name;
-        return tag_name;
+        tag_name
     } else {
         "".to_string()
     }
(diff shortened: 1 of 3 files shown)


crates/release_plz_core/src/update_checker.rs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        let tag_name = client_response
+            .json::<ClientResponse>()
+            .await
+            .ok()
+            .unwrap()
+            .tag_name;
+        tag_name
+    } else {
+        "".to_string()
```
MarcoIeni, line 26: if there is a connection issue, or github is down, an empty string is returned.
Then when you call this method you do `[13..]`, which results in a panic :/

Try running this without an internet connection, the program should crash. This is not what we want.
If the github api is not available, you can log a warning, but the program shouldn't be interrupted.

MarcoIeni:
Thanks!

MarcoIeni:
In https://github.com/MarcoIeni/release-plz/pull/477 I edited a few things.
Mainly, I removed some unwraps and I moved the check to its own sub-command: I didn't want to do a network request to GitHub every time the program was run. 

Again, thanks!

PR #472, opened 2023-01-19
chore: add tests for gitea

### LINKED ISSUE #388: Test release pr with gitea docker image
Gitea is available as a docker image.
Let's use it to test the release-pr command.

## test plan
1. Run release-plz binary
2. Assert that the pr has been created by doing a GET to gitea API

era:
Close #388
Based on https://github.com/MarcoIeni/release-plz/pull/421 (and the discussions there)

Some notes:

- I kept the pattern of panic! instead of returning Result for the tests, since the original code was doing it.
- I'm not sure how to make Github run the `docker compose up` (or even if I have permission to add such a thing). If you want me to include this in the PR somehow, could you explain to me how?
- I tested it on my machine (a Linux machine). I don't see anything that could go wrong with other platforms, but just liked to highlight :)


MarcoIeni:
If you prefer, instead of filling the form, we can also do a `docker exec` from the rust code, where we create the user. What do you think?

era:
@MarcoIeni that would be much better, I will modify the code to use docker exec instead

MarcoIeni:
Hey era, I'm going to close this PR.
This weekend I took inspiration from your work and I got https://github.com/MarcoIeni/release-plz/pull/421 working locally.
It creates one random user per test.
It's still not perfect. Next, I will mock the cargo registry and merge that PR. 👍

PR #516, opened 2023-02-03
feat(release): allow passing flags to cargo publish

### LINKED ISSUE #515: Allow passing arbitrary flags to cargo publish
## Motivations

- Would you like to implement this feature? [y/n] y
- I am working on a project that needs me to pass `--no-verify` to `cargo publish`.
- #129

## Solution

I think it would be good if you could pass arbitrary flags to `cargo publish` (for the reason above). Maybe it could be something like `release-plz publish --publish-flags "--allow-dirty"`?

## Alternatives

You can kind of configure `cargo` through environment variables, but you can't really do anything like this. Even if you could, the solution is hacky at best.

## Additional context
Here's an example of what happens in the project I'm working on:
```
$ release-plz release
...
Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: ...

  To proceed despite this, pass the `--no-verify` flag.
```
However, there is no way to pass the flag (AFAIK).

Wyatt-Stanke:
This allows for you to be able to pass any flags you want when running `plz-release publish` to `cargo publish`.

Fixes #515

### DIFF
--- a/crates/release_plz/src/args/release.rs
+++ b/crates/release_plz/src/args/release.rs
@@ -29,6 +29,9 @@ pub struct Release {
     /// Publish GitHub release for the created git tag.
     #[clap(long)]
     pub git_release: bool,
+    /// Flags to pass to `cargo publish`
+    #[clap(long)]
+    pub publish_flags: Option<String>,
     /// GitHub repository url.
     #[clap(long, value_parser = NonEmptyStringValueParser::new())]
     pub repo_url: Option<String>,
@@ -59,6 +62,7 @@ impl TryFrom<Release> for ReleaseRequest {
             dry_run: r.dry_run,
             git_release,
             repo_url: r.repo_url,
+            publish_flags: r.publish_flags,
         })
     }
 }
--- a/crates/release_plz/src/args/release.rs
+++ b/crates/release_plz/src/args/release.rs
@@ -30,8 +30,8 @@ pub struct Release {
     #[clap(long)]
     pub git_release: bool,
     /// Flags to pass to `cargo publish`
-    #[clap(long)]
-    pub publish_flags: Option<String>,
+    #[clap(long, value_delimiter = ' ')]
+    pub publish_flags: Vec<String>,
     /// GitHub repository url.
     #[clap(long, value_parser = NonEmptyStringValueParser::new())]
     pub repo_url: Option<String>,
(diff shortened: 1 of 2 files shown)


Wyatt-Stanke:
Successfully published a workspace with this!

crates/release_plz_core/src/release.rs
```diff
@@ -33,6 +33,8 @@ pub struct ReleaseRequest {
     pub git_release: Option<GitRelease>,
     /// GitHub repo URL.
     pub repo_url: Option<String>,
+    /// Extra flags for `cargo publish`.
+    pub publish_flags: Option<String>,
```
MarcoIeni, line 37: Here in the library (release_plz_core), I would prefer to have a Vec<String> instead.

crates/release_plz/src/args/release.rs
```diff
@@ -29,6 +29,9 @@ pub struct Release {
     /// Publish GitHub release for the created git tag.
     #[clap(long)]
     pub git_release: bool,
+    /// Flags to pass to `cargo publish`
+    #[clap(long)]
+    pub publish_flags: Option<String>,
```
MarcoIeni, line 34:
```suggestion
    #[clap(long, value_delimiter = ' ')]
    pub publish_flags: Vec<String>,
```
You can let clap do the parsing :)

Wyatt-Stanke: I did not know that! Thanks!

crates/release_plz_core/src/release.rs
Wyatt-Stanke: Fixed!

crates/release_plz/src/args/release.rs
Wyatt-Stanke: Fixed!

MarcoIeni:
I thought about this a bit more, and I think it's a bad idea to allow passing any flags to cargo publish.
I think we should add `--no-verify`, `--allow-dirty` and all the other flags that we are interested in, directly as release-plz args (similar to what we do for `--dry-run` for example).

The reason is that release-plz already takes some flags that are passed as arguments to cargo publish, such as `--registry` or `--dry-run`.
What happens if the flags are specified twice? For example, what should we do if the user passes both `--registry` to release-plz and to `cargo publish`?
I want to avoid this possible clash.

I'm sorry you worked on this and I closed the pr, it's my fault! I changed my mind too late.

If you don't want to add the `--no-verify` and `--allow-dirty` (and all the other flags you need) to `release-plz release`, I would understand. In this case, I will add these flags next week.

Otherwise, I'm happy to review another PR from you :)

PR #607, opened 2023-03-16
fix: do git fetch before rebase

MarcoIeni:

### DIFF
--- a/crates/git_cmd/src/lib.rs
+++ b/crates/git_cmd/src/lib.rs
@@ -93,6 +93,11 @@ impl Repo {
         Ok(())
     }
 
+    pub fn fetch(&self, obj: &str) -> anyhow::Result<()> {
+        self.git(&["fetch", "origin", obj])?;
+        Ok(())
+    }
+
     pub fn force_push(&self, obj: &str) -> anyhow::Result<()> {
         self.git(&["push", "origin", obj, "--force"])?;
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/release_plz_core/src/release_pr.rs
```diff
@@ -169,6 +169,8 @@ fn reset_branch(pr: &GitPr, commits_number: usize, repository: &Repo) -> anyhow:
     let head = format!("HEAD~{commits_number}");
     repository.git(&["reset", "--hard", &head])?;
 
+    repository.git(&["fetch"])?;
```
sassman, line 172: Yes exactly. Since fetch does not change anything to the staged / working set files it could also be executed before the `git reset` call. But I guess it's not important for now.

MarcoIeni:
I did an additional change where I specified the branch I want to fetch.
If it looks good to you, I will merge :)

crates/git_cmd/src/lib.rs
```diff
@@ -93,6 +93,11 @@ impl Repo {
         Ok(())
     }
 
+    pub fn fetch(&self, obj: &str) -> anyhow::Result<()> {
+        self.git(&["fetch", "origin", obj])?;
```
sassman, line 97: The thing with "origin" is, that it would work on ci for sure but locally not necessarily.

I for instance use always just "o" as the upstream name. No "origin" present in my case. 
But this can be determined from git.

MarcoIeni: Right now the rest of the code assumes "origin". I opened an issue for this: https://github.com/MarcoIeni/release-plz/issues/608
So I will merge this as is :)

PR #639, opened 2023-03-22
chore(refactor): introduce local package

MarcoIeni:
With the LocalPackage new type, we don't need to unwrap so often to find package path and manifest


MarcoIeni:
not worth it

PR #689, opened 2023-04-14
fix: read full commit message

### LINKED ISSUE #687: Breaking changes stated in commit messages do not bump the major version
## Bug description

Currently a commit message containing a breaking change footer does not lead to a major version bump. This happens when either `SemVerCheck` is disabled or is unable to detect a breaking API change. There is already a [code path](https://github.com/MarcoIeni/release-plz/blob/main/crates/release_plz_core/src/version.rs#L20) with a proper version detection logic for commit messages. However only commit message headers are passed, so that the footer is never evaluated.

## To Reproduce

1. Have a commit message with a  breaking change footer, e .g.
```
feat: xyz

BREAKING CHANGE: blubb
``` 
2.  Disable `SemVerCheck` with a `release-plz.toml`:
``` 
[workspace]
semver_check = "no"
``` 
3. Run `release-plz update`

Reproducable on `main`, this only leads to a minor version bump.

## Expected behavior

Major version bump.

## Environment

- OS: Manjaro Linux 
- release-plz version: upstream main

## Additional context

If I am not mistaken, the fix can actually be quite easy. Just [change the formatting](https://github.com/MarcoIeni/release-plz/compare/main...TobTheRock:release-plz:fix/breaking-changes-not-detected) of the `git log` call, so that full raw message is returned. If desired, I can open a PR with this.

TobTheRock:
By changing the formatting the whole raw commit message is returned by `git log`.This way when determining the next version in `NextVersionFromDiff` (without semverchecK), commits with `Breaking Version` footers now also lead to a major version bump.

This fixes #687 .

### DIFF
--- a/crates/git_cmd/src/lib.rs
+++ b/crates/git_cmd/src/lib.rs
@@ -242,7 +242,7 @@ impl Repo {
     }
 
     pub fn current_commit_message(&self) -> anyhow::Result<String> {
-        self.git(&["log", "-1", "--pretty=format:%s"])
+        self.git(&["log", "-1", "--pretty=format:%B"])
     }
 
     pub fn tag(&self, name: &str) -> anyhow::Result<String> {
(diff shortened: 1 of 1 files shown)


MarcoIeni (approved):
Thanks!

PR #691, opened 2023-04-15
fix: don't return early when publishing crates

satlank:
The loop over all release artefacts should not return when it finds a crate that is already published but instead continue to evaluate all other to-be-released artefacts.


MarcoIeni (approved):
Thanks!

PR #718, opened 2023-04-23
Add a flag to prevent publishing

### LINKED ISSUE #719: Configuration: add a `publish` flag to enable/disable `cargo publish`
## Motivations

People might want to run cargo-publish by themselves.

## Solution

Add publish flag to the configuration file that enables/disables cargo publish (defaults to true).

## Hints

Add a `publish` bool [here](https://github.com/MarcoIeni/release-plz/blob/d74ce3776c0d7e6daca484e935ad75b6f1f981d7/crates/release_plz/src/config.rs#L270) and use it similarly to how we use `publish_allow_dirty` and `publish_no_verify`

## Alternatives

we could have a cli flag, but from the config file, we can allow configuring this field per package, similarly to how we do for `publish_allow_dirty` and `publish_no_verify`.

## Additional context

Originally discussed [here](https://github.com/MarcoIeni/release-plz/issues/590#issuecomment-1518662733)

flying-sheep:
Fixes #719 

Adds a `--no-publish` CLI flag and `publish` config option. They can be used to prevent publishing to a create registry.

The option interacts with the `--registry` flag and the `publish` field in Cargo.toml:

- If both `--no-publish` and `--registry=...` is set, we bail, as this combination is probably an user error.
- If the `publish` field in Cargo.toml is set to a non-empty list of registries (or `true`), the package is considered “publishable”. If `--no-publish` is set, the `release_package` function will now skip running `cargo publish` for it.

One thing I don’t quite understand: will `release_package` be run for unpublishable packages? Before my change, `cargo publish` was run for the same set of packages that had Git tags and GitHub releases created for them. Will `cargo publish` just bail without triggering the `!stderr.contains("Uploading") || stderr.contains("error:")` condition for unpublishable packages?

TODO:

- [x] Link issue for this feature
- [ ] Add tests once this looks like it does what it should (see above question)
- [x] Changelog entry


MarcoIeni:
I think there was a misunderstanding, I wanted to add a new field in the configuration file, not a new CLI flag :(
This is the issue I created: https://github.com/MarcoIeni/release-plz/issues/719#issue-1680109705
Sorry if the previous direction wasn't good enough. I should have pointed to the right part of the code to modify, as I did in the issue I just created.

MarcoIeni:
No, [here](https://github.com/MarcoIeni/release-plz/blob/d74ce3776c0d7e6daca484e935ad75b6f1f981d7/crates/release_plz_core/src/release.rs#L237) we create the project. 
The project [filters](https://github.com/MarcoIeni/release-plz/blob/739cc65cfd1b543e76679a16f94bacc5a974673d/crates/release_plz_core/src/next_ver.rs) the unpublishable packages

flying-sheep:
> I think there was a misunderstanding, I wanted to add a new field in the configuration file, not a new CLI flag

Got it!

> The project [filters](https://github.com/MarcoIeni/release-plz/blob/739cc65cfd1b543e76679a16f94bacc5a974673d/crates/release_plz_core/src/next_ver.rs) the unpublishable packages

Ah, so there won’t be any git tags created for unpublishable packages either!

MarcoIeni:
> Ah, so there won’t be any git tags created for unpublishable packages either!

Yes, unpublishable packages aren't updated or released at all. Release-plz ignores them 👍

flying-sheep:
Hmm, I don’t see any tests for `release_plz`, and none for the part I modified in `release_plz_core`. I’m happy to add tests to an existing structure, but I don’t think I’m the right person to introduce tests for a whole part of the program that is so far untested.

… or I just missed where the tests are!

MarcoIeni:
You are right, there are not many tests at the moment.
But [here](https://github.com/MarcoIeni/release-plz/blob/1ca4abb0f4e17fe05ff9b7617166c6434f89655f/crates/release_plz/src/args/release.rs#L128) we test the configuration and command line args.

I will work on https://github.com/MarcoIeni/release-plz/issues/451 and https://github.com/MarcoIeni/release-plz/issues/388 soon.

Can you remove the `--no-publish` CLI flag. IMO the `publish` option is enough.

crates/release_plz/src/config.rs
```diff
@@ -262,6 +272,9 @@ impl PackageReleaseConfig {
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Default, Clone)]
 pub struct ReleaseConfig {
+    /// If `Some(false)`, don't run `cargo publish`.
+    #[serde(rename = "publish")]
```
MarcoIeni, line 276:
```suggestion
```
not needed, it's `publish` by default.

MarcoIeni (approved):
Thank you!

MarcoIeni:
added docs in https://github.com/MarcoIeni/release-plz/pull/759

PR #754, opened 2023-05-04
prevent run-semver-checks in all packages

### LINKED ISSUE #645: Run cargo-semver-checks in parallel on all packages
## Motivations

cargo-semver-checks requires some time to run.
We could speed things up by running it in parallel on all packages.

## Solution

semver checks is run [here](https://github.com/MarcoIeni/release-plz/blob/8a87d56c704f41c18353ce94c3c8e2e009ef2e62/crates/release_plz_core/src/next_ver.rs#L444).

## Alternatives

One could thing that we could run all the package update process in parallel. However at the moment we can't because we run git to checkout at different commits. So running it at the same time in the same project conflicts.

lmtan91:
Fix #645 
Move run_semver_check portion out of get_diff function


crates/release_plz_core/src/next_ver.rs
```diff
@@ -425,14 +425,32 @@ impl Updater<'_> {
...
                 &self.project.root,
             )?;
+
+            let package_path = {
+                let relative_path = p
+                    .package_path()?
+                    .strip_prefix(&self.project.root)
+                    .context("error while retrieving package_path")?;
+                repository.directory().join(relative_path)
+            };
```
MarcoIeni, line 443: This looks great, thanks!
Just one comment: can you extract this into a function? It's also repeated at line 610.

lmtan91: fix findings. Can you pls review again?

crates/release_plz_core/src/next_ver.rs
```diff
@@ -578,6 +594,18 @@ fn get_changelog(
     Ok(changelog)
 }
 
+fn get_package_path(
+    package: &Package,
+    repository: &Repo,
+    project_root: &Path,
+) -> PathBuf {
```
MarcoIeni, line 601:
```suggestion
) -> anyhow::Result<PathBuf> {
```
Let's return a Result, so that we don't panic in the unwraps.

crates/release_plz_core/src/next_ver.rs
```diff
@@ -578,6 +594,18 @@ fn get_changelog(
...
+    let package_path = package.package_path().unwrap();
+    let relative_path = package_path.strip_prefix(project_root).unwrap();
```
MarcoIeni, line 603:
```suggestion
    let relative_path = package_path.strip_prefix(project_root).context("error while retrieving package_path: project root not found")?;
```

crates/release_plz_core/src/next_ver.rs
```diff
@@ -578,6 +594,18 @@ fn get_changelog(
     Ok(changelog)
 }
 
+fn get_package_path(
+    package: &Package,
+    repository: &Repo,
+    project_root: &Path,
+) -> PathBuf {
+    let package_path = package.package_path().unwrap();
```
MarcoIeni, line 602:
```suggestion
    let package_path = package.package_path().context("error while retrieving package_path: project root not found")?;
```
Can you run `cargo fmt` after accepting these changes?

crates/release_plz_core/src/next_ver.rs
lmtan91: fixed

crates/release_plz_core/src/next_ver.rs
lmtan91: done

MarcoIeni: I still see the unwrap here D:

MarcoIeni:
can you also fix the clippy lint?

crates/release_plz_core/src/next_ver.rs
```diff
@@ -578,24 +592,33 @@ fn get_changelog(
...
+}
+
 #[instrument(
     skip_all,
     fields(package = %package.name)
 )]
 fn get_diff(
     package: &Package,
-    run_semver_check: RunSemverCheck,
+    _run_semver_check: RunSemverCheck,
```
MarcoIeni, line 615:
```suggestion
```
let's remove it since it's not used :)

crates/release_plz_core/src/next_ver.rs
```diff
@@ -425,14 +425,28 @@ impl Updater<'_> {
         let mut packages_to_check_for_deps: Vec<&Package> = vec![];
         let mut packages_to_update = PackagesUpdate { updates: vec![] };
         for p in &self.project.packages {
-            let semver_check = self.req.get_package_config(&p.name).semver_check();
-            let diff = get_diff(
+            let run_semver_check = self.req.get_package_config(&p.name).semver_check();
+            let mut diff = get_diff(
                 p,
-                semver_check,
+                run_semver_check,
```
MarcoIeni, line 431:
```suggestion
```

MarcoIeni:
I pushed a small change to remove the argument from the `get_diff` function.
Can you run `cargo fmt` again? Sorry about that 😅

crates/release_plz_core/src/next_ver.rs
```diff
@@ -578,24 +591,32 @@ fn get_changelog(
...
     project_root: &Path,
 ) -> anyhow::Result<Diff> {
-    let package_path = {
-        let relative_path = package
-            .package_path()?
-            .strip_prefix(project_root)
-            .context("error while retrieving package_path")?;
-        repository.directory().join(relative_path)
-    };
+    let package_path = get_package_path(package, repository, project_root).unwrap();
```
MarcoIeni, line 618:
```suggestion
    let package_path = get_package_path(package, repository, project_root)?;
```
let's return the error to the caller instead of unwrap 👍

crates/release_plz_core/src/next_ver.rs
lmtan91: fixed :)

crates/release_plz_core/src/next_ver.rs
```diff
@@ -425,14 +425,22 @@ impl Updater<'_> {
...
-                p,
-                semver_check,
-                registry_packages,
-                repository,
-                &self.project.root,
-            )?;
+            let run_semver_check = self.req.get_package_config(&p.name).semver_check();
+            let mut diff = get_diff(p, registry_packages, repository, &self.project.root)?;
+
+            let package_path = get_package_path(p, repository, &self.project.root).unwrap();
```
MarcoIeni, line 431:
```suggestion
            let package_path = get_package_path(p, repository, &self.project.root)?;
```
same here

MarcoIeni:
I'm testing this (from my laptop). I tried committing something in `release-plz` crate. Then I run `release-plz update`:
- from `main` branch it took 19s
- from your branch it took 9s!

This is with cargo sparse registry disabled, so it should be even faster starting from the next rust version. Thanks 🥳

PR #889, opened 2023-08-02
fix: update workspace version in dependencies

MarcoIeni:
fix https://github.com/MarcoIeni/release-plz/issues/877


MarcoIeni:
let's go
image

PR #892, opened 2023-08-05
chore: test release-plz release

MarcoIeni:
Close https://github.com/MarcoIeni/release-plz/issues/451


MarcoIeni:
cargo-http-registry doesn't support git smart protocol. Only dumb protocol.
gix only supports the smart protocol.

either:
- try with cargo http registry path (instead of git index)
- try with shipyard if docker container is available

PR #967, opened 2023-09-15
feat: pass commit ids to git-cliff

### LINKED ISSUE #954: Pass commit IDs to git-cliff
Hi, thanks for creating this project! It's really easy to use and the results have been great.

## Motivations 
When the changelog is created, the actual commit IDs don't get passed along to `git-cliff`, it uses a dummy value instead: https://github.com/MarcoIeni/release-plz/blob/a88c026880d98da1a4884f7e73a937304a90d1f3/crates/release_plz_core/src/changelog.rs#L131

If your `git-cliff` config references the commit ID in the template like in [their example](https://github.com/orhun/git-cliff/blob/890de0007e108059378b134e9b9d0af2a6965027/cliff.toml#L27-L30):
```
{% macro commit(commit) -%}
    - {% if commit.scope %}*({{ commit.scope }})* {% endif %}{% if commit.breaking %}[**breaking**] {% endif %}\
        {{ commit.message | upper_first }} - ([{{ commit.id | truncate(length=7, end="") }}](<REPO>/commit/{{ commit.id }}))\
{% endmacro -%}
``` 

it will just show the dummy value instead of the actual commit ID.

- Would you like to implement this feature? 
I'd be happy to submit a PR.

## Solution

The commit ID is already retrieved here: https://github.com/MarcoIeni/release-plz/blob/a88c026880d98da1a4884f7e73a937304a90d1f3/crates/release_plz_core/src/next_ver.rs#L709

I think we just need to pass that info along with the commit message when the changelog is created.

## Alternatives

None that I can think of.

aschey:
Resolves #954 

This change passes the commit IDs to `git-cliff` when the changelog is generated, allowing users to use commit IDs in their changelog formats.

When using a commit message without an actual commit, I changed the placeholder value to "N/A", indicating there is no corresponding commit ID available. Happy to discuss this further if you'd like to handle this case differently.

### DIFF
--- a/crates/release_plz_core/src/changelog.rs
+++ b/crates/release_plz_core/src/changelog.rs
@@ -77,18 +77,18 @@ fn default_git_cliff_config(header: Option<String>, release_link: Option<&str>) 
     }
 }
 
-pub struct ChangelogBuilder {
-    commits: Vec<String>,
+pub struct ChangelogBuilder<'a> {
+    commits: Vec<Commit<'a>>,
     version: String,
     config: Option<Config>,
     release_date: Option<NaiveDate>,
     release_link: Option<String>,
 }
 
-impl ChangelogBuilder {
-    pub fn new(commits: Vec<impl Into<String>>, version: impl Into<String>) -> Self {
+impl<'a> ChangelogBuilder<'a> {
+    pub fn new(commits: Vec<Commit<'a>>, version: impl Into<String>) -> Self {
         Self {
-            commits: commits.into_iter().map(|s| s.into()).collect(),
+            commits,
             version: version.into(),
             config: None,
             release_date: None,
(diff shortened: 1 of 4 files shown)


MarcoIeni:
This is awesome, thanks! 🙌
Just one change request: can you put `"N/A"` into a `const` and write your explanation as a comment ("there is no corresponding commit ID available")?
So that "N/A" isn't duplicated across the codebase and we avoid typos 👍

aschey:
> This is awesome, thanks! 🙌 Just one change request: can you put `"N/A"` into a `const` and write your explanation as a comment ("there is no corresponding commit ID available")? So that "N/A" isn't duplicated across the codebase and we avoid typos 👍

Very good point, made that change and fixed the clippy error.

MarcoIeni:
Thank you 🥳

aschey:
Thanks for the quick turnaround!

PR #985, opened 2023-09-23
fix: respect git-cliff sort order

aschey:
This change makes the changelog respect the `sort_commits` setting in `git-cliff`. `git-cliff` implements this check [here](https://github.com/orhun/git-cliff/blob/c3556135f1304fbacdf3e26c5c7f1013ebbbd6f2/git-cliff/src/lib.rs#L177-L181). Unfortunately, `git-cliff` defaults to [sorting by oldest commits first](https://github.com/orhun/git-cliff/blob/c3556135f1304fbacdf3e26c5c7f1013ebbbd6f2/git-cliff/src/args.rs#L179) whereas the code here currently defaults to sorting by newest commits first. Ideally, the default behavior should match, but changing everyone's commit order who hasn't specified this setting would probably be a rather disruptive change, so I left the current default behavior alone. Open to discussion on this though.

### DIFF
--- a/crates/release_plz_core/src/changelog.rs
+++ b/crates/release_plz_core/src/changelog.rs
@@ -7,6 +7,7 @@ use git_cliff_core::{
     release::Release,
 };
 use regex::Regex;
+use tracing::warn;
 
 use crate::changelog_parser;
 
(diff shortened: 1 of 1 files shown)


MarcoIeni:
> Open to discussion on this though.

I'm not too sure on this one, to be honest. Let's keep the current behavior for now, thanks!

aschey:
Sounds good, thank you!

PR #998, opened 2023-09-26
docs: clarify faq about editing changelog

robo9k:
When initially looking at the `release-plz` docs I was wondering why/how I would want to edit the changelog when it's in two places:
- the expandable generated changelog text in the merge request description
- the commited changelog file in the Git repo

`release-plz` is doing the sensible thing here, using the file in the repo. Nonetheless I was wondering if I would just edit its pull request description, but turns out that's not needed / doesn't work 😅


MarcoIeni:
Thanks! I just changed the wording to be more compatible with my writing style 👍

PR #1050, opened 2023-10-27
fix: use registry argument on publish

frenzox:
The `registry` argument was not being used when calling `cargo publish`, causing it to try to publish to `crates.io` even when a different registry was passed.


MarcoIeni:
woops, I'm sorry 😅 hopefully this didn't cause any trouble.

PR #1057, opened 2023-11-05
feat: support .release-plz.toml as a config file

robjtede:
Keeping configuration in dotfiles is often desirable. This PR supports a root `.release-plz.toml` config file if `release-plz.toml` is not found.

If you'd rather this be considered supported already through the `config` CLI arg then I would understand, but would also suggest adding support for that arg to the GH action.

### DIFF
--- a/crates/release_plz/src/args/mod.rs
+++ b/crates/release_plz/src/args/mod.rs
@@ -57,20 +57,52 @@ fn local_manifest(project_manifest: Option<&Path>) -> PathBuf {
 }
 
 fn parse_config(config_path: Option<&Path>) -> anyhow::Result<Config> {
-    let config_path: PathBuf = config_path
-        .map(|p| p.to_path_buf())
-        .unwrap_or("release-plz.toml".into());
-    match std::fs::read_to_string(&config_path) {
-        Ok(config) => {
-            info!("using release-plz config file {}", config_path.display());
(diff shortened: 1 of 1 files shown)


MarcoIeni:
Hey Rob 👋 
I like the idea of falling back to `.release-plz.toml` if `release-plz.toml` is not found.
However, if the user specifies a non-existing config file, we should fail, instead of falling back to `.release-plz.toml`. 
Can you check if the config file specified by the user (if any) exists before falling back to `release-plz.toml` and `.release-plz.toml`?

Also, can you document `.release-plz.toml` here?
https://github.com/MarcoIeni/release-plz/blob/b18eccb9dc95a057c0f2d2726530db860d72a575/website/docs/config.md#L3

> but would also suggest adding support for that arg to the GH action.

Good idea! Created https://github.com/MarcoIeni/release-plz-action/issues/92 👍

robjtede:
> if the user specifies a non-existing config file, we should fail

Okay, I've implemented it like that. Though, it seems to be a change from the existing code paths for a specified, but non-existent config file.

robjtede:
```console
 › bat foo
[bat error]: 'foo': No such file or directory (os error 2)
 › release-plz update --config='foo'
  2023-11-05T13:42:38.998146Z  INFO  release-plz config file not found, using default configuration
```

MarcoIeni:
Oh, you are right! Well, thanks for fixing that, too! 🙏

PR #1079, opened 2023-11-19
Local registries are already supported

mre:
As discussed with @MarcoIeni, this TODO is not applicable.
Users can set their own registry through the configuration.


MarcoIeni:
brilliant!

MarcoIeni (approved):

PR #1080, opened 2023-11-19
Prevent typos in overrides

### LINKED ISSUE #673: Return error if configuration file contains configuration for a package that doesn't exist
## Motivations

We want to return an error if a user writes a typo in their config file.
[This](https://github.com/MarcoIeni/release-plz/blob/2f042b5104badf3ee4a97ac321fb26215544ad26/crates/release_plz_core/src/next_ver.rs#L338) function might be a good point to return an error

## Solution

You could add another argument to this function where you pass all the packages present in the [overrides](https://github.com/MarcoIeni/release-plz/blob/2f042b5104badf3ee4a97ac321fb26215544ad26/crates/release_plz_core/src/next_ver.rs#L62).
Same thing for [these](https://github.com/MarcoIeni/release-plz/blob/2f042b5104badf3ee4a97ac321fb26215544ad26/crates/release_plz_core/src/release.rs#L128) overrides.

mre:
Fixes #673

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -311,7 +311,8 @@ impl ReleaseRequest {
 /// Release the project as it is.
 #[instrument]
 pub async fn release(input: &ReleaseRequest) -> anyhow::Result<()> {
-    let project = Project::new(&input.local_manifest, None)?;
+    let overrides = input.packages_config.overrides.keys().cloned().collect();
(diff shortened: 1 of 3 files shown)


MarcoIeni:
Great work, thanks! I slightly changed the tests because we aren't testing wrong dependency names, but just wrong package names.
I hope you don't mind I did this small change by myself :)

PR #1083, opened 2023-11-24
fix: resolve issue on Windows machines that use CRLF that would duplicate the header on each update

AntoineGS:
I ran into an issue on Windows where the header would be added (but not replaced\removed) on each update.
This is because the line `## [Unreleased]` would end with CRLF but the code would assume it was LF.
So when the `prepend` function in git-cliff would be called, the `replacen` function would not get a match.

Here is a screenshot of the generated file showing the characters:
image

I added a test too!

### DIFF
--- a/crates/release_plz_core/src/changelog_parser.rs
+++ b/crates/release_plz_core/src/changelog_parser.rs
@@ -10,12 +10,12 @@ use regex::Regex;
 ///   (in the ..anything.. case, `## ..anything..` is not included in the header)
 pub fn parse_header(changelog: &str) -> Option<String> {
     lazy_static::lazy_static! {
-        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])").unwrap();
-
+        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])(.*?)(\n)").unwrap();
         static ref SECOND_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*?)(\n## )").unwrap();
     }
+
     if let Some(captures) = FIRST_RE.captures(changelog) {
-        return Some(format!("{}\n", &captures[0]));
+        return Some(format!("{}", &captures[0]));
     }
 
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/changelog_parser.rs
```diff
@@ -10,12 +10,12 @@ use regex::Regex;
...
     lazy_static::lazy_static! {
-        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])").unwrap();
-
+        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])(.*?)(\n)").unwrap();
         static ref SECOND_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*?)(\n## )").unwrap();
     }
+
     if let Some(captures) = FIRST_RE.captures(changelog) {
-        return Some(format!("{}\n", &captures[0]));
+        return Some(format!("{}", &captures[0]));
```
MarcoIeni, line 18:
```suggestion
        return Some(&captures[0].to_string());
```

MarcoIeni (approved):
awesome, thanks!

crates/release_plz_core/src/changelog_parser.rs
```diff
@@ -10,12 +10,12 @@ use regex::Regex;
...
     lazy_static::lazy_static! {
-        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])").unwrap();
-
+        static ref FIRST_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*)(## Unreleased|## \[Unreleased\]|## unreleased|## \[unreleased\])(.*?)(\n)").unwrap();
         static ref SECOND_RE: Regex = Regex::new(r"(?s)^(# Changelog|# CHANGELOG|# changelog)(.*?)(\n## )").unwrap();
     }
+
     if let Some(captures) = FIRST_RE.captures(changelog) {
-        return Some(format!("{}\n", &captures[0]));
+        return Some(&captures[0].to_string());
```
MarcoIeni, line 18:
```suggestion
        return Some(captures[0].to_string());
```

PR #1088, opened 2023-11-25
feat: add publish_timeout to avoid release blocking issue, fix #1015

### LINKED ISSUE #1015: Add a timeout when checking if a crate has been published
@orhun had the issue where [this](https://github.com/MarcoIeni/release-plz/blob/a92629ed10b8bb42dde426c0f0001aebbb6fa70e/crates/release_plz_core/src/cargo.rs#L62) function never finished. We should add a timeout (1 minute?) so that at least we can report what the error is.

ludusrusso:
add publish_timeout to avoid release blocking issue, fix #1015

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)


MarcoIeni:
Thanks for this ❤️ 

I committed some changes to your branch, I hope you don't mind:
- I removed the `duration-str` serde integration, so that we don't have to manually implement the serialization. In this way the code looks simpler imo, because we don't have the serde macros
- I added more docs to `config.md`
- I rentroduced the `if !logged`, so that if `cargo publish` goes smooth the first time, we don't log the message.

PR #1098, opened 2023-11-27
feat: Add release flag

### LINKED ISSUE #1020: configuration: add a `release` flag that disables cargo publish, git tag and git release
## Motivations

You might have a big workspace, where you don't want release-plz to manage all crates

https://github.com/MarcoIeni/release-plz/discussions/1019#discussioncomment-7198206

## Solution

This flag tells release-plz to ignore a package. You can also set it at the workspace level to ignore every package and enable it at a package level.

When `release = false` Release-plz doesn't update the package, and doesn't release it (i.e. `cargo publish`, `git tag` and github/gitea/gitlab release).

By default this flag is true.

AntoineGS:
resolves #1020 :
This flag tells release-plz to ignore a package. You can also set it at the workspace level to ignore every package and enable it at a package level.

When release = false Release-plz doesn't update the package, and doesn't release it (i.e. cargo publish, git tag and github/gitea/gitlab release).

By default this flag is true.


crates/release_plz/src/config.rs
```diff
@@ -144,12 +144,8 @@ pub struct ReleasePrConfig {
 /// Config at the `[[package]]` level.
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub struct PackageSpecificConfig {
-    /// Options for the `release-plz update` command (therefore `release-plz release-pr` too).
-    #[serde(flatten)]
-    update: PackageUpdateConfig,
-    /// Options for the `release-plz release` command.
     #[serde(flatten)]
-    release: PackageReleaseConfig,
+    package_config: PackageConfig,
```
MarcoIeni, line 148: I'm sorry to ask you for changes, but I don't like this refactoring (in the future I would like to make the config struct more flatten, instead of nested.
Can you revert this change and put the option in this struct (together with `changelog_path` and `changelog_include` which are also common to the `update` and `release` commands)?

MarcoIeni: Any particular reason why you did this change?

AntoineGS: > Any particular reason why you did this change?

The struct `PackageConfig` had the same definition as the struct `PackageSpecificConfig` (minus the two `changelog_*`)so as I was adding the configuration it seemed to made sense to centralize them.
Though now that you mention that the `changelog_*` configs were also common I am not seeing what the `PackageSpecificConfig` does versus the `PackageConfig`.
My understanding is that `PackageConfig` is the default configuration for packages and the `PackageSpecificConfig` is for the overrides.
Since the `changelog_*` were only in the package specific config I assumed that those could not be set as the default (Workspace) level, but if they should behave the same, maybe we could have a single (flatter) struct?

For the common object I did so as it had been done the same way on the `Workspace` struct so I thought I was following the existing patterns 🤣

PS: No problem if I do end up reverting things, happy to see you are challenging changes!

MarcoIeni: > My understanding is that PackageConfig is the default configuration for packages and the PackageSpecificConfig is for the overrides.

This is also my understanding (I wrote the code a while back!).

> Since the changelog_* were only in the package specific config I assumed that those could not be set as the default (Workspace) level

You should be able to set changelog config at the workspace level. See docs here:
https://github.com/MarcoIeni/release-plz/blob/986bbc4c187258eb644f6b6361e3d25168975550/website/docs/config.md#L50

> maybe we could have a single (flatter) struct?

That's my goal, yes. I want to do a refactoring and remove many of those nested structs. But let's do it in another PR. 

> happy to see you are challenging changes!

Same for me! :)

AntoineGS: Sounds good, I will make the change to match what `changelog_config` does and do another PR for the structs!
PS: Finally getting back to working on this ;)

MarcoIeni: I'm happy to hear you're back and feeling better 🎉

MarcoIeni:
In https://github.com/MarcoIeni/release-plz/pull/1182 I implemented the refactoring of the configuration so that now the struct is flatten and better reflects the structure of the config itself. I'm curious to know what you think about the code now :)
Also, I'm happy to solve the merge conflicts so you can focus on more interesting work :)

AntoineGS:
That was a lot easier to integrate the new option, awesome! I've made the changes (refractor) I just have not tested the feature with it plus the merges. Should be ready soon.

AntoineGS:
Seems to still work fine, ready for review/merge!

crates/release_plz_core/src/next_ver.rs
```diff
@@ -1174,4 +1218,12 @@ mod tests {
         .unwrap();
         assert_eq!(old, new)
     }
+
+    #[test]
+    fn project_new_no_release_will_error() {
+        let local_manifest = Path::new("../fake_package/Cargo.toml");
+        let result = get_project(local_manifest, None, HashSet::default(), false);
+        assert!(result.is_err());
+        assert_eq!(result.unwrap_err().to_string(), ERR_NO_PUBLIC_PACKAGE);
```
MarcoIeni, line 1227: nice test and mocking strategy ⭐

MarcoIeni:
Great work, thanks! 🙌

PR #1101, opened 2023-11-29
feat: new generate-schema command to generate a JSON schema for the configuration

AntoineGS:
- New generate-schema command
- Updated the documentation
- Added the first iteration of the schema to then create a PR on SchemaStore once merged into main
- New CI check that there are no uncommitted changes to the schema in the PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,22 @@ jobs:
       - name: Clippy check
         run: cargo clippy --all-targets --all-features --workspace -- -D warnings
 
+  schema:
+    name: Schema
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - name: Run release-plz generate-schema
+        run: cargo run --package release-plz --bin release-plz generate-schema
+      - name: Check schema
+        run: |
+          gitdiff=$(git status .schema/latest.json | grep "modified:")
+          test -z "$gitdiff" && (exit 0) || (exit 1)
+
   docs:
     name: Docs
     runs-on: ubuntu-22.04
--- a/crates/release_plz/src/generate_schema.rs
+++ b/crates/release_plz/src/generate_schema.rs
@@ -2,7 +2,7 @@ use crate::config;
 use schemars::schema_for;
 use std::fs;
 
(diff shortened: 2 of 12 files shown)


AntoineGS:
Alright, finally got the CI down! Ready for review/merge.

.github/workflows/ci.yml
```diff
@@ -100,6 +100,27 @@ jobs:
...
+        run: mv .schema/latest.json .schema/latest_orig.json
+      - name: Run release-plz generate-schema
+        run: cargo run --package release-plz --bin release-plz --release generate-schema
+      - name: Check schema
+        uses: LouisBrunner/diff-action@v0.2.0
+        with:
+          old: .schema/latest_orig.json
+          new: .schema/latest.json
+          mode: addition
+          tolerance: same
```
MarcoIeni, line 122: I was thinking..we could have a cargo test to do this check instead of adding this CI job 🤔

MarcoIeni: Apart from this minor point, this PR looks perfect, thanks 🙏 
I will test this asap.

crates/release_plz/src/generate_schema.rs
```diff
@@ -0,0 +1,23 @@
...
+    let schema = schema_for!(config::Config);
+    let mut json = serde_json::to_string_pretty(&schema).unwrap();
+    // As of now, Schemars does not support the $id field, so we insert it manually.
+    // See here for update on resolution: https://github.com/GREsau/schemars/issues/229
+    json = json.replace(
+        SCHEMA_TOKEN,
+        &format!("{}\n  {}{}{}\",", SCHEMA_TOKEN, ID, FOLDER, FILE),
+    );
+    fs::create_dir_all(FOLDER).unwrap();
+    fs::write(format!("{}{}", FOLDER, FILE), json).unwrap();
```
MarcoIeni, line 22: can you extract `format!("{}{}", FOLDER, FILE)` into a variable since you use it also at line 19?

.github/workflows/ci.yml
AntoineGS: Yeah.. that would have been a lot easier!
I will make the change, plus it should be quicker to run.

AntoineGS:
I made the changes, ready for round two!
I did end up scrapping the code used in one of the three comments, as I found out about the Path struct.
It plays nicer with changing of OSes versus manually built paths.

MarcoIeni (approved):

AntoineGS:
🎉
Thanks for the review, now I know about the anyhow::context!
I will work on getting a PR on ShemaStore to add this, probably over the weekend.

PR #1143, opened 2023-12-16
feat: return error if tag exists and package isn't published

MarcoIeni:
Closes https://github.com/MarcoIeni/release-plz/issues/1139

### DIFF
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -745,7 +745,7 @@ impl Updater<'_> {
             .git_tag(&package.name, &package.version.to_string());
         let tag_commit = repository.get_tag_commit(&git_tag);
         if tag_commit.is_some() {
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/next_ver.rs
```diff
@@ -740,12 +740,18 @@ impl Updater<'_> {
...
-                .project
-                .git_tag(&package.name, &package.version.to_string());
-            repository.get_tag_commit(&git_tag)
-        };
+        let git_tag = self
+            .project
+            .git_tag(&package.name, &package.version.to_string());
+        let tag_commit = repository.get_tag_commit(&git_tag);
+        if tag_commit.is_some() {
+            let registry_package = registry_package.with_context(|| format!("package `{}` not found in the registry, but tag {git_tag} exists. Consider running `cargo publish` manually to publish this package.", package.name))?;
```
MarcoIeni, line 748:
```suggestion
            let registry_package = registry_package.with_context(|| format!("package `{}` not found in the registry, but the git tag {git_tag} exists. Consider running `cargo publish` manually to publish this package.", package.name))?;
```

PR #1168, opened 2023-12-29
feat(next_version): support configuration

orhun:
This PR makes it possible to configure the `next_version` crate to customize the behavior for cases like https://github.com/orhun/git-cliff/issues/343

e.g.:

```rust
let next_version = semver
	.next(
		vec![],
		Some(NextVersionConfig {
			uncontrolled_minor_bump: false,
			initial_major_increment: false,
		}),
	)
	.to_string();
```

A couple of points:

- This is a breaking change due to the newly added `config` parameter to the `next` function. Let me know if you are fine with it or not and we can probably think of a way to make it non-breaking. I think it is fine though.
- I'm not sure about the names of the config variables.
- I tried this out with `git-cliff` directly and it seems to work as expected. However, I must say it was really hard to wrap my head around the conditions so I might have missed something. Please double check.

:bear:


crates/next_version/src/config.rs
```diff
@@ -0,0 +1,26 @@
+/// Configuration for version incrementation rules.
+#[derive(Debug)]
+pub struct NextVersionConfig {
+    /// Indicates whether uncontrolled minor version bumps are enabled.
+    /// When false, minor version increments are not performed and they
+    /// should be manually done to signal higher API stability to the user.
+    ///
+    /// Default value is `false`.
```
MarcoIeni, line 8:
```suggestion
    /// Default value is `true`.
```
I see that the default is true.

MarcoIeni:
I would prefer a non-breaking change. Let's keep the default behavior as it is.
Let's create another struct and use the builder pattern to configure it.
I will write an example now.

crates/next_version/src/config.rs
orhun: This comment was wrong, updated.

crates/next_version/src/config.rs
```diff
@@ -0,0 +1,26 @@
+/// Configuration for version incrementation rules.
+#[derive(Debug)]
+pub struct NextVersionConfig {
+    /// Indicates whether uncontrolled minor version bumps are enabled.
+    /// When false, minor version increments are not performed and they
+    /// should be manually done to signal higher API stability to the user.
```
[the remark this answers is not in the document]
orhun: Actually no, it is for `0.1.0` -> `0.2.0` for feature commits.

orhun:
Added tests.

MarcoIeni:
```rust
struct VersionUpdater {
    pub uncontrolled_minor_bump: bool,
    pub initial_major_increment: bool,
}

impl VersionUpdater {
    // builder lite pattern https://matklad.github.io/2022/05/29/builder-lite.html
    fn with_uncontrolled_minor_bump(mut self, uncontrolled_minor_bump) -> Self { self.uncontrolled_minor_bump = uncontrolled_minor_bump; self }
    
    fn next(&self, version: Version, commits: I) -> Version { }
}
```
Then Version::next calls VersionUpdater with the default options.

orhun:
I didn't get it, do you wanna define another struct called `VersionUpdater` and have the configuration values there?

crates/next_version/src/config.rs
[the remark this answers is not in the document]
MarcoIeni: oh, ok, the comment doesn't mention feature commits, so I didn't get it.

> minor version increments are not performed and they
> should be manually done to signal higher API stability to the user.

This is only true if the version is 0.0.x, right?

Maybe the name should be `features_always_increment_minor`?

crates/next_version/src/config.rs
```diff
@@ -0,0 +1,26 @@
...
+    ///
+    /// Default value is `false`.
+    pub uncontrolled_minor_bump: bool,
+
+    /// Indicates whether initial major version increments are enabled.
+    /// When true, allows for a major version increment from 0 to 1,
+    /// indicating a breaking change in the API.
+    ///
+    /// Default value is `false`.
+    pub initial_major_increment: bool,
```
MarcoIeni, line 16: I think this could be named `breaking_always_increment_major`

MarcoIeni:
> do you wanna define another struct called VersionUpdater and have the configuration values there?

yes

> How about config::NextVersionConfig?

Let's delete it. The VersionUpdater has the default configuration, which you can override with the builder pattern.

orhun:
Sorry, I'm not following. Do you want to achieve the following API?

```rust
let next_version = semver
	.next(
		vec![]
	)
    .with_uncontrolled_minor_bump(true)
    .with_initial_major_increment(true)
	.to_string();
```

orhun:
Or is it something like this?

```rust
VersionUpdater::new(semver)
    .with_uncontrolled_minor_bump(true)
    .with_initial_major_increment(true)
    .next(vec![])
```

MarcoIeni:
```rust
VersionUpdater::new()
    .with_uncontrolled_minor_bump(true)
    .with_initial_major_increment(true)
    .next(semver, vec![])
```
I was thinking this what do you think?

orhun:
Sure, makes sense. Implemented in d3d3fb07cec3e519999dd43428a4ffc4171e33ef

PR #1201, opened 2024-01-11
feat: use github graphql api for commits

### LINKED ISSUE #1200: Signing commits made by release-plz bot
## Motivations

Hey, I've seen the #956 which added the `Signed-off-by: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>` to the commit message. This however doesn't alone result in `verified` status on github. We have a repository setting which requires all commits to be signed and thus we cannot merge the release PR.

For comparison bot's commit and my commit
image

And the merge button is red with a message:

> Merging is blocked
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

Is there a possibility for getting a verified status for release-plz bot?

- Would you like to implement this feature? [y/n]

## Solution

It looks like the dependabot has a power to get verified status, but I'm not sure if it isn't kinda specially supported by github.

image

zvolin:
closes #1200.

This PR adds a new flag '--api-commit' to `release-pr` command. If present, a release commit will not be created in local copy and pushed. Instead the changes will be encoded with base64 or just marked as deleted and the commit will happen using [github's graphql api](https://docs.github.com/en/graphql), namely `createCommitOnBranch`.

The main benefit is that the commits created using the graphql api will be associated with the token owner. This allows for better flexibility in who is the author, allowing things like getting [Verified](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) status.

This is currently an mvp. Let me know if you like the idea and I can continue working on docs and tests and polish what's needed.

### DIFF
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -18,7 +18,6 @@ fn get_graphql_endpoint(remote: &Remote) -> Url {
     base_url
 }
 
-/// TODO: add tests
 /// Commit all the changes (except typestates) that are present in the repository
 /// using GitHub's [GraphQL api](https://docs.github.com/en/graphql).
 pub async fn commit_changes(client: &GitClient, repo: &Repo, message: &str) -> Result<()> {
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -20,16 +20,20 @@ fn get_graphql_endpoint(remote: &Remote) -> Url {
 
 /// Commit all the changes (except typestates) that are present in the repository
 /// using GitHub's [GraphQL api](https://docs.github.com/en/graphql).
-pub async fn commit_changes(client: &GitClient, repo: &Repo, message: &str) -> Result<()> {
+pub async fn commit_changes(
+    client: &GitClient,
+    repo: &Repo,
+    message: &str,
+    branch: &str,
+) -> Result<()> {
     let owner_and_repo = format!("{}/{}", client.remote.owner, client.remote.repo);
-    let branch = repo.get_current_branch()?;
     let current_head = repo.current_commit_hash()?;
     let deletions = removed_files(repo)?;
     let changes = changed_files(repo)?;
 
     let commit_query = format_commit_query(
         &owner_and_repo,
-        &branch,
+        branch,
         message,
         &current_head,
         &deletions,
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -19,7 +19,7 @@ fn get_graphql_endpoint(remote: &Remote) -> Url {
 }
 
 /// Commit all the changes (except typestates) that are present in the repository
-/// using GitHub's [GraphQL api](https://docs.github.com/en/graphql).
+/// using GitHub's [GraphQL api](https://docs.github.com/en/graphql/reference/mutations#createcommitonbranch).
 pub async fn commit_changes(
     client: &GitClient,
     repo: &Repo,
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -201,7 +201,7 @@ mod tests {
         fs::write(&changed_path, b"changed").await.unwrap();
 
         let removed = "removed.txt";
-        let removed_path = repo_dir.join("removed.txt");
+        let removed_path = repo_dir.join(removed);
         fs::write(&removed_path, b"removed").await.unwrap();
 
         let type_changed_path = repo_dir.join("type_changed.txt");
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -22,7 +22,7 @@ fn get_graphql_endpoint(remote: &Remote) -> Url {
 /// using GitHub's [GraphQL api](https://docs.github.com/en/graphql).
 pub async fn commit_changes(client: &GitClient, repo: &Repo, message: &str) -> Result<()> {
     let owner_and_repo = format!("{}/{}", client.remote.owner, client.remote.repo);
-    let branch = repo.current_branch()?;
+    let branch = repo.get_current_branch()?;
     let current_head = repo.current_head()?;
     let deletions = removed_files(repo)?;
     let changes = changed_files(repo)?;
--- a/crates/release_plz_core/src/git/github_graphql.rs
+++ b/crates/release_plz_core/src/git/github_graphql.rs
@@ -177,3 +177,102 @@ async fn format_additions(
 
     Ok(additions)
 }
+
+#[cfg(test)]
+mod tests {
+    use tempfile::tempdir;
+
+    use super::*;
+
+    use crate::copy_dir::create_symlink;
+
+    #[tokio::test]
+    async fn github_commit_query() {
+        let temporary = tempdir().unwrap();
+        let repo_dir = temporary.as_ref();
+        let repo = Repo::init(repo_dir);
+
+        // make the initial commit on top which we'll make changes
+        let unchanged_path = repo_dir.join("unchanged.txt");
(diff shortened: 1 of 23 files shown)


zvolin:
I've managed to successfully create this PR https://github.com/zvolin/test-release-plz/pull/3

MarcoIeni:
Sorry if I haven't reviewed this yet, I will get to it asap!

zvolin:
Thanks for checking :+1: 

As I mentioned above, I haven't fully committed into the change yet and wanted to first know your thoughts about approach. If you like it.

MarcoIeni:
thanks for digging into this. What's the con of removing this flag and making this the default behavior?

zvolin:
I haven't checked gitea and gitlab api for equivalents but it's likely it's github specific

MarcoIeni:
> I haven't checked gitea and gitlab api for equivalents but it's likely it's github specific

yes, I mean that we could use this approach with the GitHub backend.

> Also I haven't used any graphql Rust crate because they require downloading and hardcoding whole graphql Schema so that all types are generated. I thought for a single api hit, the string manipulation will be simpler

Yes, that's totally fine :)

zvolin:
hmmm if we gonna use it for github by default, I would need to check what happens if someone tries that with the default GITHUB_TOKEN on CI

zvolin:
so should I proceed with the impl?

zvolin:
Linking the previous example commit because I was uncareful and the PR was force-pushed with the regular release-plz one.
The commit created by this change is there: https://github.com/zvolin/test-release-plz/commit/23053bfbd061c8cb1ec501a357e36f105d75e75e

zvolin:
Also worth noting that the verified status doesn't depend on owner's GPG key (how could it) and it is rather verified by github's (some kind of one-time?) signature:
image

zvolin:
> hmmm if we gonna use it for github by default, I would need to check what happens if someone tries that with the default GITHUB_TOKEN on CI

This actually looks good, everything seems to work: https://github.com/zvolin/test-release-plz/pull/4

zvolin:
I think we can make it a default

MarcoIeni:
> so should I proceed with the impl?

I just want to make sure that the graphql api is the simplest way to obtain the "verified" badge. 🤔 
I would like to find out if there's a way of obtaining it by committing from the cli, which is simpler, but you probably researched this already 👍

zvolin:
Hmmm yes I was researching that however I'll appreciate double check. Here is [some write-up (https://github.com/Nautilus-Cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md) from 3y ago.
Also https://github.com/google-github-actions/release-please-action/issues/399.

In short, I think the alternative to API is to have a second github account for bot with gpg key set and to also pass the signing key as a secret to the workflow, configure the author and email to that second account and then commit. I haven't researched the github apps tho

MarcoIeni:
I did some research and I think your approach is the best one because it avoids storing GPG secrets in the repository.
If it works well, I would like to make it the default for the github backend.

MarcoIeni:
I don't think there should be support for the old behavior unless people complain and they provide a reason.

We could update the release-pr command docs, yes.

For tests it's difficult, because we don't test github api.

zvolin:
> There's also a small merge conflict right now. Can you resolve it?

Sure.

> For tests it's difficult, because we don't test github api.

I'll add some unit tests then and call it a day.

I'm not sure I can get this ready over the weekend as I'll be in travel, we'll see.

zvolin:
I removed the API_ENDPOINT constant in favor of taking base_url of the GitClient. I thought it might be useful at some point due to the remote mock in existing tests. Maybe not even useful but if someone in future tries to write some tests will not be surprised that he hits github api

zvolin:
Sooo....
I wanted to write some tests and I gave up. I had a test for this newly added `get_graphql_endpoint` that checks if it fails for other git upstreams and I wanted to write one which checks for correct inclusion of unchanged, changed, type_changed, added and deleted files in repo. However, even the output of stuff like `format_additions` cannot be parsed as a json (graphql doesn't quote key names, like `path: "foo"`) and the test ended up being really cumbersome to the point I think it's not worth it

zvolin:
actually after writing that I noticed that the domain check in get_graphql_endpoint defeats it's purpose so removed also that

zvolin:
Updated the docs but please check those carefully, English is not one of my advantages :upside_down_face: This is ready by me

zvolin:
> Do you expect someone to miss the old behavior for any reason?

just noticed I haven't answered that. So I think I've seen some complaints that the changes were so big that encoding them as base64 and pushing was overflowing some (i think 10mb but can't be sure as don't remeber source) constrain on graphql api. However, I don't expect that to happen with `release-plz release-pr`.

crates/release_plz_core/src/git/github_graphql.rs
```diff
@@ -0,0 +1,169 @@
...
+              deletions: {deletions},
+              additions: {additions}
+            }}
+          }}) {{
+            commit {{
+              author {{
+                name,
+                email
+              }}
+            }}
```
MarcoIeni, line 112: Is this the result returned by the mutation? If yes, do we need it?
I don't think we are reading it.

crates/release_plz_core/src/git/github_graphql.rs
```diff
@@ -0,0 +1,168 @@
...
+
+// format a graphql query to create commit on branch
+async fn format_commit_query(
+    owner_and_repo: &str,
+    branch: &str,
+    message: &str,
+    current_head: &str,
+    deletions: &[impl AsRef<Path>],
+    additions: &[impl AsRef<Path>],
+    repo_dir: impl AsRef<Path>,
```
MarcoIeni, line 88: can you collect all these parameters into a struct? E.g. `GitHubCommit`

crates/release_plz_core/src/git/github_graphql.rs
```diff
@@ -0,0 +1,168 @@
...
+    repo.changes(|line| !line.starts_with("T ") && !line.starts_with("D "))
+}
+
+// get the list of removed files in repository
+fn removed_files(repo: &Repo) -> Result<Vec<String>> {
+    repo.changes(|line| line.starts_with("D "))
+}
+
+// format a graphql query to create commit on branch
+async fn format_commit_query(
```
MarcoIeni, line 81: Can you add a unit test for this where you test that the correct query is returned?
No need to send the query over the network, let's just test that the query is constructed correctly.

For the directories deletions and additions, you could create some files in a temporary directory like we do [here](https://github.com/zvolin/release-plz/blob/32bd7aa6e99475d14a29fb7766076a817b5a331d/crates/release_plz_core/src/copy_dir.rs#L108)

MarcoIeni (changes requested):
I did some small refinements, but overall looks great! 🚀
I just requested a few minor changes. 🙏

crates/release_plz_core/src/git/github_graphql.rs
zvolin: it is a selection of some data from what the mutation returns. In graphql you must have some selection of fields up to some primitives. Eg.
```
commit {{ }}
```
would complain that you didn't select any subfields of commit.

It is not read anywhere but there must be some selection.

crates/release_plz_core/src/git/github_graphql.rs
zvolin: done

crates/release_plz_core/src/git/github_graphql.rs
zvolin: done

crates/release_plz_core/src/git/github_graphql.rs
MarcoIeni: oh, right! Sounds good then :)

MarcoIeni:
image
Just tested it and got the juicy "verified" badge. Thanks for reporting the issue, researching how to solve it, submitting a POC, and refining the PR after seeking feedback. Awesome job! 🙌

MarcoIeni (approved):
Thanks ❤️

zvolin:
Thanks too :)

Issue #1204, opened 2024-01-14
Error during `release-plz release-pr` after initial release


dousto:
## Bug description

First of all, excellent work on this well done project.  I have set it up through a GitHub action to release my project https://github.com/dousto/redact-composer.

After committing the action, I saw my project successfully published to crates.io (which actually surprised me a bit since I was expecting to have to merge a PR, but maybe that's just a quirk of releasing a brand new crate).

Additionally a new PR was created (https://github.com/dousto/redact-composer/pull/8), which I merged since it had the new changelog.

From this point on, (including the merged PR commit), the github action for `release-plz release-pr` [begins failing](https://github.com/dousto/redact-composer/actions/runs/7520985190/job/20471269733) with:
```
  2024-01-14T18:25:25.648604Z ERROR  cannot revert changes introduced when comparing packages

Caused by:
    error while running git with args `["checkout", "/tmp/.tmpqi157G/redact-composer/Cargo.lock"]:
    - stderr: error: pathspec '/tmp/.tmpqi157G/redact-composer/Cargo.lock' did not match any file(s) known to git
```

I tried adding another commit but that run also ran into the same issue.

I think I have somehow gotten myself into an invalid state here, but I can't seem to pinpoint exactly where it's gone wrong and how to dig myself out of it to resume normal behavior. Maybe merging the PR was a mistake?

Any help would be appreciated. In the meantime I'll be attempting to debug.

- Would you like to work on a fix? Not yet sure if it's a bug

## To Reproduce

```
git clone https://github.com/dousto/redact-composer.git
cd redact-composer
release-plz release-pr --git-token dummy
```

## Expected behavior

Expected subsequent `release-plz` github actions to run successfully after initial publish, and merge of PR.

## Environment

(in GitHub action environment)
- OS: linux-x64
- release-plz version: 0.3.38

dousto:
Ok, just a progress update of my debugging efforts. In this particular repo there *was* a prior state in the commit history where the `Cargo.lock` was not committed and I think that might be part of (if not entirely) the issue. Will further update as I get a more complete picture.

dousto:
After some more debugging I did confirm that the prior commit history where my `Cargo.lock` did not exist was causing an issue. I created PR #1205 to fix.

MarcoIeni:
closed by https://github.com/MarcoIeni/release-plz/pull/1205

thanks for the bug report and fix. I will release a new version of release-plz and github action in the following days!

[...]

PR #1205, opened 2024-01-15
fix: Prevent error if Cargo.lock doesn't exist during portions of commit history

dousto:
This PR addresses #1204.

Before this change, when traversing prior commits during an update diff, if the history started with `Cargo.lock` existing but at some point during iteration reaches a commit where the `Cargo.lock` file did not exist, an error would be encountered. Since the `Cargo.lock` path is only needed to revert to a clean state during loop iteration, I think this is a safe change to make, however let me know if there may be any edge cases I hadn't thought about.


MarcoIeni (approved):

MarcoIeni:
thanks!

PR #1217, opened 2024-01-20
fix: Correct dependency update behavior for release-pr

dousto:
Changed the update command to use cargo metadata from the tmp manifest, which give it the correct dependency paths used for comparison during dependency updates.

### DIFF
--- a/crates/release_plz_core/src/command/update.rs
+++ b/crates/release_plz_core/src/command/update.rs
@@ -131,6 +131,8 @@ pub fn update(input: &UpdateRequest) -> anyhow::Result<(PackagesUpdate, TempRepo
         crate::next_versions(input).context("failed to determine next versions")?;
     let local_manifest_path = input.local_manifest();
     let local_metadata = cargo_utils::get_manifest_metadata(local_manifest_path)?;
+    // Read packages from `local_metadata` to update the manifest of local
+    // workspace dependencies.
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/command/update.rs
```diff
@@ -130,8 +130,8 @@ pub fn update(input: &UpdateRequest) -> anyhow::Result<(PackagesUpdate, TempRepo
     let (packages_to_update, repository) =
         crate::next_versions(input).context("failed to determine next versions")?;
     let local_manifest_path = input.local_manifest();
-    let all_packages: Vec<Package> =
-        cargo_utils::workspace_members(input.cargo_metadata())?.collect();
+    let local_metadata = cargo_utils::get_manifest_metadata(local_manifest_path)?;
+    let all_packages: Vec<Package> = cargo_utils::workspace_members(&local_metadata)?.collect();
```
MarcoIeni, line 134:
```suggestion
    // Read packages from `local_metadata` to update the manifest of local
    // workspace dependencies.
    let all_packages: Vec<Package> = cargo_utils::workspace_members(&local_metadata)?.collect();
```

MarcoIeni:
Thanks for identifying this bug and fixing it!
Errors like this shouldn't happen. Ideally, we should have a test for this :/
Unfortunately, I didn't have the time to add more tests.

If anybody wants to add a test case for this, PR welcome!

Issue #1233, opened 2024-01-25
Failed to determine next versions (rust nightly involved?)


fasterthanlime:
Here's part of a log of https://github.com/hapsoc/fluke/actions/runs/7656751643/job/20865803972 :

```
Updating crates.io index
note: Downloading fluke 0.1.0
 Downloading crates ...
  Downloaded fluke v0.1.0
note: Cloning into "/tmp/.tmpRzFkIx"
note: Downloading fluke-buffet 0.1.0
 Downloading crates ...
  Downloaded fluke-buffet v0.1.0
note: Cloning into "/tmp/.tmpRzFkIx"
note: Downloading fluke-maybe-uring 0.1.1
 Downloading crates ...
  Downloaded fluke-maybe-uring v0.1.1
note: Cloning into "/tmp/.tmpRzFkIx"
note: Downloading fluke-hpack 0.3.0
 Downloading crates ...
  Downloaded fluke-hpack v0.3.0
info: syncing channel updates for 'nightly-2023-10-02-x86_64-unknown-linux-gnu'
info: latest update on 2023-10-02, rust version 1.75.0-nightly (e0d7ed1f4 2023-10-01)
info: downloading component 'cargo'
info: downloading component 'llvm-tools'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: installing component 'cargo'
info: installing component 'llvm-tools'
info: installing component 'rust-std'
info: installing component 'rustc'
  2024-01-25T15:29:06.337733Z ERROR  failed to update packages

Caused by:
    0: failed to determine next versions
    1: failed to retrieve difference between packages
Error: failed to update packages

Caused by:
    0: failed to determine next versions
    1: failed to retrieve difference between packages
    2: cannot compare packages
    3: cannot determine packaged files of local package "/tmp/.tmpbcoUN1/fluke/crates/fluke"
    4: error while running `cargo package`: info: syncing channel updates for 'nightly-[202](https://github.com/hapsoc/fluke/actions/runs/7656751643/job/20865803972#step:4:208)3-10-02-x86_64-unknown-linux-gnu'
       info: latest update on 2023-10-02, rust version 1.75.0-nightly (e0d7ed1f4 2023-10-01)
       info: downloading component 'cargo'
       info: downloading component 'llvm-tools'
       info: downloading component 'rust-std'
       info: downloading component 'rustc'
       info: installing component 'cargo'
       info: installing component 'llvm-tools'
[...]

PR #1235, opened 2024-01-26
Don't rely on stderr to check for success

fasterthanlime:
Have run-cargo check for the exit code instead,
show stderr if it was nonzero.

Closes #1203 and #1233

(Also, is a duplicate of #1234 I've just noticed, woops)


MarcoIeni:
Thanks for opening this. 🙏 I'm merging my version because I prefer to check the status code in the caller 👍

PR #1251, opened 2024-02-05
fix: update local dependencies specified in the workspace manifest

zvolin:
We've encountered issues when [using release-plz at lumina](https://github.com/eigerco/lumina/actions/runs/7783144141/job/21220976044#step:5:225). It seems like currently `release-plz` is capable of updating dependencies specified directly in packages' manifests and is ignoring dependencies specified on workspace level. The latter doesn't get version bumps which results in failures in resolving.

I'm not sure if there are additional things needed when updating also workspace-wide dependencies tho. Let me know wdyt


MarcoIeni:
Nice catch, thanks!
I didn't even think about specifying local dependencies at the workspace level :)

PR #1255, opened 2024-02-06
feat: create annotated tags instead of lightweight

### LINKED ISSUE #1215: release-plz and it's tests hangs indefinitely if automatic signing of tags is configured
## Bug description

`release-plz release` as well as `cargo test` hangs if someone has enabled automatic signing of tags.

From what I see `release-plz` tries to create lightweight tags. If someone has automatic signing, then git will try to create an annotated tag instead and the `git_cmd` will hang waiting for the tag's message. Git itself advertises lightweight tags as only to be used locally during the development and to use the annotated tags everywhere else, so it might not be bad idea to just slap there some message like: 'release {tag-name}' and have the annotated tag instead?

From `git help tag`:

> Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG signature.
> Whereas a "lightweight" tag is simply a name for an object (usually a commit object).

Also adding `-m` implies `-a` so bare `-m "message"` would work.

## To Reproduce

add following to the `~.gitconfig`

```
[user]
  ...
  signingkey = SIGNINGKEY
[tag]
  gpgSign = true
```

## Expected behavior

The tag should be created.

## Environment

- OS: archlinux 6.6.10-zen
- release-plz version: 0.3.39

zvolin:
fixes #1215


MarcoIeni:
cargo-release creates annotated tags 👍 
https://github.com/crate-ci/cargo-release/blob/d5a7288f8c438f00341a14fadccb460539e590af/src/ops/git.rs#L163

MarcoIeni:
I tested this on https://github.com/MarcoIeni/rust-workspace-example/commit/bd517f1ced67feeed564e83f4869adc4f33a63f6
and it seems it's working as expected :)
image

Issue #677, opened 2023-04-08
Configuration: add ability to change the git release name.


MarcoIeni:
## Motivations

The release now is called `{crate_name}v{tag}` if workspace or `v{tag}` if single crate.
Add ability to customize the name.

## Solution

In the `git_release` section of the configuration add something like this:

```toml
[package.my_pkg.git_release]
# Set an alternate name format for the git release
# Default: v${tag}.
name = "${tag}" # do not include `v`
```

Let's use `tera` for this since we already expose it to the user via git-cliff.

LukeMathWalker:
I'm currently using the following configuration for `cargo-release`:

```toml
# Config for 'cargo release'
[workspace.metadata.release]
shared-version = true
consolidate-commits = true
tag-name = "{{version}}"
```

It lets me release all packages in my workspace with a single version and it generates a single tag (e.g. `0.1.9`).d
I was trying to migrate my release flow to `release-plz` but I can't seem to find an equivalent setting?

(Commenting here since it looks like the most appropriate issue)

MarcoIeni:
Are you using a workspace version? Docs [here](https://doc.rust-lang.org/cargo/reference/workspaces.html)
![image](https://github.com/MarcoIeni/release-plz/assets/11428655/bf0bd167-2a7e-4edd-95b5-7abf500de537)

Release-plz supports workspace versions, but it will create one tag for each package. But having a single tag for the entire workspace is something interesting I want to support. Let's understand together how!

https://github.com/MarcoIeni/release-plz/issues/1023 sounds related to your use case as well.

MarcoIeni:
for the tag-name template, we have https://github.com/MarcoIeni/release-plz/issues/680
I didn't get why we would need `consolidate-*` 🤔 

The way we could do it is by using a workspace version and the following release-plz.toml config file:

```toml
[workspace]
# Disable git releases by default
git_release_enable = false

# Disable git tags by default
git_tag_enable = false

# Options for the package I care the most
[[package]]
name = "my_main_package"
# Customize the git tag name to remove the `my_main_package` prefix.
# This is not implemented yet: https://github.com/MarcoIeni/release-plz/issues/680
git_tag_name = "v${version}" 

# Customize the git release name to remove the `my_main_package` prefix.
# This is not implemented yet. This is the issue to track it.
git_release_name = "v${version}" 

# Enable git releases for this package
git_release_enable = true

# Disable git tags for this package
git_tag_enable = true
```

[...]

PR #1260, opened 2024-02-11
feat: custom git-release-name by tera template. #677

giangndm:
This PR add custom git release name feature (https://github.com/MarcoIeni/release-plz/issues/677) :

- setting global by git-release-name field in workspace
- override in each project by git-release-name field

Default:

- `{{ package }}-{{ version }}` for workspace 
- `v{{ version }}` for single project 

Ps: Only Project.release_name is tested, maybe we need some way to test it but I don'f familiar with GitTea for adding more tests.

### DIFF
--- a/.schema/latest.json
+++ b/.schema/latest.json
@@ -42,6 +42,7 @@
         "dependencies_update": null,
         "git_release_draft": null,
         "git_release_enable": null,
+        "git_release_name": null,
         "git_release_type": null,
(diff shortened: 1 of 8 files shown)


MarcoIeni:
> maybe we need some way to test it but I don'f familiar with GitTea for adding more tests.

I added an integration test 👍

MarcoIeni (approved):

PR #1272, opened 2024-02-13
chore: avoid `git checkout` when visiting git history

MarcoIeni:
We can retrieve all logs at once with `git log -10 --pretty=format:///%B///%H/// -- /path/to/crate`
with this approach we can retrieve git commits in batches (in this example of 10).
We don't need to do `git checkout`.
`///` is a separator. We can use something more specific, such as `@@release@plz@@` to avoid conflicts with git commit messages.

With this, we don't need to copy to a temporary directory when doing `release-plz update`.

- [x] check git history of crates in parallel? Answer: not possible for now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,7 +326,7 @@ dependencies = [
  "ignore",
  "im-rc",
  "indexmap",
- "itertools",
+ "itertools 0.11.0",
  "jobserver",
  "lazycell",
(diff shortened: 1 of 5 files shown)


MarcoIeni:
this approach doesn't work. We need to do `git checkout` because we need to compare the content of the directory

PR #1287, opened 2024-02-22
feat: add `all-static` feature

hack3ric:
This adds the ability to build against system libraries like libgit2 and OpenSSL, when the feature is not present.

### DIFF
--- a/crates/release_plz/Cargo.toml
+++ b/crates/release_plz/Cargo.toml
@@ -14,13 +14,14 @@ categories = ["command-line-utilities", "development-tools"]
 workspace = true
 
 [features]
-default = ["docker-tests"]
+default = ["docker-tests", "all-static"]
 # Run tests that need a docker runtime to be executed
 docker-tests = []
+all-static = ["release_plz_core/all-static"]
 
 [dependencies]
 git_cmd = { path = "../git_cmd", version = "0.4.20" }
-release_plz_core = { path = "../release_plz_core", version = "0.19.0" }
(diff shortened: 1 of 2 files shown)


orhun:
The use case is <https://gitlab.archlinux.org/archlinux/packaging/packages/release-plz/-/merge_requests/1>

MarcoIeni:
Thanks for the contribution! I'm happy to accept it. I'll think a bit about the name of the feature.

So on arch linux, by default the release-plz binary will use the system libraries such as git2 and openssl, right?
Is this a policy of all arch linux packages?

crates/release_plz/Cargo.toml
```diff
@@ -14,13 +14,14 @@ categories = ["command-line-utilities", "development-tools"]
 workspace = true
 
 [features]
-default = ["docker-tests"]
+default = ["docker-tests", "cargo-vendored-libs"]
```
MarcoIeni, line 17: let's call this feature `all-static`, to follow `cargo` convention :)

hack3ric: Fixed.

MarcoIeni:
I see, thanks for the explanation, it makes sense :)

By default, I vendor everything so people have fewer issues when running the binary in all various systems. But package maintainers control the dependencies of the binary, so it makes sense to not vendor the other libraries.

Thanks for the contribution 🙏

MarcoIeni (approved):

PR #1292, opened 2024-02-25
fix: an issue with modified files due to lfs

### LINKED ISSUE #1186: Can't checkout to head after calculating diff
## Bug description

I am trying to integrate `release-plz` into the [`rattler`](https://github.com/mamba-org/rattler) project. You can see my configuration from [this commit](https://github.com/orhun/rattler/commit/87525f3ec2442df0187d5a41e9edd083e4a20c3e).

However, I'm getting the following CI error during testing:

```
note: Downloading rattler_shell 0.14.0
 Downloading crates ...
  Downloaded rattler_shell v0.14.0
note: Cloning into "/tmp/.tmpNZmJeJ"
note: Downloading tools 0.0.0
 Downloading crates ...
  Downloaded tools v0.0.0
  2024-01-04T10:06:11.179251Z ERROR  can't checkout to head after calculating diff
Error: can't checkout to head after calculating diff

Caused by:
    error while running git with args `["checkout", "main"]:
    - stderr: error: Your local changes to the following files would be overwritten by checkout:
    	test-data/channels/dummy/linux-64/repodata.json
    Please commit your changes or stash them before you switch branches.
    Aborting
```

Does that mean a file is changing during the calculation of the next release? I think this shouldn't happen.

- Would you like to work on a fix? y

## To Reproduce

1. Clone [`rattler`](https://github.com/mamba-org/rattler)
2. Integrate `release-plz` as mentioned before.

## Expected behavior

Successful CI build.

## Environment

- OS: GitHub Actions
- release-plz version: `MarcoIeni/release-plz-action@v0.5`

## Additional context

See https://github.com/orhun/rattler/actions/runs/7408098693/job/20155636867

baszalmstra:
This is an attempt to fix #1186 . 

I haven't quite figured out the exact issue but it appears that when looking for commits with `get_package_diff` the repository is left in a modified state due to some weird interaction with `lfs` files being moved from- or into `git lfs` that previously weren't or shouldn't. Im not entirely sure. 

Anyway, the "fix" that I implemented is to "force" the checkout of a certain commit by adding '-f' to the arguments of `git checkout`. This fixes the issue that occurs in #1186 but I cant fully comprehend if this has some negative consequences for other parts the execution. Feedback is very welcome.

### DIFF
--- a/crates/git_cmd/src/lib.rs
+++ b/crates/git_cmd/src/lib.rs
@@ -213,7 +213,7 @@ impl Repo {
 
     #[instrument(skip(self))]
     pub fn checkout(&self, object: &str) -> anyhow::Result<()> {
-        self.git(&["checkout", object])?;
+        self.git(&["checkout", "-f", object])?;
         Ok(())
     }
 


MarcoIeni:
I wanted to try fix this today, but I didn't get to it :/
If you want to try to fix it, let me know how can I help.

I'm afraid that `git checkout -f` could bring to user data loss somehow, so I prefer to find a more careful approach

baszalmstra:
I would indeed love some help because Im kinda unsure on where to go from here.

Since the problem with `git checkout` only seems to happens in the `checkout_head` function I could also add a `force_checkout` method that is only called there. I assume that since this function is only called after trying to figure out the diffs on a fresh clone of the repository there is very little risk of data loss. Let me know what you think.

MarcoIeni:
Hey sorry if I didn't provide the right guidance on this. 
Anyway I think we need a better way to dealing with lfs which is not force checkout.
Not sure how to do it yet. I haven't had the time to look into it.

By the way if I remember correctly you found a workaround. 👍

So I'm closing this PR and hopefully we'll find a good solution to deal with lfs in the future.

PR #1294, opened 2024-02-25
chore: error if committed Cargo.lock is in `.gitignore`

MarcoIeni:
Close https://github.com/MarcoIeni/release-plz/issues/1286


MarcoIeni:
Use is-ok and check this at the beginning of the program, not on the temp directory

PR #1305, opened 2024-02-26
feat: add git-cliff GitHub integration

orhun:
This PR aims to bump `git-cliff-core` to the latest version to take advantage of the latest features: https://git-cliff.org/blog/2.0.0

There are a couple of todos that I want to resolve and I added comments regarding them. Mainly:

- We need to call `update_github_metadata` function for the GitHub integration. I'm wondering if we already have access to GitHub PRs/commits at the point where this needs to be called.
- There were some types which are almost identical to `git-cliff-core` definitions. I suggest we use them from upstream instead.

Also see: <https://github.com/orhun/git-cliff/issues/468> (I think it would be better if we can get a simple GitHub integration going without cache as the first step)

### DIFF
--- a/crates/release_plz/src/changelog_config.rs
+++ b/crates/release_plz/src/changelog_config.rs
@@ -1,5 +1,5 @@
 use anyhow::Context;
-use git_cliff_core::config::ChangelogConfig;
+use git_cliff_core::config::{ChangelogConfig, RemoteConfig};
 use regex::Regex;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -37,6 +37,7 @@ impl ChangelogCfg {
 }
 
 /// Used for modifying commit messages.
+// TODO: why re-define this instead using it from git-cliff?
 #[derive(Serialize, Deserialize, Default, PartialEq, Eq, Debug, Clone, JsonSchema)]
 pub struct TextProcessor {
     /// Regex for matching a text to replace.
@@ -59,6 +60,7 @@ impl TryFrom<TextProcessor> for git_cliff_core::config::TextProcessor {
     type Error = anyhow::Error;
 }
 
+// TODO: why re-define this instead using it from git-cliff?
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum Sorting {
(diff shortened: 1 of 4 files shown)


crates/release_plz/src/changelog_config.rs
```diff
@@ -37,6 +37,7 @@ impl ChangelogCfg {
 }
 
 /// Used for modifying commit messages.
+// TODO: why re-define this instead using it from git-cliff?
 #[derive(Serialize, Deserialize, Default, PartialEq, Eq, Debug, Clone, JsonSchema)]
```
MarcoIeni, line 41: these structs don't have all the derives release-plz needs. That's the main reason why I duplicated them.
E.g. TextProcessor has `#[derive(Debug, Clone, Serialize, Deserialize)]`. 
git-cliff TextProcessor has `pattern: Regex`, which makes it difficult to implement `Eq` for example.

MarcoIeni:
> I'm wondering if we already have access to GitHub PRs/commits at the point where this needs to be called.

No. We might have them when running `release-plz release-pr`, but we don't have them when we run `release-plz update`, so I suggest to keep the flow consistent among the two commands.

crates/release_plz/src/changelog_config.rs
orhun: Got it, I'm down to implement the necessary traits for these types!

orhun:
I see, can you give me an example of how you retrieve the GitHub commits/PRs? I think then I can just map them to `git-cliff-core`s `GitHubPullRequest`/`GitHubPullCommit` type.

crates/release_plz/src/changelog_config.rs
MarcoIeni: > I'm down to implement the necessary traits for these types!

Definitely not urgent. Plus, there are some pros to have these structs in release-plz. E.g. I can document them in the context of release-plz and have the json schema update accordingly.

MarcoIeni:
> I see, can you give me an example of how you retrieve the GitHub commits/PRs?

[Here](https://github.com/MarcoIeni/release-plz/blob/6c716cca439230dbbb2cc8f67cb57c764947d5ff/crates/release_plz_core/src/command/release_pr.rs). However as I was saying this is only available in `release-plz release-pr`. Git-cliff should also work in `release-plz update`. Why do you want to map this PR to the GitCliff PR type? 🤔

orhun:
I think it would be best to pair-program this when you have time 🐻

MarcoIeni:
Sure Orhun, let's do it!
I'm a bit busy in this period. I will DM you in the first week of April to find some time :)

orhun:
Sounds awesome, let's do that!

MarcoIeni:
In the meantime, we could upgrade to git-cliff-core 2 without the github feature, so that release-plz uses git-cliff latest version at least

orhun:
Sure, I did that in another PR (#1361) to not lose the discussion here.

MarcoIeni:
superseed by https://github.com/MarcoIeni/release-plz/pull/1635
Thanks ❤️

PR #1346, opened 2024-03-15
chore: test freebsd

MarcoIeni:


MarcoIeni:
the test pass, but CD failed.

Issue #1360, opened 2024-03-24
Failed to update workspace packages


orhun:
## Bug description

In my [daktilo project](https://github.com/orhun/daktilo) I switched to workspaces ([split the crate into 2 crates](https://github.com/orhun/daktilo/pull/66)) and now `release-plz` errors with:

```
     Updating crates.io index
note: Downloading daktilo 0.5.0
 Downloading crates ...
  Downloaded daktilo v0.5.0
note: Cloning into "/tmp/.tmpUPqxNn"
  2024-03-24T13:37:32.693781Z  WARN  Package `daktilo_lib@*.*.*` not found
    in download with self=PackageDownloader { packages: ["daktilo", "daktilo_lib"], directory: "/tmp/.tmpUPqxNn", registry: None, cargo_cwd: None }
    in next_versions
    in update
    in release_pr

  2024-03-24T13:37:33.303035Z ERROR  failed to update packages

Caused by:
    0: failed to determine next versions
    1: failed to retrieve difference between packages
    2: cannot compare README files
    3: cannot determine hash of "/tmp/.tmp7VfYce/daktilo/crates/daktilo/../../README.md"
    4: No such file or directory (os error 2)

Error: failed to update packages

Caused by:
    0: failed to determine next versions
    1: failed to retrieve difference between packages
    2: cannot compare README files
    3: cannot determine hash of "/tmp/.tmp7VfYce/daktilo/crates/daktilo/../../README.md"
    4: No such file or directory (os error 2)
```

It might be about:

```toml
[package]
name = "daktilo"
readme.workspace = true
# <strip>
```

But I'm not sure.

- Would you like to work on a fix? - if possible

## To Reproduce

See the CI workflow: https://github.com/orhun/daktilo/blob/main/.github/workflows/cd.yml

Example run: https://github.com/orhun/daktilo/actions/runs/8409724309/job/23027336861

## Expected behavior

No errors.

## Environment

- OS: GitHub Actions (ubuntu)
- release-plz version: `MarcoIeni/release-plz-action@v0.5`

## Additional context

None.

MarcoIeni:
something bad is going on [here](https://github.com/MarcoIeni/release-plz/blob/7a7fb5f9c31f06f32a49794cf9e761df33f8d4c6/crates/release_plz_core/src/package_compare.rs#L148).

Possible solutions:
- return true if `local_package_readme_path` doesn't exist. Not sure if this is the correct thing to do.
[...]

PR #1363, opened 2024-03-25
chore: emit a warning on readme file mismatch

orhun:
See #1360

> Possible solutions:
>
> - return true if `cannot compare README files` and emit a warning. Probably this is the best short-term solution,


MarcoIeni:
thanks! This is a temporary fix, so I'll leave your issue open 👍

PR #1368, opened 2024-03-30
feat: add json output

MarcoIeni:
required for https://github.com/MarcoIeni/release-plz-action/issues/88

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -426,12 +426,12 @@ pub struct Release {
 #[derive(Serialize)]
 pub struct PackageRelease {
     name: String,
-    version: Version,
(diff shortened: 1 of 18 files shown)


crates/release_plz_core/src/command/release.rs
```diff
@@ -417,9 +418,25 @@ pub struct GitRelease {
...
+pub struct Release {
+    packages: Vec<PackageRelease>,
+}
+
+#[derive(Serialize)]
+pub struct PackageRelease {
+    name: String,
+    /// Git tag name. It's not guaranteed that release-plz created the git tag.
+    /// In fact, users can disable git tag creation in the [`ReleaseRequest`].
+    /// We return the git tag name anyway, because users might use this to create
```
MarcoIeni, line 431:
```suggestion
    /// We return the git tag name anyway, because users might use it to create
```

PR #1383, opened 2024-04-04
chore: improve the github graphql query

obmarg:
I was digging around in release-plz and noticed your graphql code was putting everything into the query string.  This isn't necessarily wrong, but it's easier to just use variables.

This commit makes that change - no worries if you don't want it.


MarcoIeni:
tested in https://github.com/MarcoIeni/rust-workspace-example/pull/195/
It works, thanks for the cleanup!

MarcoIeni (approved):

MarcoIeni (approved):

PR #1408, opened 2024-04-20
fix: avoid potential deadlock in large repo

gtema:
`Command.wait` with reading stdin/stdout through pipe can cause a deadlock in a project with big amount of files (see https://stackoverflow.com/questions/64398866/why-does-bufreader-hang-when-reading-from-stderr for details).

Replace `Command.wait_with_output` with `Command.output` to prevent this when calling `cargo` command (similarly to other places in the code).

### DIFF
--- a/crates/release_plz_core/src/cargo.rs
+++ b/crates/release_plz_core/src/cargo.rs
@@ -5,8 +5,7 @@ use tracing::{debug, info};
 
 use std::{
     env,
-    io::{BufRead, BufReader},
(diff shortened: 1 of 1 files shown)


gtema:
The problem was found on the real repo once it grew over certain size. `release-plz` just hangs without any reasonable information (not even with trace log enabled).

MarcoIeni (approved):

MarcoIeni:
thanks! 🙌

PR #1459, opened 2024-05-12
fix: updating pr persists gh verified status

### LINKED ISSUE #1406: release-pr rebase looses verified status
## Bug description

So currently for github PR's, release plz will create the commit through github api to have the `Verified` badge. However when pulling in new changes, it rebases and pushes it using the `git` command which results in updated commit not having the verified status.

## To Reproduce

You can check eg. https://github.com/eigerco/lumina/pull/189. Currently we have to `git commit --amend --S` for each release-plz PR which was rebased by the action.

## Additional context

I've re-checked the [github API](https://docs.github.com/en/graphql/reference/mutations#createcommitonbranch) and there seems to be no way to do it via it. I can't think of any clean solution right now rather than opening an issue for github to add some new API to achieve that.

zvolin:
When updating PR on github, force push the rebased branch first and create commit through API to keep the verified status of the commit.

Fixes #1406

### DIFF
--- a/crates/git_cmd/src/lib.rs
+++ b/crates/git_cmd/src/lib.rs
@@ -133,6 +133,11 @@ impl Repo {
     }
 
     pub fn force_push(&self, obj: &str) -> anyhow::Result<()> {
+        // `--force-with-lease` is safer than `--force` because it will not overwrite
+        // changes on the remote that you do not have locally.
+        // In other words, it will only push if no one else has pushed changes to the remote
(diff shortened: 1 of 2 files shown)


zvolin:
I'll undraft after I test it

zvolin:
Tested it manually (but with auto signing disabled) https://github.com/zvolin/test-release-plz/pull/5

MarcoIeni (approved):

PR #1469, opened 2024-05-18
fix: multiple packages in one changelog with same version

MarcoIeni:


crates/release_plz_core/src/changelog.rs
```diff
@@ -46,10 +47,6 @@ impl Changelog<'_> {
     /// Update an existing changelog.
     pub fn prepend(self, old_changelog: impl Into<String>) -> anyhow::Result<String> {
         let old_changelog: String = old_changelog.into();
-        if is_version_unchanged(&self.release) {
-            // The changelog already contains this version, so we don't update the changelog.
-            return Ok(old_changelog);
-        }
```
MarcoIeni, line 52: I need to investigate. Does removing this have some implications?
Also, I noticed that when `release-plz update` updated Cargo.toml the changelog is correctly added. Why? Compare what happens in the test here with `release-plz update` vs `release-plz update --registry test-registry`.

PR #1502, opened 2024-06-02
feat: add release_link to context

R11baka:
I added release_link to context according to task https://github.com/MarcoIeni/release-plz/issues/1470


MarcoIeni:
That's great, thanks! 
Can you also change this function?
https://github.com/MarcoIeni/release-plz/blob/32c74dd800a77eb5e697b8e88c0eb50d66ed1f3a/crates/release_plz_core/src/changelog.rs#L310
So that release-plz uses `release_link` itself :)

You could also update these docs:
https://github.com/MarcoIeni/release-plz/blob/32c74dd800a77eb5e697b8e88c0eb50d66ed1f3a/website/docs/config.md#L652

MarcoIeni:
I have the fix on my laptop. There's a typo in the changelog body default string. I will push the fix asap

Issue #1506, opened 2024-06-05
 Release request doesn't handle `CARGO_REGISTRIES_<name>_TOKEN` env variable properly


andreyLetenkovWefox:
## Bug description

According to the [doc](https://doc.rust-lang.org/cargo/reference/config.html#credentials), tokens for custom registries may be specified with environment variables of the form CARGO_REGISTRIES_<name>_TOKEN where <name> is the name of the registry in all capital letters. But it appears that at the moment while handling the env variable the registry name is not being put in uppercase, meaning if the custom registry name is `foo` release-plz will try to fetch the corresponding token by `CARGO_REGISTRIES_foo_TOKEN` key, not `CARGO_REGISTRIES_FOO_TOKEN`

- Would you like to work on a fix? [y]

## To Reproduce

Call `release-plz release --registry foo` against custom registry while having `CARGO_REGISTRIES_FOO_TOKEN`  env variable set.

## Environment

- OS: macOS 14.4.1
- release-plz version:  0.3.70

PR #1507, opened 2024-06-05
fix: Handle CARGO_REGISTRIES_<name>_TOKEN env with uppercase

andreyLetenkovWefox:
Fixes [#1506](https://github.com/MarcoIeni/release-plz/issues/1506)

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -176,7 +176,7 @@ impl ReleaseRequest {
             .and_then(|r| {
                 // Credentials for a specific registry can be set using environment variables.
                 // https://doc.rust-lang.org/cargo/reference/config.html#credentials
-                let env_token = env::var(format!("CARGO_REGISTRIES_{r}_TOKEN"))
+                let env_token = env::var(format!("CARGO_REGISTRIES_{}_TOKEN", r.to_uppercase()))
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/command/release.rs
```diff
@@ -754,8 +755,19 @@ fn last_changelog_entry(req: &ReleaseRequest, package: &Package) -> String {
...
+    use lazy_static::lazy_static;
+
+    use fake_package::metadata::fake_metadata;
+
     use super::*;
 
+    lazy_static! {
+        // Trick to avoid the tests to run concurrently
+        static ref NO_PARALLEL: Mutex<()> = Mutex::default();
+    }
```
MarcoIeni, line 769: you don't want tests to run concurrently because you don't want the env var to be set in other tests?
If yes, can you add a comment to explain this?

andreyLetenkovWefox: Yes, exactly. I'll update the comment 👍

MarcoIeni:
Thanks! I don't use release-plz with private registries myself, so I expect some bugs to appear in that code path from time to time 😅

MarcoIeni:
Even though integration tests cover part of it because they use a private registry hosted in gitea with docker 👍

PR #1571, opened 2024-07-22
feat: automatically set git-cliff remote from git remote

JadedBlueEyes:
This automatically sets git cliff's `config.remote.*.owner` / `.repo`, so that it matches the [git-cliff CLI behavior](https://github.com/orhun/git-cliff/blob/e7807e13c4b38aaa4a735ff05b69fdd6b57a7a85/git-cliff/src/lib.rs#L117) and more of git-cliff's example templates work by default.

### DIFF
--- a/crates/release_plz/src/args/update.rs
+++ b/crates/release_plz/src/args/update.rs
@@ -162,18 +162,7 @@ impl Update {
 
             let changelog_req = ChangelogRequest {
                 release_date,
-                changelog_config: Some(
-                    self.changelog_config(
-                        &config,
-                        self.get_repo_url(&config)
-                            .map(|r| Remote {
-                                owner: r.owner.clone(),
-                                repo: r.name.clone(),
-                                token: None,
-                            })
-                            .ok(),
-                    )?,
-                ),
+                changelog_config: Some(self.changelog_config(&config)?),
             };
             update = update.with_changelog_req(changelog_req);
         }
@@ -190,11 +179,16 @@ impl Update {
         Ok(update)
     }
 
-    fn changelog_config(
-        &self,
-        config: &Config,
-        upstream_remote: Option<Remote>,
-    ) -> anyhow::Result<GitCliffConfig> {
+    fn changelog_config(&self, config: &Config) -> anyhow::Result<GitCliffConfig> {
+        let upstream_remote = self
+            .get_repo_url(config)
+            .map(|r| Remote {
+                owner: r.owner.clone(),
+                repo: r.name.clone(),
+                token: None,
+            })
+            .ok();
+
         let default_config_path = dirs::config_dir()
             .context("cannot get config dir")?
             .join("git-cliff")
--- a/crates/release_plz/src/args/update.rs
+++ b/crates/release_plz/src/args/update.rs
@@ -4,8 +4,9 @@ use anyhow::Context;
 use cargo_metadata::camino::Utf8Path;
 use chrono::NaiveDate;
 use clap::builder::{NonEmptyStringValueParser, PathBufValueParser};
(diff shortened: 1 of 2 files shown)


MarcoIeni:
Thanks! Can you update release-plz docs here https://release-plz.ieni.dev/docs/changelog-format#context ?

JadedBlueEyes:
I've added a little bit of documentation! Hopefully, that's sufficient 😅

crates/release_plz/src/args/update.rs
```diff
@@ -205,6 +216,27 @@ impl Update {
                 .try_into()
                 .context("invalid `[changelog] config")?
         };
+        if let Some(remote) = upstream_remote {
+            // This matches https://github.com/orhun/git-cliff/blob/e7807e13c4b38aaa4a735ff05b69fdd6b57a7a85/git-cliff/src/lib.rs#L117
+            // even though it will only set the first unset remote
```
MarcoIeni, line 221: Hey, I now have some time to look into this!

First, with this code, will this ever work with anything other than github?
Have you tested this PR? I don't think we have remote git-cliff features enabled, so even if I merge this, is this going to work?

Also, the release-plz [changelog](https://release-plz.ieni.dev/docs/config#the-changelog-section) configuration doesn't let you set the remote.
This will work only if you use the [changelog_config](https://release-plz.ieni.dev/docs/config#the-changelog_config-field) field which is deprecated.

MarcoIeni: I tried [this](https://github.com/MarcoIeni/release-plz/pull/1575) approach, that autodetects the remote.

Of course git-cliff does more than that, because it also detects the contributors.

JadedBlueEyes: Hi!

> First, with this code, will this ever work with anything other than github?

I essentially copied what git-cliff does in the link in the comment above, so it'll match the behaviour exactly. My reading of the code is that it sets the first unset option, although I haven't tested it in depth. This probably isn't the user-expected behaviour, though!

> Have you tested this PR? I don't think we have remote git-cliff features enabled, so even if I merge this, is this going to work?

I tested it on one of my repositories, yes. I just wanted the `remote.github.owner` and `remote.github.repo` context fields to be set, so I could template links to the repo. And you're right, no optional git-cliff-core features are enabled, so no actual data about the remotes is filled in, [although it would be automatically if those features were enabled](https://github.com/orhun/git-cliff/blob/c208a9791e3899668d48947f59757613f4e5fba1/git-cliff-core/src/changelog.rs#L67).

> the [changelog_config](https://release-plz.ieni.dev/docs/config#the-changelog_config-field) field which is deprecated

I didn't know that it was depreciated! I use it, if just to separate things a bit, so it's good to know

> I tried https://github.com/MarcoIeni/release-plz/pull/1575 approach, that autodetects the remote.

That solves my use case, thank you! The `remote.link` field is especially helpful 😄

MarcoIeni: What are you using the link for? If it's for the release link (i.e. the tag comparison) release-plz adds it already in the context!

JadedBlueEyes: I was using it to generate a link to each commit (like ``[`{{ commit.id | truncate(length=7, end="") }}`]({{ self::remote_url() }}/commit/{{ commit.id }})`` to get [`21c51b9`](https://github.com/JadedBlueEyes/messageformat/commit/21c51b9038d9b74a8cd13b75237f20b1ed11c8c4))

MarcoIeni: Cool! 😁
I'll try to get my pr merged next week 👍

MarcoIeni:
That pr is now released 👍

PR #1602, opened 2024-08-09
release dev-dependencies from workspace before self

a10y:
## Issue

`cargo publish` will check all dependencies, including `dev-dependencies` when verifying a project. You can minimally verify this behavior by adding a non-existent dev-dep to any Cargo project and attempting to run `cargo package`.

Here's a tiny synthetic example to illustrate:

image

This is causing the release publishes to fail when you have a workspace where one crate depends on the other thru dev-dependencies, for example if used in benchmarking code.

We're hitting this issue in the Vortex repo trying to cut releases of our mega-workspace.

* Link to failure: https://github.com/spiraldb/vortex/actions/runs/10319405282/job/28567713447#step:7:787
* link to `Cargo.toml` for the crate that's failing to publish: https://github.com/spiraldb/vortex/blob/develop/vortex-serde/Cargo.toml#L45

## Contribution requirements

* I've verified that `cargo fmt` runs cleanly
* Some tests are failing, however these tests are already failing on `main`:

```
failures:
    changelog::can_generate_single_changelog_for_multiple_packages_in_pr
    changelog::can_generate_single_changelog_for_multiple_packages_locally
    changelog::release_plz_adds_changelog_on_new_project
    changelog::release_plz_adds_custom_changelog
    changelog::release_plz_does_not_open_release_pr_if_there_are_no_release_commits
    changelog::release_plz_releases_a_new_project
    helpers::gitea::gitea_new::can_create_gitea_repository
    release::release_info_contains_prs_in_changelog
    release::release_plz_does_not_release_a_new_project_if_release_always_is_false
    release::release_plz_does_not_releases_twice
    release::release_plz_releases_a_new_project_with_custom_release
    release::release_plz_releases_a_new_project_with_custom_tag_name
    release::release_plz_releases_after_release_pr_merged
```

### DIFF
--- a/crates/release_plz_core/src/release_order.rs
+++ b/crates/release_plz_core/src/release_order.rs
@@ -77,7 +77,7 @@ fn is_dep_in_features(pkg: &Package, dep: &str) -> bool {
 /// Check if the dependency should be released before the current package.
 fn should_dep_be_released_before(dep: &Dependency, pkg: &Package) -> bool {
     // Ignore development dependencies. They don't need to be published before the current package...
-    matches!(dep.kind, DependencyKind::Normal | DependencyKind::Build)
+    matches!(dep.kind, DependencyKind::Normal | DependencyKind::Build | DependencyKind::Development)
       // ...unless they are in features. In fact, `cargo-publish` compiles crates that are in features
       // and dev-dependencies, even if they are not present in normal dependencies.
       || is_dep_in_features(pkg, &dep.name)


a10y:
I took a second look at this repo and noticed that here you use path specifiers for workspace crates used in `dev-dependencies`.

Apologies as I'm a bit new to the multiple-locations stuff for workspaces 😅 . My overall goal is

* Want to use `release-plz` to publish all crates in my workspace in lockstep
* Want to allow for regular `[dependencies]` consumption of crates from other crates in the workspace
* Want to also allow for `[dev-dependencies]` consumption of crates from other crates

So is the recommendation:

* For `[dependencies]`, you use multiple-locations, and then something (?) erases the `path` location before publish
* For `[dev-dependencies]`, since it is only run in the context of a checkout, you can use `path` locations for workspace-local deps

If so, I can close this PR, just wanna make sure I'm doing the right thing ™️ 

Thanks in advance for your RustInfra expertise @MarcoIeni :)

MarcoIeni:
I found that also cargo-release ignores dev dependencies: https://github.com/crate-ci/cargo-release/blob/e30b1c6c645700f4189b76130de0896e19f8d0b1/src/ops/cargo.rs#L478
which means it's probably a good idea to do so 😅 
I remember I also had issues with dev-dependencies, but I don't remember why 😅 I wrote everything I could in those comments.

However let's try to find a solution for your repo 💪

> Some tests are failing

You need docker for some of them. Check CONTRIBUTING.md 👍

Some questions:
1. Do you have a dependency cycle? I.e. vortex-serde and vortex-fastlanes depends on each other? If so, can you remove this cycle?
2. Do you need this dev-dependency to be published in crates.io? If you don't, you can add `publish = false` to Cargo.toml.

If you don't find a solution with points 1 and 2, can you recreate this bug in a much smaller repository? Your repo is giant, so it's hard to reason about it 😅
Once you do this, please close this PR and create a new issue 👍

a10y:
Hey @MarcoIeni thanks for checking this out. I sent a minimal non-workspace example in my original comment. I think the issue is that people should use path specifiers with dev-dependencies, otherwise they should use regular `{ workspace = true }` specifiers, which at build time get replaced with the version specifier.

I did this here: https://github.com/spiraldb/vortex/pull/587 and it's been working. I'll close this PR.

MarcoIeni:
I'm glad you found a solution 😁

Issue #1604, opened 2024-08-09
bug: {{ commit.author.name }} not working


gacallea:
## Bug description

`{{ commit.author.name }}` not working in `{% macro print_commit(commit) -%}` 

- Would you like to work on a fix? [n]
- I wish I could but I am still studying The Book. Pretty n00b still. If you think I could try even at this level, I'd be happy to help!

## To Reproduce

1. Given [this template](https://github.com/gacallea/freesound-credits/blob/main/release-plz.toml) - (btw I want to contribute a PR for this template once we fixed this very issue)
2. Given [this cliff.toml](https://gist.github.com/gacallea/d42949306eb7855e8d661baace32a45d) - (this is just for manual testing/comparison, not used by release-plz in the end)
3. Given an existing commit history.
5. run `git cliff -o CHANGELOG.md` to fill the file a first time
6. make some commits just to test the next step
7. run `release-plz update` and see that under Unreleased the author is missing (line ending with `by` on its own)

## Expected behavior

each commit should end with an author. e.g: `Adding pre-commit hook for convetional - (924c906) by gacallea`

## Screenshots

0.2.12 by [release-plz GitHub Action](https://github.com/gacallea/freesound-credits/blob/main/.github/workflows/release-plz.yml)
0.2.11 and below generated by `git cliff -o CHANGELOG.md` with the process described in the steps above

<img width="705" alt="Screenshot 2024-08-09 at 10 22 47 PM" src="https://github.com/user-attachments/assets/842b1ec2-fe44-4afc-8b23-bbf19eedee2e">

## Environment

- OS: macOS Sonoma 14.6
- release-plz version:  0.3.83

## Additional context

Please ignore the quirks of the `cliff.toml` -- namely the broken rewrites and broken links -- as they are irrelevant here.

MarcoIeni:
This feature is missing. It's not a bug.
https://github.com/MarcoIeni/release-plz/issues/989

gacallea:
Sorry about the bug report then! Thank you for release-plz. <3

MarcoIeni:
No worries 😁

gacallea:
one question @MarcoIeni: do you like the template at all? should I PR? (ofc minus the broken stuff, which I am removing as we speak). ref: https://github.com/gacallea/freesound-credits/blob/main/CHANGELOG.md

MarcoIeni:
Sure, the template looks dope! 🙌
[...]

Issue #1009, opened 2023-09-30
Add ability to re-generate the entire changelog


MarcoIeni:
## Motivations

Release-plz only appends the new changes to the changelog.
It would be nice if it could generate the whole changelog from scratch. This can be useful because:
- if you change the changelog configuration, you can apply the changes also to previous releases
- if you haven't used release-plz before, you can generate the changelog for previous versions

## Solution

This should be a new release-plz subcommand. `release-plz update` should only append the changes for the new version.

TODO

## Additional context

Suggested [here](https://github.com/MarcoIeni/release-plz/discussions/1002#discussioncomment-7151904)

intgr:
You can also call `git cliff` directly. By default it will generate the whole changelog.

MarcoIeni:
yes, we just need to make sure to use the same release-plz config:
https://github.com/MarcoIeni/release-plz/blob/c5eed2fb325420582b8ad7ede4ccbae74fa69fd9/crates/release_plz_core/src/changelog.rs#L225

I need to document this better

Issue #989, opened 2023-09-24
List contributors in GitHub release


MarcoIeni:
## Motivations

It's nice to give contributors some recognition for their work.
GitHub tags the author of the pr when you autogenerate a release from their UI. Release-plz should tag the authors as well.

## Solution

For every PR number in the changelog, call [this](https://github.com/MarcoIeni/release-plz/blob/378060645e25228d956e549d6d5d975ee80b3eb0/crates/release_plz_core/src/git/backend.rs#L347) function to retrieve the pr author (and contributor?).

Than, in the [release_body](https://github.com/MarcoIeni/release-plz/blob/378060645e25228d956e549d6d5d975ee80b3eb0/crates/release_plz_core/src/command/release.rs#L419) add a list of contributors:

```md
## Contributors

- `@author1`
- `@author2`
```

Only if we are filling a github release. I don't know if gitea supports this.

MarcoIeni:
Check https://github.com/github/contributors

MarcoIeni:
We can use git-cliff github integration. To do this, set the ReleaseConfig in here: https://github.com/MarcoIeni/release-plz/blob/ed8a348c235dd5bff60af07849a36177a734eda6/crates/release_plz_core/src/changelog.rs#L120 (default_git_cliff_config fn).
Then, git-cliff will fetch github data, and will add it to the tera context.

MarcoIeni:
you can now add contributors to the changelog but not in the github/gitea/gitlab release yet

MarcoIeni:
closed by https://github.com/MarcoIeni/release-plz/pull/1736

PR #1607, opened 2024-08-10
docs(changelog): add example as approved in #1604

gacallea:
matching https://github.com/gacallea/freesound-credits/blob/ac12bc0cf2df64df935f8f1fbd6e2a3888f22ca6/CHANGELOG.md

I would add some wording and screenshot but I am keeping simnple for now, matching the docs style.

cheers!


MarcoIeni:
The config page is already huge so I don't want to add changelog configuration examples there. Imo we should create a new page similar to [this](https://git-cliff.org/docs/templating/examples/) one. Do you want to work on it?

gacallea:
I agree with you. I didn’t know if I could.
I am on it!
  
On Thu, Aug 15, 2024 at 8:15 AM, Marco Ieni ***@***.***&gt; wrote:  The config page is already huge so I don't want to add changelog configuration examples there. Imo we should create a new page similar to this one. Do you want to work on it?

gacallea:
@MarcoIeni ignore pushes for now. I will let you know when it's ready.

gacallea:
@MarcoIeni I believe I am done phew....

In the end I preferred splitting the single file into its own category and adding sections.

Examples wise: this was the hard part to fix/adjust/test rinse and repeat. I decided to include the three most interesting ones instead of copying over from git-cliff. Some of them may require things you don't have in your schema and if people really want what I considered redundant and uninteresting, they can be my guest... it was a PITA :P

I also added some language conf to docusaurus because highlight of toml and json wasn't working in code blocks.

Let me know what you think!

gacallea:
ps: I have not being able to avoid the messed menu in Examples (see on the right, it includes mdx-code-block when it shouldn't ) but I believe an up to date docusaurus will fix that.

gacallea:
hi @MarcoIeni , I see some things fail. Link I cabn't do much about it since the examples folder doesn't exist yet. Anything else let me know if you need anything from me.

website/docusaurus.config.js
```diff
@@ -157,6 +157,7 @@ const config = {
       prism: {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
+       additionalLanguages: ['toml', 'json'],
```
MarcoIeni, line 160:
```suggestion
        additionalLanguages: ['toml', 'json'],
```

MarcoIeni:
> I believe an up to date docusaurus will fix that.

I updated the branch so that we are using a more recent version 👍
Running `npm i` brings you to 3.5.1

I'm having a quick look now, but I don't have time to do a full review. Maybe later today 👍

website/docusaurus.config.js
gacallea: sorry about that!

MarcoIeni:
Thanks for adding json and toml syntax highlight btw. It looks awesome 🤩

gacallea:
trying to fix the broken toc on examples. requesting backup https://github.com/facebook/docusaurus/discussions/10409

gacallea:
> trying to fix the broken toc on examples. requesting backup [facebook/docusaurus#10409](https://github.com/facebook/docusaurus/discussions/10409)

not much I can do without breaking at least something. I am leaving this as is.

gacallea:
@MarcoIeni I would like to add two/three more examples:

1 - default (showing release-plz default config/output)
2 - [github](https://github.com/orhun/git-cliff/blob/main/examples/github.toml) and/or [github+keepachangelog](https://github.com/orhun/git-cliff/blob/main/examples/github-keepachangelog.toml)

for 1: where can I find this to easily add an example file?
for 2: this depends if release-plz uses the same *.github.* variables but I haven't been able to find them in the schema.

cheers

MarcoIeni:
> for 2: this depends if release-plz uses the same .github. variables but I haven't been able to find them in the schema.

Yes, I haven't implemented those yet. https://github.com/MarcoIeni/release-plz/issues/989
I plan to work on this soonish 👍

gacallea:
> > for 2: this depends if release-plz uses the same .github. variables but I haven't been able to find them in the schema.
> 
> Yes, I haven't implemented those yet. #989 I plan to work on this soonish 👍

ok then :) I think I am done for real now :)

let me know if you need anything else!

MarcoIeni:
I noticed the default changelog isn't producing the correct format. I removed the extra new line in the body but I still need to understand why the header doesn't have the new line. Maybe it's a bug 👀
image

Also, I think it's better to inline the examples in the docs instead of having them in a separate examples dir.

gacallea:
> I noticed the default changelog isn't producing the correct format. I removed the extra new line in the body but I still need to understand why the header doesn't have the new line. Maybe it's a bug 👀 image
> 
> Also, I think it's better to inline the examples in the docs instead of having them in a separate examples dir. In this way, people do less back-and-forth

Fair enough :)

I added the extra line because of that "bug" and because markdown wants lines in between headers and such. I do so in my config as well.

Regarding the examples dir vs inline, I also tend to agree if it is clean enough. I just did what git-cliff was doing.

I am going to check these changes later, if there is anything I need to do please let me know.

gacallea:
ps: I have an idea for inlining the conf:

just adding a toml code block in the same fashion as the output ones (folded):

- Tomo configuration
- Raw output
- Rendered output

gacallea:
> ps: I have an idea for inlining the conf:
> 
> just adding a toml code block in the same fashion as the output ones (folded):
> 
> * TOML configuration
> * Raw output
> * Rendered output

I implemented this, let me know if it's good now :)

gacallea:
@MarcoIeni regarding markdown-cli failing: I tried fixing those header coming from `mdx-code-block` code but there is no way around it. (see above discussion on docusaurus repo). I'd opt to remove those entirely. What do you reckon?

website/docs/configuration/changelog_examples.md
```diff
@@ -0,0 +1,643 @@
...
+{% endif -%}
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }}\
+    {% endfor %}
+{% endfor %}\n
+"""
+
+footer = """
```
MarcoIeni, line 207: release-plz doesn't support the footer. That's because git-cliff only recently added the ability to have templates there so at the beginning I didn't add support for it.
I created https://github.com/MarcoIeni/release-plz/issues/1621

Also, I'm not a fan of commit_parsers. With those commit parsers a commit like `fix: add abc` will be identified as "Changed".

Since (except for the footer part), this Keep A Changelog config is very similar to the default one, imo we can delete this example entirely. What do you think?

gacallea: footer: sorry I didn't know and didn't check. my bad. I don't even like having the footer.

keep a changelog can go then :)

commit_parsers: I am new to this and I liked the bling bling, it's your call tho :)

let me know if I should change it or if you arew taking care of it.

gacallea: done, let me know if you prefer to keep mdx-blocks . they were mostly an annoyance.

gacallea: ps: feel free to change the commit_parsers too, but I think it's good to give alternatives examples otherwise why would even have this section?

gacallea: alternatively: redefining the parsers grouping to your liking?

MarcoIeni: For the mdx-blocks we can disable the lint 👍
I'll think about it

gacallea: let me know, I have currently removed them.

website/docs/configuration/tips-and-tricks.md
```diff
@@ -0,0 +1,89 @@
...
+- ⚡ Performance
+- ⚙️ Miscellaneous
+
+Then strip the tags in the template with the series of filters:
+
+```jinja2
+### {{ group | striptags | trim | upper_first }}
+```
+
+### Use GitHub PR labels as groups
```
MarcoIeni, line 57: we don't have this feature yet.

website/docs/configuration/changelog_examples.md
```diff
@@ -0,0 +1,440 @@
...
+"""
+
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [unreleased]
+{% endif %}\
+{% if previous %}\
+    {% if previous.commit_id and commit_id %}
```
MarcoIeni, line 337: https://github.com/gacallea/release-plz-contrib/blob/f05ecfe613564ab7dbdd412d71c47b236d84a4db/crates/release_plz_core/src/changelog.rs#L290
This is never populated

website/docs/configuration/changelog_examples.md
```diff
@@ -0,0 +1,440 @@
...
+header = """# Changelog
+
+All notable changes to this project will be documented in this file.
+
+"""
+
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
```
MarcoIeni, line 333: We never go into the `else` because release-plz always populates the `version` field. I'm removing the detailed template because it doesn't fit for release-plz. If needed, we can add it into another PR 👍

website/docs/configuration/changelog_examples.md
gacallea: sorry. (see next comment)

website/docs/configuration/changelog_examples.md
gacallea: sorry about that!

truth be told I have been struggliong a lot with testing out stuff because I had to rely on `git cliff -o CHANGELOG.md` during my contribution. I am yet to find a way (even for mty personal use) to generate a changelog from scratch entirely with release-plz. Is there a command to do that?

MarcoIeni: >  to generate a changelog from scratch entirely with release-plz. Is there a command to do that?

no 😂 https://github.com/MarcoIeni/release-plz/issues/1009

> sorry about that!



Don't worry :)

MarcoIeni:
I reverted back the config.md file. I'm not sure it's the right thing to do because changelog is a part of the configuration but at least in this way we don't break existing links (maybe there's a way to redirect, not sure). Anyway for now it should be good enough 👍

gacallea:
> I reverted back the config.md file. I'm not sure it's the right thing to do because changelog is a part of the configuration but at least in this way we don't break existing links (maybe there's a way to redirect, not sure). Anyway for now it should be good enough 👍

it is your project/site and the choice is ultimately yours. i did my best categorizing things and fixed all links `npm run build` complained about. unless you are worried about SEO here.

MarcoIeni (approved):
Yay, we did it! Thanks 🎉

gacallea:
yay! thank you Marco for this amazing tool! I did my best but you had to comb over lots! Sorry it is my first contrib with Rust (even if just docs). I'll do better next time :P

(not right now but I would like to have a go at #1009 when I finish The Book -- ssssh don't tell anyone!)

MarcoIeni:
It was a great contribution. I'm happy you did your first Rust contribution in release-plz 🥳

That issue is really hard! If you want a less hardcore experience check [good-first-issues](https://github.com/MarcoIeni/release-plz/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) 👍

PR #1630, opened 2024-08-19
feat: add remote author username to changelog

MarcoIeni:
Close https://github.com/MarcoIeni/release-plz/issues/989

- [ ] add docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ duration-str = { version = "0.11.2", default-features = false }
 expect-test = "1.5.0"
 fake = "2.9.2"
 fs-err = "2.11.0"
-git-cliff-core = { version = "2.4.0", default-features = false }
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ duration-str = { version = "0.11.2", default-features = false }
 expect-test = "1.5.0"
 fake = "2.9.2"
 fs-err = "2.11.0"
-git-cliff-core = { version = "2.4.0", default-features = false }
+git-cliff-core = { version = "2.4.0", default-features = false, features = ["github", "gitea"] }
```
MarcoIeni, line 28: it feels bad that we want to just add some values and our cargo.lock exploded. Maybe we can edit git-cliff to add arbitrary fields to the commits, so that we don't need to enable these features.

MarcoIeni: opened https://github.com/orhun/git-cliff/issues/803

PR #1651, opened 2024-09-03
feat: add gitlab support for the `release-pr` command

mbodmer:
Hi Marco

I have tried to finish support for Gitlab. Can you please check whether there is a path I was missing (create/update/close MR is what I have found) and guide me accordingly if necessary.

Hope the code is in your sense, just tell me.

Marc


MarcoIeni:
Thanks ❤️

Did you test it?
E.g. you ran `release-plz release-pr` on gitlab and it worked?

mbodmer:
Yes, certainly. on our internal instance. It's my goal to use it for our internal projects.

MarcoIeni:
That's great! I will review this asap. I've been waiting years for someone to implement gitlab support 😁
Can you update the docs to document how to set it up? E.g. creating tokens, ci jobs (if you use one), etc.

mbodmer:
I have added documentation and streamlined it with the existing one from release-plz release. It is essentially the same. For gitlab-ci I cannot be of any help, sorry. We don't use it.

crates/release_plz_core/src/git/backend.rs
```diff
@@ -422,7 +560,7 @@ impl GitClient {
         })?;
 
         let prs = match self.backend {
-            BackendType::Github => {
+            BackendType::Github | BackendType::Gitlab => {
```
MarcoIeni, line 563: are you sure this works?
The response returned by [this](https://docs.gitlab.com/ee/api/commits.html#list-merge-requests-associated-with-a-commit) endpoint doesn't seem compatible with GitPR

MarcoIeni: btw I reviewed the rest of the code and done some few changes. It looks great!
I just need to test this once we resolve this comment and then we should be done 👍

MarcoIeni:
Out of curiosity, do you think we could run gitlab in docker the same way we run gitea for our integration tests?
Not blocking for this pr anyway

mbodmer:
I was impressed at how good the Gitea tests work. I am sure it could also be done for Gitlab and would be beneficial for stability. I have no experience maintaining or setting up a Gitlab instance however.

My colleagues have a hard time maintaining data integrity and upgrading, but just a standard setup for testing doesn't look that scary to me: https://docs.gitlab.com/ee/install/docker/installation.html#install-gitlab-using-docker-compose

crates/release_plz_core/src/git/backend.rs
mbodmer: I was also surprised this worked right away and did not investigate it further. But you are right, this doesn't look sound.

mbodmer: Well spotted, I don't know why this worked. But the decision whether or not to build a release worked correctly and still does after the fix now :-)

MarcoIeni:
> The hard part would probably be to add some default testing repository to it. How did you do that for Gitea?

with api calls https://github.com/MarcoIeni/release-plz/blob/4cae59899b40838ebccb1d5084456fb2d576e976/crates/release_plz/tests/all/helpers/gitea/gitea_new.rs#L60

MarcoIeni:
I'm not opening an issue for gitlab support for now, I think the maintainance burden might be higher than the actual benefits, in the sense that we already have gitea to cover the same code paths expect gitlab api calls which should be stable.

If we notice a lot of regression in gitlab, we can add gitlab integration tests later.
I think our efforts are better spent on other issues for now.

That being said, if that's an issue you are very passionate about adding these integration tests, I won't stop you and I'll review PRs :)

MarcoIeni:
image
This PR worked at first try, thank you so much!

mbodmer:
Thanks, last year I have attended your talk in Florence, this year I probably won't make it. Maybe next year. Thanks also for your commitment here.

MarcoIeni:
> Thanks, last year I have attended your talk in Florence, this year I probably won't make it. Maybe next year.

wow, that's great! See you next time than 😁

> Thanks also for your commitment here.

It was a pleasure :)

PR #1678, opened 2024-09-14
fix: reset git config (if issue with global config)

davidB:
small change to avoid an error during test with git, when the git's local configuration is `commit.gpgsign = true` but the  `user` defined in the test doesn't have an ssh or gpg private key.


MarcoIeni (approved):

MarcoIeni:
the title and the description is much more complicated than your change.
I just wanted to double check that you committed everything.

davidB:
Yes it was completed at least to fix the issue I got with my config.
But I found maybe a more general solution (ignoring user global configuration). So I'll move this PR back to draft to test it.

davidB:
The way to reset (more config) is to set some environment variables. I'll close the PR, until I got a better idea (to just reset config for test, and not for every commands)

PR #1703, opened 2024-09-20
fix: changelogs only include commits that edit packaged files

### LINKED ISSUE #1701: Commits to child crate is added to changelog of parent crate in workspace
## Bug description

When I have multiple crates in the same workspace and make a commit with changes only to a child crate then the commit is also added to the parent crate in the changelog and the version of the parent crate is bumped.

## Example
https://github.com/Profiidev/release-plz-test/pull/1

Profiidev:
# Changes

This change will compare the source files of each crate in a multi crate workspace with the changed files in a commit and only add the commit to the changelog if they overlap. Source files include all files in a crates directory that are not exclude either by `.gitignore` ore the package.exclude field in the `Cargo.toml`

## Open Questions
- [x] What should happen if a commit has no changed files?
- [x] What should happen if a crate has no source files?
- [x] Should this be a setting? Default enabled or disabled?

fixes #1701


crates/release_plz_core/src/next_ver.rs
```diff
@@ -599,6 +599,27 @@ impl Updater<'_> {
         registry_packages: &PackagesCollection,
         repository: &Repo,
     ) -> anyhow::Result<Vec<(&Package, Diff)>> {
+        // Get the source files for each package and save them to a hashmap
+        // This ignores all files in .gitignore ore in the package.exclude field in the Cargo.toml
+        let path = repository.directory();
+        let context = cargo::util::context::GlobalContext::default()?;
+        let ws = cargo::core::Workspace::new(&path.as_std_path().join("Cargo.toml"), &context)?;
```
MarcoIeni, line 606: isn't this the same of https://github.com/MarcoIeni/release-plz/blob/986c949b4ecbd5be2fd902fc08a5ae0e4f484df1/crates/release_plz_core/src/package_compare.rs#L105

Profiidev: Oh yes. But i get an additional result `Cargo.toml.orig`

Profiidev: Should i use your option and just hardcode to skip `Cargo.toml.orig`?

MarcoIeni: Yes please. I hope to remove cargo as a dependency one day https://github.com/MarcoIeni/release-plz/issues/1331

Profiidev: already done

Profiidev:
What do you think about the open questions? I believe that the first two are impossible but should this be a setting?

thenightness:
Is there any Update in this?

MarcoIeni:
I didn't find the time to review this yet. I'll try to review it this week.

I have a full time job so I can only dedicate my spare time to release-plz, since my github sponsor barely cover my spotify subscription 😅

The open questions don't need extra work I think

Profiidev:
NP. Take your time

crates/git_cmd/src/lib.rs
```diff
@@ -105,6 +105,22 @@ impl Repo {
...
+    pub fn changes_current_commit(
+        &self,
+        filter: impl FnMut(&&str) -> bool,
+    ) -> anyhow::Result<Vec<String>> {
+        let output = self.git(&[
+            "show",
+            "--oneline",
+            "--name-only",
+            "--pretty=format:''",
+            &self.current_commit_hash()?,
```
MarcoIeni, line 117: what's the point of adding the current commit? Doesn't it default to it?

crates/git_cmd/src/lib.rs
```diff
@@ -105,6 +105,22 @@ impl Repo {
...
+            "-r",
```
MarcoIeni, line 118: what's `-r`?
I tried this command with and without it and the result is the same

MarcoIeni: I didn't find it in https://git-scm.com/docs/git-show

Profiidev: its recursive for submodules

Profiidev: https://stackoverflow.com/questions/424071/how-do-i-list-all-the-files-in-a-commit

MarcoIeni: `-r` is related to `diff-tree` instead you used the `show` subcommand. Anyway I will switch to `diff-tree` because they recommend it 👍

MarcoIeni:
Also, let me know if you want to write a test for this (in another per so we merge this soon).

You can fine the tests directory in the release-plz crate

Profiidev:
I can write a Test but in a different PR

crates/git_cmd/src/lib.rs
Profiidev: I think we can drop it because submodules would maybe have their ow release workflow

MarcoIeni:
I used two hashsets instead of a vec to improve performance. I remember you also used hashsets at the beginning, right? 👍

Profiidev:
Nice. I think i didn't use them.

Profiidev:
I used a HashMap for crate lookup

crates/release_plz_core/src/next_ver.rs
```diff
@@ -900,11 +901,25 @@ impl Updater<'_> {
         tag_commit: Option<&str>,
         diff: &mut Diff,
     ) -> anyhow::Result<()> {
+        let package_files = get_package_files(package_path, repository)?;
```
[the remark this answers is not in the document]
MarcoIeni: I moved this inside the loop

MarcoIeni:
I found an optimization. We could refactor the code to run `cargo package` once.

We run it again here:

https://github.com/Profiidev/release-plz/blob/f64e86e80a4a3054cb48ec725efff5f61218a1b4/crates/release_plz_core/src/package_compare.rs#L41

Do you want to do it in another PR?

MarcoIeni:
Interesting, I broke some tests 😅

MarcoIeni:
I'm done for today. Hopefully tests will pass! Let me know what you think about my changes and ramblings.

My fear is that after this change some commits will be omitted by mistake. That's why I'm putting a lot of effort into this PR.

crates/release_plz_core/src/next_ver.rs
```diff
@@ -905,6 +906,12 @@ impl Updater<'_> {
         loop {
             let current_commit_message = repository.current_commit_message()?;
             let current_commit_hash = repository.current_commit_hash()?;
+
+            // Check if files changed in git commit belong to the current package.
+            // This is required because a package can contain another package in a subdirectory.
+            let are_changed_files_in_pkg =
```
Profiidev, line 912: I think we can just compute the result here

MarcoIeni: By using a closure the function is executed only if needed. I.e. If we reach the else if cases.

Instead, if we compute the result there, the function is always executed.

By using a closure we save compute. What do you think?

MarcoIeni:
- [ ] what happens if a package includes a file from a parent directory? If that file is edited, is that commit added to the changelog?

crates/release_plz_core/src/next_ver.rs
Profiidev: Thats true but in most cases the closure will be executed. But besides that i tested it and everything is working properly

Profiidev:
> * [ ]  what happens if a package includes a file from a parent directory? If that file is edited, is that commit added to the changelog?

I don't think that is possible

Profiidev:
I test it and you can't include files from a parent directory

MarcoIeni:
Is it forbidden by cargo? I see 👍

Profiidev:
When i include a file in by sub_crate in the test repo with `../../.gitignore` for example, it does not show in `cargo package --list`

Profiidev:
I also think that it would be impossible to represent in the registry because the file would need to be stored in a higher folder than the crate it self which could lead to polution of your filesystem

MarcoIeni:
ok, I'm merging this, thank you for your PR and quick responses :)
If you would enjoy it, feel free to open the related PRs we discussed here (tests and passing `cargo package --list` across functions so that it's only invoked once) 👍

PR #1716, opened 2024-09-27
fix: Disable changelog in PR body if it goes over github character limit

### LINKED ISSUE #1715: PR body is too long
I'm evaluating starting to use release-plz over at [`atsamd-rs`](https://github.com/atsamd-rs/atsamd). One particularity of that project is that we're maintaining over 50 crates in the same workspace. We sometimes make sweeping changes that triggers a release for many crates at once.

So far I've found that release-plz would (impressively!) take care of all our needs when it comes to releasing, with one caveat: When there are too many changes at once, github returns a 422 error when trying to open a PR. It turns out the PR body goes over the API's character limit (65536) . See here for an example of a failed run: https://github.com/jbeaurivage/atsamd-release-test/actions/runs/11062726303/job/30738197929

Would there be a way of customizing the PR body to somehow make it shorter? I think this could be accomplished by optionally not including the crates' changelogs in the body, and only leaving the change summary in there.

jbeaurivage:
As described in #1715, some workspaces may have enough changes to publish that trying to open a PR fails because the body is too long. This PR adds a configuration option to disable adding the changelog to the PR body, leaving only the changes summary.

### Alternatives

One alternative would be to let the user specify whether they want to include the changelog in the PR body for each package. Another would be to forgo adding a user option and automatically fail over to the summary-only body if we detect that the body is too long.

Closes #1715.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,7 +127,7 @@ zero_sized_map_values = "warn"
 
 # Allowing optional lints has no effect, but we list them to
 # make clear that we reviewed them and we decided to allow them
-decimal_literal_representation = "warn"
+decimal_literal_representation = "allow"
 map_unwrap_or = "allow"
 needless_raw_string_hashes = "allow"
 redundant_else = "allow"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,6 @@ cast_lossless = "warn"
 cast_possible_wrap = "warn"
 cast_precision_loss = "warn"
 cast_sign_loss = "warn"
-decimal_literal_representation = "warn"
 default_trait_access = "warn"
 doc_markdown = "warn"
 exit = "warn"
@@ -128,6 +127,7 @@ zero_sized_map_values = "warn"
 
 # Allowing optional lints has no effect, but we list them to
 # make clear that we reviewed them and we decided to allow them
+decimal_literal_representation = "warn"
 map_unwrap_or = "allow"
 needless_raw_string_hashes = "allow"
 redundant_else = "allow"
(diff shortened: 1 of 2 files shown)


jbeaurivage:
Modified this PR to:
* Automatically only include the header, summary and footer in the PR body if it's longer than 65536 bytes when including the full changelog.
* If it's still too long, simply truncate the body to make it short enough.

@MarcoIeni, this should be ready to go!

MarcoIeni:
I think the limit is on characters, not bytes, right?

If yes, can you avoid referring to bytes in the code and refer to characters instead?

jbeaurivage:
Honestly, I'm really not sure whether the limit is on characters or on. The GH API docs are unclear on this (in fact, the character limit isn't even mentioned).

I just went with the more conservative route and limit the body to 65535 bytes, because a 65535 byte-long string will always have fewer than 65535 characters, but the same isn't true the other way around.

If you'd prefer to impose a limit on characters instead just let me know!

MarcoIeni:
Can you test this? 🤔
I.e. Try to open a pr with both those limits and see how it behaves

MarcoIeni:
Also the `len` function counts the number of characters, not bytes, right?

jbeaurivage:
Okay so I opened a few test PRs. The GH API indeed imposes a limit on characters and not bytes. It will also accept exactly 65536, but not 65537 chars. I amended this PR to reflect that. The only (minor) downside is that counting characters is O(n) whereas counting bytes ([that's what `String::len` does!](https://doc.rust-lang.org/std/string/struct.String.html#method.len)) was O(1).

MarcoIeni (approved):

MarcoIeni:
Thanks! I'll try to release this in a few hours

PR #1768, opened 2024-10-12
feat: add `remote.pr_number` to `remote` context of commits in changelog

### LINKED ISSUE #1767: Add `remote.pr_number` to `remote` context of commits in changelog
## Motivations

Needed for ratatui!

## Solution

Fill `pr_number` of `RemoteContributor` struct.
Do it here: https://github.com/MarcoIeni/release-plz/blob/51416d4e9e69cfcdd3a12df069bace6d6f839b8f/crates/release_plz_core/src/changelog_filler.rs#L46

KekmaTime:
In the `changelog_filler.rs` file, I added logic to retrieve the associated PRs for a commit and extract the PR number, which then gets stored in the `RemoteContributor` struct.

In the `backend.rs` file, I introduced a new field, `pr_number: Option<u64>`, to the `RemoteCommit` struct to store the PR number. I also updated the code to make sure this new field gets populated when creating a `RemoteCommit`. These changes help the system link pull request numbers with commits, making it easier to generate detailed changelogs and track which PRs are included in each release.

closes #1767


.gitignore
```diff
@@ -1,6 +1,6 @@
-/target
+target/
```
MarcoIeni, line 1: why did you change this? This is the default if you do `cargo new` 🤔

crates/release_plz_core/src/changelog_filler.rs
```diff
@@ -25,6 +25,7 @@ pub async fn fill_commit<'a>(
     if let Some(existing_commit) = all_commits.get(&commit.id) {
         commit.author = existing_commit.author.clone();
         commit.committer = existing_commit.committer.clone();
+        commit.remote = existing_commit.remote.clone();
```
MarcoIeni, line 28: thanks for the fix!

.gitignore
KekmaTime: I wanted to clarify that the target folders inside the test directories were also being included during commits. That's why I made this change to ensure that recursive folders don't accidentally get committed. Hope that is okay!

crates/release_plz_core/src/changelog_filler.rs
KekmaTime: anything for the community ❤️

.gitignore
MarcoIeni: Ok, sure!

PR #1790, opened 2024-10-19
changing some typo

Wizdomcss:


MarcoIeni:
Why do you think these are typos?

PR #1815, opened 2024-11-05
feat: separate pull requests per package in the workspace

maartenflippo:
Adds a configuration option `one_pr_per_package` (boolean, default false). If true, release plz will open a separate PR for each package in the workspace.

The option interacts as expected with the existing templating capabilities for `{{ package }}` from the config.

Closes https://github.com/MarcoIeni/release-plz/issues/1814.

### DIFF
--- a/.schema/latest.json
+++ b/.schema/latest.json
@@ -50,6 +50,7 @@
         "git_release_type": null,
         "git_tag_enable": null,
         "git_tag_name": null,
+        "one_pr_per_package": null,
         "pr_body": null,
         "pr_branch_prefix": null,
         "pr_draft": false,
@@ -603,6 +604,14 @@
           "description": "Tera template of the git tag name created by release-plz.",
           "type": [
             "string",
+            "null"
+          ]
+        },
+        "one_pr_per_package": {
+          "title": "One PR per Package",
+          "description": "- If `true`, create a separate PR for every package. - If `false` or [`Option::None`], a single PR releasing all packages at once.",
+          "type": [
+            "boolean",
             "null"
           ]
         },
--- a/crates/release_plz/src/main.rs
+++ b/crates/release_plz/src/main.rs
@@ -44,13 +44,18 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
             let cargo_metadata = cmd_args.update.cargo_metadata()?;
             let config = cmd_args.update.config()?;
             let update_request = cmd_args.update.update_request(&config, cargo_metadata)?;
-            let request = get_release_pr_req(&config, update_request)?;
-            let release_pr = release_plz_core::release_pr(&request).await?;
+            let requests = get_release_pr_reqs(&config, update_request)?;
+
+            let mut prs = vec![];
+            for request in requests.iter() {
+                let release_pr = release_plz_core::release_pr(request).await?;
+
+                if let Some(pr) = release_pr {
+                    prs.push(pr);
+                }
+            }
+
             if let Some(output_type) = cmd_args.output {
-                let prs = match release_pr {
-                    Some(pr) => vec![pr],
-                    None => vec![],
-                };
                 let prs_json = serde_json::json!({
                     "prs": prs
                 });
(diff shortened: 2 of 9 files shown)


maartenflippo:
@MarcoIeni There is currently one test that is failing: https://github.com/MarcoIeni/release-plz/blob/183dd271690cb0925901cdf36e7a05bdba0f6da6/crates/release_plz/tests/all/release.rs#L17

In the test case I have merged 1 of 2 release prs, and expect a single release to be created. Instead, two releases are created. I was wondering whether this is because the initial condition has both packages unreleased. This would mean that the code could detect that package "two" with a version in the manifest `0.1.0` can be released as there is no existing package with that version.

In any event, I am not sure how to fix this issue. If you have any input that would be appreciated.

.schema/latest.json
```diff
@@ -50,6 +50,7 @@
         "git_release_type": null,
         "git_tag_enable": null,
         "git_tag_name": null,
+        "one_pr_per_package": null,
```
MarcoIeni, line 53: For the name, what do you think about `pr_per_package`?

crates/release_plz/src/main.rs
```diff
@@ -44,13 +44,18 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
...
+
+            let mut prs = vec![];
+            for request in &requests {
+                let release_pr = release_plz_core::release_pr(request).await?;
+
+                if let Some(pr) = release_pr {
+                    prs.push(pr);
+                }
+            }
+
```
MarcoIeni, line 57: `prs` is only used at line 60. Can you calculate `prs` iat line 59 as it was before?
Maybe you can have a vector of `release_prs` outside of the `if let Some` 👍

crates/release_plz/src/main.rs
```diff
@@ -82,7 +87,39 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
...
+                let update_request = update_request
+                    .clone()
+                    .with_single_package(package.name.clone());
+
+                let release_pr_request = get_single_release_pr_req(config, update_request)?;
+
+                let branch_prefix = release_pr_request.branch_prefix().to_owned();
+
+                let release_pr_request = release_pr_request
+                    .with_branch_prefix(Some(format!("{branch_prefix}{}-", package.name)));
```
MarcoIeni, line 111:
```suggestion
                    .with_branch_prefix(Some(format!("{branch_prefix}-{}", package.name)));
```
this is a typo, right?

crates/release_plz/tests/all/helpers/test_context.rs
```diff
@@ -81,10 +81,14 @@ impl TestContext {
         context
     }
 
-    pub async fn merge_release_pr(&self) {
+    pub async fn merge_release_prs(&self) {
         let opened_prs = self.opened_release_prs().await;
         assert_eq!(opened_prs.len(), 1);
-        self.gitea.merge_pr_retrying(opened_prs[0].number).await;
+        self.merge_release_pr(&opened_prs[0]).await;
```
MarcoIeni, line 87: this function is called "merge_release_prs" but you only merge the first one 🤔

crates/release_plz_core/src/command/release_pr.rs
```diff
@@ -62,6 +62,10 @@ impl ReleasePrRequest {
         self
     }
 
+    pub fn branch_prefix(&self) -> &str {
+        self.branch_prefix.as_str()
```
MarcoIeni, line 66:
```suggestion
        &self.branch_prefix
```
nit: more concise :)

website/docs/config.md
```diff
@@ -317,6 +318,11 @@ Where:
 - `{{ package }}` is the name of the package.
 - `{{ version }}` is the new version of the package.
 
+#### The `one_pr_per_package` field
+
+- If `true`, release-plz creates a separate branch and PR for every package version.
+- If `false`, release-plz creates s single branch and PR combining all package releases. *(Default)*.
```
MarcoIeni, line 324:
```suggestion
- If `true`, release-plz creates a release PR for every package version.
- If `false`, release-plz creates a single release PR combining all package releases. *(Default)*.
```
Imo creating a branch is implicit when creating a PR, so no need to say it 👍

MarcoIeni:
> I was wondering whether this is because the initial condition has both packages unreleased.

Exactly. To fix it, first run `release`, then update the packages with another commit and then run `release_pr`. If the test still fails, let me know and I'll give it a try. In worst case we don't add a test.

You can already run your fork on release-plz in your repo and check if it works :)

.schema/latest.json
maartenflippo: That does sound better

crates/release_plz/src/main.rs
maartenflippo: Not really. From the documentation for the `branch_prefix` configuration option, I infer that the prefix contains whatever separator character is desired at the end, which is why there is no dash between the prefix and the package name. The '-' is a suffix for the package name to separate it from the timestamp.

I am open to suggestions on how that can be made more clear

crates/release_plz/tests/all/helpers/test_context.rs
maartenflippo: Good catch. This is intentional, and I wanted to discuss it but forgot to point it out. 

This is the original `merge_release_pr` function. I was hesitant to change its behavior since I was unsure whether it would impact the existing tests that relied on this specific implementation.

The tests do pass if we actually close all PRs, so it does not seem to be a problem. It might change some subtle behavior I am unaware of though.

marcoieni:
closing for inactivity

PR #1833, opened 2024-11-13
In Quickstart, release-plz-pr should "need" release-plz-release

danking:
It seems to me weird that these two actions race with one another. If release-plz-release somehow releases before release-plz-pr runs, a new release PR is created (because the local version equals the crates.io version). If release-plz-pr runs first, which is the most likely outcome, no PR is created.

Forcing an ordering between them removes this non-determinism and doesn't seem to do any harm.

### DIFF
--- a/website/docs/github/quickstart.md
+++ b/website/docs/github/quickstart.md
@@ -162,6 +162,7 @@ jobs:
   # and you want to update `Cargo.toml` versions and changelogs by yourself,
   # remove this job.
   release-plz-pr:
+    needs: [ release-plz-release ]
     name: Release-plz PR
     runs-on: ubuntu-latest
     # The concurrency block is explained below (after the code block).


MarcoIeni:
> If release-plz-release somehow releases before release-plz-pr runs, a new release PR is created (because the local version equals the crates.io version)

If the local version equals the crates.io version, the release pr is created only if the two packages differs.

Here's an example from release-plz logs:

image


> doesn't seem to do any harm

the downside of having ordering is that release-pr runs after a delay, so the release PR is created more slowly 


My theory is that the two jobs can run concurrently without any race condition. But happy to discuss if you think there could be side effects. I certainly want to avoid them 👍

MarcoIeni:
If you can reproduce a race condition (maybe by adding sleeps) or if you think I'm missing something let me know!

MarcoIeni:
I think a race condition can happen when the first version of the package is released.
- if `release` finishes before `release-pr`, the pr is not opened
- otherwise, the release-pr is opened.

But if it's only for the first `cargo publish` I think we shouldn't slow down all other use cases.

PR #1843, opened 2024-11-19
feat(init): check if Cargo.toml constains the `version` field in local dependencies

### LINKED ISSUE #1763: release-plz init: check if the local dependencies of public crates have a `version` field
## Motivations

Example error:

```
 2024-10-10T10:30:11.050960Z ERROR  failed to release package

Caused by:
    failed to publish expandable-impl:     Updating crates.io index
    error: all dependencies must have a version specified when publishing.
    dependency `rust-grammar-dpdfa` does not specify a version
    Note: The published dependency will use the version from crates.io,
    the `path` specification will be removed from the dependency declaration.
```

If a dependency has a `path`, it means it's a local dependency.
All local dependencies need to be like this:

```toml
marco-test-one = { path = "crates/marco-test-one", version = "0.3.21" }
```

I.e. if `path` is present, `version` must be present to.

## Solution

Add this check to `release-plz init` and report an error if there's a dependency that doesn't have the `version` field.

You can look the metadata of the packages to check if `version` is present.

Angr1st:
Implements #1763: adds additional check for `release-plz init` that ensures all local dependencies have a version specified. Specifying `version = "*"` is treated like having no version specified, which follows the behavior of cargo.

Closes #1763.

### DIFF
--- a/crates/release_plz_core/src/project.rs
+++ b/crates/release_plz_core/src/project.rs
@@ -163,6 +163,7 @@ impl Project {
     // Check mandatory fields for crates.io
     pub fn check_mandatory_fields(&self) -> anyhow::Result<()> {
         let mut missing_fields = Vec::new();
+        let mut missing_version_errors = Vec::new();
 
         for package in &self.publishable_packages() {
             if package.license.is_none() {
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/project.rs
```diff
@@ -171,21 +172,53 @@ impl Project {
...
+{}",
+                missing_versions.join("\n")
+            ));
         }
+        error_message.push_str("\nNote: to disable this check, set the `--no-toml-check` flag.");
+        anyhow::bail!(error_message);
+    }
+}
+
+fn check_local_dependencies(package: &Package, missing_versions: &mut Vec<String>) {
```
MarcoIeni, line 204: can you return `missing_versions` from this function instead of taking it as input?
I think it's more clear.

I use this pattern mainly for recursive functions or similar

crates/release_plz_core/src/project.rs
```diff
@@ -171,21 +172,53 @@ impl Project {
...
+            }
+            missing_version.push_str(&format!(
+                "\t• local dependency `{}` is missing a `version` entry\n",
+                dependency.name
+            ));
+            has_missing_version = true;
+        }
+    }
+    if has_missing_version {
+        missing_versions.push(missing_version);
```
MarcoIeni, line 221: I would expect this function to return a vector of dependencies, i.e. `dependency.name` for the package. Than the actual error string is calculated outside of this function (e.g. in another function).
If you do it like this, you don't need the `missing_version` variable because in the other function you can add the string `- package `{}`:\n` if the vector is empty.

crates/release_plz_core/src/project.rs
Angr1st: Done.

crates/release_plz_core/src/project.rs
Angr1st: Done.

crates/release_plz_core/src/project.rs
```diff
@@ -163,6 +163,7 @@ impl Project {
     // Check mandatory fields for crates.io
     pub fn check_mandatory_fields(&self) -> anyhow::Result<()> {
         let mut missing_fields = Vec::new();
+        let mut missing_versions = Vec::new();
```
MarcoIeni, line 166:
```suggestion
        let mut missing_versions_errors = Vec::new();
```

MarcoIeni:
Nice thanks! I just left one last minor comment

Angr1st:
Good catch, I renamed the variable. Is there anything else you need me to do?

crates/release_plz_core/src/project.rs
Angr1st: Done. I lost an 's' and used `missing_version_errors` instead of `missing_versions_errors`

MarcoIeni:
Thanks!

PR #1855, opened 2024-11-22
fix: find Cargo.lock file in monorepo

cfsnyder:
A monorepo may have multiple separate cargo projects within it. The prior logic assumed that the Cargo.lock file would be present at the root of the repo. This update crawls up the directory tree to find the first Cargo.lock file relative to the package directory.

### DIFF
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -938,9 +938,8 @@ impl Updater<'_> {
 
             // Check if files changed in git commit belong to the current package.
             // This is required because a package can contain another package in a subdirectory.
-            let are_changed_files_in_pkg = || {
-                self.are_changed_files_in_package(package_path, repository, &current_commit_hash)
-            };
+            let are_changed_files_in_pkg =
+                || are_changed_files_in_package(package_path, repository, &current_commit_hash);
 
             if let Some(registry_package) = registry_package {
                 debug!(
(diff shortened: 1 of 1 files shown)


MarcoIeni:
> The prior logic assumed that the Cargo.lock file would be present at the root of the repo. 

Can Cargo.lock be in a subdirectory? If yes, why would you do it? 🤔
Isn't the Cargo.lock shared with all the packages?

If you can link some resources I would like learn more about this before merging this 🙏

cfsnyder:
I did submit this to resolve an error that I encountered while running release-plz. In my monorepo configuration, I have multiple release-plz workflow configurations, one for each project that's in the repo. So in the Google example, I'd have separate release-plz configurations for `calendar` and `drive`, each triggered by separate actions triggers based upon the paths of changed files, and each with a different configured working directory. The `--manifest-path` option didn't seem to change where the code was looking for the `Cargo.lock` file.

cfsnyder:
Essentially I got an error saying that the Cargo.lock file didn't exist because it was always looking in the root of the git repo, regardless of where the corresponding Cargo.toml was.

MarcoIeni:
ok, makes sense!

MarcoIeni:
I don't like the "recursive" approach.
We have the manifest path from the flag you mentioned. Why we aren't finding the Cargo.lock? 🤔

Are you able to debug this? I'll try to look into this when I find some time.

MarcoIeni:
I was able to reproduce the bug you described and I fixed it in https://github.com/release-plz/release-plz/pull/1894
Thanks a lot 🙌

I'm going to release the new version of release-plz now. Let me know if you still notice this issue with the new version. 👍

Also, thanks a lot for trying to fix this issue. Your solution definitely worked, but as you saw there was a simpler fix.

cfsnyder:
thanks @MarcoIeni !

PR #1861, opened 2024-11-24
♻️ refactor: use tera template for default PR body

### LINKED ISSUE #1788: pr_body context: use a tera template for the default configuration
https://github.com/MarcoIeni/release-plz/pull/1778 added the pr_body context.
However it maintained the old logic to construct the default pr_body:

https://github.com/MarcoIeni/release-plz/blob/main/crates/release_plz_core/src/pr.rs#L134

This is totally fine because I like small PRs 👍

However, it would be cool to use a tera template to render the template of the PR body.
In this way, we could easily document what's the default pr_body, so that people can create their own template more easily

KekmaTime:
closes #1788 

1. Moved the default PR body template to a constant
2. Using Tera templating  for both default and custom templates
3. Simplifying the `pr_body_default` function

### DIFF
--- a/crates/release_plz_core/src/pr.rs
+++ b/crates/release_plz_core/src/pr.rs
@@ -8,8 +8,8 @@ pub const DEFAULT_BRANCH_PREFIX: &str = "release-plz-";
 pub const OLD_BRANCH_PREFIX: &str = "release-plz/";
 pub const DEFAULT_PR_BODY_TEMPLATE: &str = r#"
 {% macro get_changes(releases, type="text") %}
-    {%- for release in releases %}
-    {%- if release.title and release.changelog %}{% if releases | length > 1 %}
+{%- for release in releases %}
+{%- if release.title and release.changelog %}{% if releases | length > 1 %}
 ## `{{ release.package }}`
 {% endif %}
 <blockquote>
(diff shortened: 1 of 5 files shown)


MarcoIeni:
Hey! I had the feeling that this won't work, so I added a test in the main branch and updated your PR. Can you make the test pass, please? 🙏

crates/release_plz_core/src/pr.rs
```diff
@@ -135,29 +144,29 @@ fn pr_body_default(
...
-        format!(
-            "<details><summary><i><b>Changelog</b></i></summary><p>\n\n{changes}\n</p></details>\n"
-        )
-    };
+    let mut context = tera::Context::new();
+    context.insert("summary", &packages_to_update.summary());
+    context.insert(
+        "changes",
+        &packages_to_update.changes(project_contains_multiple_pub_packages),
+    );
```
MarcoIeni, line 152: I saw you added other variables to the context (summary and changes).
Instead, it would be nice to re-use the same variables we already documented in https://release-plz.ieni.dev/docs/config#the-pr_body-field

KekmaTime: will do it asap. i was busy with my finals hope that is fine.

marcoieni (approved):
Thanks for doing the initial work on this.
This is issue was more complicated than expected 😅

KekmaTime:
> Thanks for doing the initial work on this. This is issue was more complicated than expected 😅

ohh i see. i was struggling there couldnt debug what exactly the issue was 😅

PR #1872, opened 2024-11-28
feat: Add git-only mode

### LINKED ISSUE #1144: Add a "git only" mode

eviltak:
Resolves #1144. Based off of @ozgunozerk's work in #1497.


MarcoIeni:
💪 let me know when you want me to take a look 👍

PigeonF:
Hey, I know this in draft mode, but maybe my feedback is still useful to you.

I tried out your changes, but ran into issues because the repo I use mixes different tagging schemes

- rust crates are tagged as `{{ package }}-v{{ version }}`
- there are some other tags for non-rust related things as well (both of the form `{{ package }}-v{{ version }}`, and `{{ version }}`).

With your current implementation, some of these unrelated tags get picked up instead, which might not contain any of the rust packages when checked out. If I understood your approach correctly, you try to find the latest git tag (in general), check out the repository at that tag, and then use the checked out repository as the manifest path, correct? I'm pretty sure this can easily desync, since you check out _any_ latest version, instead of the latest version of a specific package.

I tried out an alternative approach (inspired by yours) [in my fork](https://github.com/PigeonF/release-plz/tree/PigeonF/push-rzwyrkwrtrvr): for each package marked as `git_only` `release-plz` searches for tags that match that package's tagging scheme (instead of any matching tag), and then adds a git worktree for that specific version. In my testing this worked quite well (though the code is proof of concept only, and needs cleaning up). In case this is useful to you @eviltak , feel free to copy (and/or ignore) as much of my version as you want.

P.S.: I'm not sure what the etiquette is for jumping onto a draft PR like this is, so if I overstepped I apologize in advance 😅

eviltak:
Thanks for the references @PigeonF! The contents of this PR are currently simply what were included in #1497 - the issues you have outlined are known (https://github.com/release-plz/release-plz/pull/1497#issuecomment-2185000159), and I have code locally that addresses the issue. I wasn't using git worktrees to check out the different package manifests at the time of release as I am not familiar with them, but they may end up working than the approach I had planned.

eviltak:
@MarcoIeni While writing tests for this feature I noticed that the `release` command does not work without a specified git client, even if we disable the creation of git releases. Besides the git release creation process, I noticed that the git client is only being used to retrieve PR contributor information, and to check if a release PR was merged. It looks to be pretty straightforward to skip these two things in the absence of a configured git client - do you think it would make sense to allow this behavior? It shouldn't impact normal operation - it would just let you release packages (i.e. create git tags and run `cargo publish`) in the absence of remote git repositories (which, for example, is what my test relies on).

MarcoIeni:
You can have a look at the integration tests of release_plz crate to see how to have a remote repository.

We use gitea for that.

Sorry for the brief response. Let me know if it's helpful or you still need more help

eviltak:
I managed to figure out a way by using a local repo as the upstream, and a `GitHubMockServer` to mock the `GitBackend`. I have a few more tests I want to add, but otherwise this PR should be ready for you to start taking a look at!

While you review, I'd love to hear your thoughts on what to do in the situation where old releases had a different tag name template (the second task in the PR description). The current implementation would simply ignore the old release tags generated using the old template(s). My hunch is such a change would lead to `update` not being able to find the latest release after such a change, which would in turn lead to it not incrementing the version (potentially causing a subsequent `release` to re-release the same version), but I haven't actually tested what would happen. I shall write a test to see what happens in this scenario and check for expected behavior (regardless of whether we choose to handle this case or not).

If we were to handle this situation correctly, it would require checking out each version of the release-plz config file at each potential release tag and using the tag name template from the checked out config file. This can be a bit tricky because the config file parsing code is entirely in the CLI `release-plz` crate, and not the core release-plz library crate. There are other ways to do this, but they all would require the core library being able to parse config files.

While investigating this, I also noticed that other code (for example [here](https://github.com/release-plz/release-plz/blob/main/crates/release_plz_core/src/next_ver.rs#L799-L801) and [here](https://github.com/release-plz/release-plz/blob/main/crates/release_plz_core/src/next_ver.rs#L899-L901)) seems to implicitly rely on the assumption that the git tag template  never changes: they also always use the current tag template to reference older tags, and do not consider the possibility of an old release having a different tag template. If this is the case, then it might be worth making that assumption (along with any potential side effects of violating it) explicit in the documentation.

MarcoIeni:
Thanks, I'll have a look asap!

I think for this pr we can assume the tag template never changes for simplicity, yes.

Fyi, Another reason why the tag template might change is when the project moves from single package to workspace.

We can merge this and open a new issue when we try to address the tag name problem. Not sure what the solution might be there.

PigeonF:
I just tested this PR in a private repo, and it works 🎉 (though I do not publish any packages to a registry, so I did not test the interaction between `git_only` and non-`git_only` packages).

The only thing that is counter-intuitive when running `release-plz release` is that `git_only` is a bit of a misnomer:

In order for the package to be considered, the `Cargo.toml` has to set `package.publish = true` (or to some registry). However, setting `workspace.package.git_only` in `release-plz.toml` still tries to publish to `crates.io`. This means that in order to only publish git tags, and not to any registry, one has to

1. Set `package.publish=true` in `Cargo.toml`
2. Set `workspace.package.git_only=true` in `release-plz.toml`
3. Set `workspace.package.publish=false` in `release-plz.toml`

I think this should be noted in docs somewhere, because (as a non-native english speaker) the name `git_only` made me think that release-plz would not try to publish to a registry. Alternatively, `git_only` could override `publish` in this case (but this might also be surprising), or release-plz could pick up packages that have `package.publish=false` set in `Cargo.toml`, but `workspace.package.git_only` set in `release-plz.toml`.

MarcoIeni:
I think we still want to have `package.publish = false`. That's a safety measure to avoid publishing the package to a registry.

Probably if `git_only = true` we should also consider packages with `publish = false`.

Release-plz shouldn't attempt to publish to a registry if `git_only = true` imo.

> release-plz could pick up packages that have package.publish=false set in Cargo.toml, but workspace.package.git_only set in release-plz.toml.

exactly 👍

eviltak:
You bring up a great point @PigeonF. I shall modify the code so that both `release` and `update` consider `git_only` packages as publishable, and add tests to verify that crates with `publish = false` in their `Cargo.toml` work in `git_only` mode. 

To that end, I do believe that we should use a more descriptive config option name than `git_only`. Something along the lines of `treat_git_tags_as_published_versions`, while more verbose, makes the implication of setting that option a lot clearer. Open to any suggestions!

MarcoIeni:
I agree that `git_only` is a bad name.

`publish_git_tag_only`? Open to suggestions as well 😂

eviltak:
I think `publish_git_tag_only` is a bit of a misnomer, since it implies that the option limits release publishing to git tags only, which is not the case. The option makes `release-plz` use git tags to determine published package versions versus retrieving them from the registry, so IMO it doesn't make sense for it to imply that it affects publishing (since the current implementation does not affect publishing).

I am rewording my previous suggestion to `git_tags_provide_published_versions`. If no one has any objections or better suggestions by the time I finish the other pending tasks in this PR, I shall proceed with renaming the option to it.

elpiel:
I will chime in as I'm interested in the PR, if naming is one of the last remaining tasks 😅.

I like your suggestion, I'd change only 1 thing:
`git_tags_as_published_versions`
But again, I'd rather see the feature implemented than argue about the naming so if @MarcoIeni likes your suggestions I'm also up for it.

elpiel:
I will chime in as I'm interested in the PR, if naming is one of the last remaining tasks 😅.

I like your suggestion, I'd change only 1 thing:
`git_tags_as_published_versions`
But again, I'd rather see the feature implemented than argue about the naming so if @MarcoIeni likes your suggestions I'm also up for it.

Edit: another option is to just use what cargo is using. 
`publish= false`
or
`cargo_publish = false`

MarcoIeni:
> it implies that the option limits release publishing to git tags only, which is not the case.

So with this option set, you still publish to crates.io?

Maybe it can be confusing.

What's the use case where somebody wants to use git tags to determine the next published version, but still publish to crates.io?

The only one I can think of is if someone has a private registry and doesn't want release-plz to access it.

Another design choice we could do is forcing people to set `publish = false` in the Cargo.toml, so that people don't publish by mistake to crates.io.
Another option is force people to set the `publish` field (not sure if possible to verify if the publish field is set to true from cargo metadata).

I'm interested in hearing the opinion of everyone involved in this PR 👍

Btw, I might have some time to review this pr this weekend 👍

MarcoIeni:
> Edit: another option is to just use what cargo is using. 
> `publish= false`
> or
> `cargo_publish = false`

We already have something like this in our config. 
It disables cargo publish. But assumes you run it by yourself

eviltak:
> So with this option set, you still publish to crates.io?

Yes, if publishing is enabled for that crate/workspace.

> What's the use case where somebody wants to use git tags to determine the next published version, but still publish to crates.io?
> 
> The only one I can think of is if someone has a private registry and doesn't want release-plz to access it.

One other scenario where this would be useful is if someone wants to publish only some released versions of a package to a registry. Selectively publishing versions with `release-plz` without modifying the config file may not even be possible today, but it is possible that a command-line argument is added to the `release` command in the future to skip publishing versions for selected packages (overriding the config file).

I agree that in almost all cases (like my intended use case for this feature), if you want to use git tags to determine the next published version you will _only_ be publishing to git tags and _not_ the registry. However, since the feature does not rely on this implication at all to function correctly (and works irrespective of whether you publish to a registry or not), I am in favor of keeping the separate option (with a better name) without tying it into the `publish` option.

eviltak:
Another thing I just noticed: for obvious reasons, `git_tag_enable = false` is incompatible with `git_only = true`. What is the recommended way to show such an incompatibility to the user? 

I think this is severe enough to be considered an error; I can check this condition and return an error when the `release` command [checks the git tag enabled condition](https://github.com/release-plz/release-plz/blob/main/crates/release_plz_core/src/command/release.rs#L752), but is there a better place to do this, such as in the config schema or in a config validation routine?

elpiel:
> Another thing I just noticed: for obvious reasons, `git_tag_enable = false` is incompatible with `git_only = true`. What is the recommended way to show such an incompatibility to the user? 
> 
> I think this is severe enough to be considered an error; I can check this condition and return an error when the `release` command [checks the git tag enabled condition](https://github.com/release-plz/release-plz/blob/main/crates/release_plz_core/src/command/release.rs#L752), but is there a better place to do this, such as in the config schema or in a config validation routine?

As far as I know, in such a scenario it would be better to use an enum instead of 2 Boolean fields. Maybe that will solve the whole question of naming the config?

I'm not sure how release-plz currently handles such compatibility where configuration value is changed from boolean to a different type, however, I though I would share this idea.

MarcoIeni:
## Answers
> One other scenario where this would be useful is if someone wants to publish only _some_ released versions of a package to a registry. 

true, but very rare, I think I've never seen an example of this. I can see release-plz not supporting this use case since it's very rare. For this case, people could just run `cargo publish` when needed (also because workspace support for `cargo publish` is progressing).

> Selectively publishing versions with `release-plz` without modifying the config file may not even be possible today, but it is possible that a command-line argument is added to the `release` command in the future to skip publishing versions for selected packages (overriding the config file).

It's not possible, yes, but as I said above I don't see a strong case for it 🤔

## Proposal
> I am in favor of keeping the separate option

Ok, what about "forcing" users to take the decision? I.e. forcing people to set the [`publish`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) field of the Cargo.toml.

Pro:
- If they want to keep this package away from crates.io, they set `publish = false`, which is a best practice
- If they want to publish the package when using `git_only` it's a weird decision, so we want them to explicitly say so by setting `publish = true`.

Of course the con is that people need to update their Cargo.toml, but at least they do this to follow a best practice (setting `publish = false`).


## Implementation detail of proposal
Unfortunately, `cargo_metada` doesn't distinguish between `publish = true` and the `publish` field being absent ([publish](https://docs.rs/cargo_metadata/0.19.1/cargo_metadata/struct.Package.html#structfield.publish) is None in both cases), so we would need to read the Cargo.toml file.

## `git_tag_enable`

The enum [elpiel](https://github.com/elpiel) proposed is an option we can think about in the future. For now, let's keep things as they are to keep the size of this pr reasonable 👍 

We can validate that the two fields are compatible [here](https://github.com/release-plz/release-plz/blob/76175590af6e45a8bf60043a6ce1d1a413ca0b9c/crates/release_plz/src/args/config_command.rs#L11), so that we fail very early.

[eviltak](https://github.com/eviltak), if you prefer to have the enum in this pr, we can have it of course 👍

> I'm not sure how release-plz currently handles such compatibility where configuration value is changed from boolean to a different type, however, I though I would share this idea.

We would need to release release-plz version 0.4 because it's a breaking change, but it's not an issue, since I haven't shipped a breaking change in years 😂

MarcoIeni (commented):
some initial comments. Didn't manage to review everything yet

crates/release_plz_core/tests/all/git_only.rs
```diff
@@ -0,0 +1,175 @@
...
+use release_plz_core::PackagesUpdate;
+use std::path::PathBuf;
+
+fn touch(path: impl Into<PathBuf>) -> std::io::Result<()> {
+    fs_err::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .open(path)
+        .map(|_| ())
+}
```
MarcoIeni, line 12: cool function! can you use it in your changes to `crates/release_plz/tests/all/release_pr.rs` as well?

crates/release_plz_core/tests/all/helpers/git_only_test.rs
MarcoIeni, line 1: probably we can extract some duplicate code from this file 👍

MarcoIeni (commented):
I managed to review everything. Thanks, awesome work 🙌

I have left minor comments but in general it looks great!

eviltak:
In my opinion it doesn't make sense to combine `git_only` and `git_tag_enable` into one enum option. Though the former option is dependent on the latter, the latter is completely independent of the former. I also am not sure what we would call the amalgam of these two options. 

However, if we decide to enforce not publishing to a registry when using this feature (which I am completely fine with), we can take another approach (inspired by @elpiel's suggestion) that will solve our naming woes:

1. Eliminate the `git_only` option
2. Make `publish` an enum, with three variants (open to suggestions on naming): `Registry`, `GitTagsOnly`, `Disabled`
3. Enable this feature for all packages that have `publish == Publish::GitTagsOnly`
4. Override the value of `git_tag_enable` and always push tags  if `publish` is set to `GitTagsOnly` (this could also be a config validation error, as discussed eariler)

If we choose to go with this approach, it might be worth taking it a step further, which in my opinion will both improve naming and prevent a breaking change to the release-plz config schema:

1. Create a new config field, `publish_to`, which describes the target (registry, git tags or nothing) to publish packages to and is set to the enum as described above
2. Mark the `publish` field as deprecated and map `publish = false` to `publish_to = Disabled` (might rename `Disabled` to `None` or similar) and `publish = true` to `publish_to = Registry` if no value for `publish_to` is specified

MarcoIeni:
What a great idea, I like it but I would like to give it some thoughts!

> (this could also be a config validation error, as discussed eariler)

I would prefer this 👍

A good reason to deprecate `publish` is because it's easy to confuse release-plz `publish` and Cargo.toml `publish`.

Let me think about `publish_to` a bit more. I have doubts because `Registry` also implies "git tags", while `GitTagsOnly` doesn't imply "registry". What if tomorrow we introduce a new publish strategy, e.g. a POST to an endpoint (I just made this up)? Maybe individual booleans make more sense 🤔🤔🤔🤔

eviltak:
I got to thinking about this problem a bit more, and I believe I  have found a good approach inspired by my previous suggestion that should simplify things a lot. I am going to post below a (perhaps long-winded) description of my understanding of the problem, which will hopefully convey the reasoning behind my suggestion.

## Background

Conceptually, release-plz supports a set $T$ of $N$ publish "targets". Based on the configuration, we publish each package $p$ to a subset $T_p \subseteq T$ of these targets (if $T_p$ is empty, we do not publish the package in any form). 

In order to determine the next version of a package $p$, we must look up the latest version of that package published to _any_ target in the set $T_p$ (i.e. the latest version of this package retrievable from any source) . In most cases, the published version would be the same in all targets, but in some cases (e.g. if a new target was added to the set $T_p$ recently) there may be a discrepancy. We thus pick the latest version published to _any_ target to avoid re-publishing already published versions to a target.

## What release-plz does today

Today, the set of supported targets $T$ consists of 3 targets: publishing to a registry, git tags, and git repository releases, and the set $T_p$ of targets we publish package $p$ to is determined by the config options `publish` (for the registry), `git_tag_enable` (for git tags) and `git_release_enable` (for git releases). In the future, $T$ could have an additional target, like sending a custom POST request to a server.

However, release-plz currently only supports looking up the latest version of a package published to the "registry" target; which means that we cannot retrieve the latest published version of packages that aren't published to a registry (but may be published to git tags and/or git releases and/or any other form of publishing that is added in the future). 

## This PR

This PR aims to address part of that problem - support for retrieving the latest versions of packages that aren't published to a registry, but are published to git tags. However, it does so by introducing a _binary_ (exclusive OR) alternative to fetching latest published versions from a registry (enabled by the `git_only` option). This means that a package has to _choose_ whether to retrieve its latest version from the registry or git tags; as we discuss below, this need not be the case.

## What release-plz (and this PR) should be doing

As discussed in the Background section above, for a package $p$, rather than fetching latest published versions from a specific target (e.g. registry vs. git tags), release-plz should retrieve the latest version from all targets in the set $T_p$, and use the latest of all those versions.

In the context of this PR, this would mean getting rid of the `git_only` option altogether. Furthermore, when fetching the latest version of a package $p$, we would do the following:

- If the package has `publish = true` (i.e. is published to a registry):
  - Fetch the latest version `registry_version` (if any) from the registry
- If the package has `git_tag_enable = true` (i.e. is published to git tags):
  - Fetch the latest version `tag_version` (if any) from git tags
- Return the latest (greater) version between `registry_version` and `tag_version`

This allows release-plz to automatically support packages that are not published to a registry (but are published to git tags) without having the user manually set an option. 

In the case where a package has an older version published to the registry and a newer one published to a git tag (or vice versa), we pick the latest tag version because if we chose the older registry version, we would end up overwriting the git tag (or attempting to overwrite a version already published to the registry) after the version increment. 

## In the future
Note that even with this PR, retrieving published versions from the third currently supported publish target (git releases) is unsupported. In other words, release-plz will not be able to determine the latest published version of a package that is only published to git releases. This might be worth adding in the future

It might also be worthwhile considering a refactor to add a `PublishTarget` abstraction with an enum and/or a trait that looks like the following:

```rust
trait Target {
    async fn publish(&self, package: &Package) -> Result<()>;
    async fn fetch_latest_package(&self, package: &Package) -> Result<Option<Package>>;
}

struct Registry {
    // All publish_* config options
}

struct GitTag {
    // All git_tag_* config options
}

struct GitRelease {
    // All git_release_* config options
}

// Could use #[enum_dispatch]
enum PublishTarget {
    Registry(Registry),
    GitTag(GitTag),
    GitRelease(GitRelease),
}
```

This allows `publish_to` to be an array of the targets you wish to publish a package to (in code, a `Vec<PublishTarget>`).

```rust
fn release(package: &Package) {
    // ...
    for target in &package.config.publish_to {
        target.publish(package).await?;
    }
}

fn next_version(package: &Package) {
    let mut latest_package: Option<Package> = None;
    for target in &package.config.publish_to {
        let target_latest_package = target.fetch_latest_package(package).await?;
        if latest_package.map(|p| p.version) < target_latest_package.map(|p| p.version) {
            latest_package = target_latest_package;
        }
    }
}
```

MarcoIeni:
I like this approach a lot because we don't need to edit our configuration to introduce this new feature. Thanks for taking the time to think about it!

I just have a few thoughts.

> In the case where a package has an older version published to the registry and a newer one published to a git tag (or vice versa), we pick the latest tag version because if we chose the older registry version, we would end up overwriting the git tag (or attempting to overwrite a version already published to the registry) after the version increment.

To make this change less disruptive for existing release-plz users, if `publish = true` I will keep the existing behavior (where the registry takes precedence over the git tags in picking the next version), and maybe log a warning that says that there's a git tag with a newer version. We could always tweak this later if needed 👍

If `publish = false` we should pick the latest between git tags and registry. One might think that if `publish = false` we can completely ignore the registry version. However, release-plz users who have `publish = false`, run cargo publish manually or in other ways today. So we should expect that they publish the version to the registry.

After this PR, there will be people who use `publish = false` to achieve the git only functionality.

## In the future

For git releases, I'm not sure if it's worth adding, because I would expect the same info to be available in git tags. Let's see if somebody requests this and why.

> This allows publish_to to be an array

Let's keep the booleans for now I would say. We can discuss switching to `publish_to` in another issue 👍

Regarding the refactoring, feel free to do whatever you want 👍
Just be aware of the size of the PR. Nothing stops you from sending other PRs after we merge this one to do additional improvements.

Again, thanks a lot for pushing this forward, I really like that you bias towards providing solutions rather than asking questions 🙌

crates/release_plz_core/src/published_packages.rs
```diff
@@ -0,0 +1,274 @@
...
+            // We do this by simply creating a new worktree pointing to the release tag.
+            // We could also do this in other ways:
+            // 1. Find relative path to package manifest and checkout package contents
+            //    (see git read-tree and checkout-index) into temp dir
+            // 2. Use `cargo package` to create tarball, and extract it
+            // But the simplest is to use a worktree
+
+            repo.add_worktree(package_store_dir.as_std_path(), release_tag)?;
+
+            let metadata = cargo_utils::get_manifest_metadata(
```
antonbaliasnikov, line 166: Could you please check that the `get_manifest_metadata` `Cargo.toml` path is correct in case of cargo workspace and crates in some nested directory.

Locally, I see that the worktree contains the whole workspace checked out in the `<tmpdir>/<package_name>` and the function is unable to correctly find the package `Cargo.toml` that would be in  `<tmpdir>/<package_name>/<path_to_the_package_in_workspace>/<package_name>/Cargo.toml` .

antonbaliasnikov: This is a proper path: https://github.com/release-plz/release-plz/pull/1872/files#r1882718672

eviltak: This is intentional - this line loads the manifest of the workspace in the repository root, and [later](https://github.com/release-plz/release-plz/pull/1872/files#diff-4890bc1059a5dffa7f018d59acead0a2e0d874e222fdb10f720f5b3ae1a6d9acR173-R176) finds the right package in the manifest.

antonbaliasnikov: Good, thanks. I just checked it doesn't work locally in my case when I have several workspaces in different directories.

MarcoIeni: even when using `release-plz --manifest-path`? If so, we should fix it

antonbaliasnikov: Sorry for that. I've just executed `release-plz` from the subdirectory without `--manifest-path`. If I return back to the repository root and use `--manifest-path`, then everything works. Sorry for the confusion!

MarcoIeni: still, we should have a better error message for it! :)

eviltak: You do bring up a good point though - the code does not support repositories that do not have a `Cargo.toml` in the repository root. I can amend the code to use the same relative path to find the manifest as in the current commit, but like the git tag template, this has the same issue of requiring the relative path of the manifest in the repository to be the same forever. In other words, you can never move the workspace root.

For now, I can make this change and log an error if the manifest root cannot be found in the same relative path; in a follow-up issue, it might be worth exploring the possibility of adding a _repository_ release-plz config (or just a small file) that stores the relative path to the manifest. There are likely better ways of keeping track of where the workspace root is (maybe just find the uppermost `Cargo.toml` file?) but I'll leave it for a later discussion.

crates/release_plz_core/tests/all/git_only.rs
eviltak: Is there a way I can share this utility function with the release_plz tests crate as well? If nothing else, I can copy the `touch` function over.

eviltak:
@MarcoIeni I have been distracted by work and the holiday season, but I think I have finally reached a point where this PR is ready for review again. Should you choose to start reviewing right now, please note that there are a couple of small outstanding items as listed in the PR description, but they are pretty minor and shouldn't impact the PR as a whole too much. The PR branch itself also has a conflict that needs resolving, but as before it shouldn't affect the end result much. The following is a summary of the changes since the last review:

- The `git_only` option has been removed. This PR now introduces _no new_ configuration options, and all new functionality is introduced transparently (as in: existing release-plz users should not need to make any changes to their workflow or their configs).
- Published package retrieval has been refactored a bit to support multiple published package sources (for example, registries and git tags).
- Published packages obtained from git tags are stored in a format similar to that of `cargo package` to better support the existing package diff mechanisms.
- Some miscellaneous bugs discovered by the tests introduced in the PR have also been squashed.

> To make this change less disruptive for existing release-plz users, if `publish = true` I will keep the existing behavior (where the registry takes precedence over the git tags in picking the next version), and maybe log a warning that says that there's a git tag with a newer version.

This is currently the only outstanding functionality item. Implementing it will require a breaking change to the `release_plz_core` library (mainly because it requires the addition of a new field in the `ReleaseMetadata` struct), so if this is something we possibly going to revert in the future I would rather not do it in the first place. 

As far as I understand, in the case where a user is publishing packages to a registry via `release-plz` (i.e. has `publish = true` in the release-plz config), the only way it is possible for a git tag version to take precedence over the registry version is if a user either:
- Manually creates a git tag that conforms with the configured release-plz tag template at a newer unreleased version of the package; or
- Temporarily disables registry publishing via release-plz (i.e sets `publish = false` in the release-plz config file) and releases a package before setting `publish = true` again (this is pretty much the first option, but using release-plz to create the tag rather than doing it manually).

To me, both cases seem obscure enough to not warrant the effort (and the addition of the only breaking change to this PR) to differently handle this case of `publish = true`. On the other hand, though it would require a breaking change to `release_plz_core`, implementing this should be relatively straightforward. As such, I would slightly prefer not handling this special case, but am otherwise indifferent to the approach we choose. What do you think?

MarcoIeni:
I don't care about breaking changes in release_plz_core.

We are the only one using in on crates.io

PigeonF:
One thing to watch out for (especially since the solution is "no changes to config necessary"): previously packages with `package.publish = false` in their Cargo.toml [were disregarded for publishing](https://github.com/release-plz/release-plz/pull/1872/files#diff-9a84559c1216068a78c1fb386f9471251ac3947dc69962ca618b96e4c6753d09L534-R534), which means that if `git_release_enable = true`  and  `git_tag_enable = true` in the release-plz config, the new version will create a new git tag / release for the packages that had `publish = false` up until now.

This might be undesired, so this change probably still constitutes a breaking change with regards to the configuration (I personally prefer the new behaviour since it more closely matches the configuration options' names, and don't think it should be changed, but the new behavior should definitely be documented).

The solution is to explicitly mark the packages as `git_release_enable = false` and `git_tag_enable = false`

```toml
#  release-plz.toml

# ...

[[package]]
name = "xtask"
git_release_enable = false
git_tag_enable = false
```

crates/release_plz_core/src/published_packages/git_tags.rs
```diff
@@ -0,0 +1,368 @@
...
+            let lockfile_contents = cargo::ops::resolve_to_string(&resolve_ws, &new_resolve)?;
+            let cargo_lock_file_name = Utf8Path::new(CARGO_LOCK).to_path_buf();
+            fs_err::write(target_dir.join(&cargo_lock_file_name), &lockfile_contents)?;
+
+            published_package_files.push(cargo_lock_file_name);
+            tracing::trace!("wrote {CARGO_LOCK} contents:\n{}", lockfile_contents);
+        }
+
+        Ok(published_package_files)
+    }
```
PigeonF, line 367: This part of the code seems to be missing some logic for resolving workspace references. When trying the current version of the code, the `release-pr` subcommand fails when the publish package has workspace references.

None of the workspace references in `[package]` or `[dependency]` seem to resolve. There is also no difference between the `Cargo.toml` and the `Cargo.toml.orig` in the generated tmp folder. This leads to a failure in <https://github.com/release-plz/release-plz/blob/af0b46b6babcb69b4ae6857a01227dd4b4868d50/crates/release_plz_core/src/published_packages/git_tags.rs#L215-L220> because cargo cannot read the workspace root.

eviltak: Ah yeah, I chose to skip the normalization Cargo performs because it also removes path dependencies, but missed this possibility. I'll look into adding the manifest normalization without touching path dependencies, and also add workspace references to a test so this behavior is tested for. Thanks for pointing that out!

MarcoIeni:
> To me, both cases seem obscure enough to not warrant the effort (and the addition of the only breaking change to this PR) to differently handle this case of `publish = true`. On the other hand, though it would require a breaking change to `release_plz_core`, implementing this should be relatively straightforward. As such, I would slightly prefer not handling this special case, but am otherwise indifferent to the approach we choose. What do you think?

So your preference is to consider a version as published if there's a tag for it, even if the version isn't in crates.io?

I'm fine with that and actually it could mitigate an issue we have with yanked crates.
Let's see if too many people complain when they see their release PRs 😅
The only ask I have is to log the reason why the crate has been considered published very explicitly. I would say a warning maybe 🤔

website/docs/faq.md
```diff
@@ -1,19 +1,31 @@
...
 
-Release-plz publishes all packages, except:
+Release-plz can release packages in multiple ways, such as:
+
+- Publishing to a Cargo registry
+- Creating git tags
+- Creating repository releases
+
+By default, release-plz considers all packages in a workspace for releasing.
+The following are not published to a Cargo registry:
```
MarcoIeni, line 12:
```suggestion
By default, release-plz considers all packages in a workspace for releasing.
The following are not published to a Cargo registry:
```
Sorry but I strongly disagree with this.

If a package has `publish = false` in the `Cargo.toml` we shouldn't have them in the release PR and we shouldn't create a git tag for it UNLESS there's an explicit `git_tag_enable = true` for that crate (i.e. workspaces would require a `[[package]]` override). If this is too boring, we could consider adding a new config flag for this in another PR.

Otherwise this will be a huge breaking change for every existing user because people will start seeing tags for crates that until now where ignored in their workspace.

What do you think?

MarcoIeni (commented):
I don't have time to read the code today, but I looked at the docs changes 👍

crates/release_plz_core/src/published_packages/git_tags.rs
eviltak: This turned out to be easier than I thought - the path dependency stripping happens in a separate step right before packaging, so I was able to just used the normalized manifest generated by `cargo` when loading a manifest. This should be fixed according to the tests I added, but can you please run your workflow again to ensure that it works?

eviltak:
> So your preference is to consider a version as published if there's a tag for it, even if the version isn't in crates.io?

Yes, that's correct. The assumption is that if a git tag matching the git tag template exists, it would have been created during a `release-plz release` call, implying that the user already released the package version at that tag. Naturally, this behavior is required for packages that aren't published to the registry.  In fact, I would even say that this behavior is is good to have in general - your example of yanked versions (i.e. if the latest version in the registry is yanked) is a good illustration of why.

Of course, it goes the other way around too - if a version exists in the registry but doesn't have a tag (e.g. if `git_tag_enable` is false), it is also considered published. 

> The only ask I have is to log the reason why the crate has been considered published very explicitly. I would say a warning maybe 🤔

If you mean log the source (git tags vs registry) used to obtain the latest published package, that should be fairly straightforward. Since `release-plz` shows INFO traces by default, it might be better to log it at the info level - logging at the warn level might incorrectly imply that something went wrong. I don't mind going either way, though.

## Breaking change for `package.publish = false` packages

(For coherence, replying to your review comment https://github.com/release-plz/release-plz/pull/1872#discussion_r1909430182 about this below)

> Otherwise this will be a huge breaking change for every existing user because people will start seeing tags for crates that until now where ignored in their workspace.

This is certainly possible, and I agree that this is a big enough change to be considered as breaking. However, I would argue that this breakage is warranted and shouldn't be handled differently (beyond documenting the changed behavior).

### Why (I think) we should keep the breaking change as-is

- This makes the behavior of release-plz more closely match the configuration options (as brought up by @PigeonF a couple days ago in https://github.com/release-plz/release-plz/pull/1872#issuecomment-2577915703). It was always possible to release packages via release-plz without publishing them to the registry, so releasing packages that cannot be published to the registry (i.e. have `package.publish = false`) via release-plz should also be possible.
- Replicating the old behavior (i.e. disabling release-plz processing) for packages with `package.publish = false` is simple - just set `release = false` in the release-plz config for those packages.
- Since this breaking change would cause a major version bump, I think it is safe to assume that a user would have been aware of any breakages (assuming they are properly documented) when upgrading to a new major version of release-plz - as with any library or executable.

With all that being said, as the maintainer it is your decision that ultimately chooses how we proceed. If you still think we should avoid the breaking change, I have some suggestions.

### Avoiding the breaking change

There is a pretty straightforward way to ensure that we don't release packages with `package.publish = false` (henceforth referred to as "unpublishable packages"). As mentioned above, all we need to replicate the old behavior (and thus not break anything) with unpublishable packages is to mark them as `release = false` in the config. To do this transparently to the user, we just change the configuration so that it assumes a default override of **`false`** (rather than taking the workspace default of `true`) for the `release` config option for unpublishable packages.

In terms of implementation, this would mean making the `release` field in the `UpdateConfig` and `ReleaseConfig` an `Option<bool>` (so we know when it is unspecified), changing `ReleaseMetadataBuilder::get_release_metadata` to take a `&Package` (so we can check if the package is unpublishable or not), and in that method returning `Some` only if the `release` field in the package config was explicitly set to true or if the package is publishable.

crates/release_plz_core/src/published_packages/git_tags.rs
PigeonF: With the new changes (`843c55c2781bf460a0f1d6e10de08bf57bdbc19c`) my workflow works (for reference, I do not publish to a registry (all cargo manifests have `package.publish = false`), and have one package that I do not want git tags for.

The relevant excerpts from my release-plz config look like

```toml
[workspace]
publish = false
release = true

[[package]]
name = "workspace-hack"
release = false
```

I have not tested without setting `workspace.publish` to `false`, let me know if you want me to

eviltak: Thanks! If you can test with `workspace.publish` set to false, that would also be great. It should work without any issues, but would still be good to confirm.

eviltak:
> usually if a package has `publish = false` people don't even care about updating its version.

While I don't fully agree with this personally, it is entirely reasonable, and you definitely have a better idea of the intended usage of this tool than I do. Perhaps it's because I envision using release-plz to manage and release versions in many scenarios with unpublishable packages - contrary to the initial inspiration behind release-plz of simplifying the workflow of updating and publishing crates to registries.

As such, I am fully on board with ensuring that existing workflows with unpublished packages do not break because of this change. I think the main thing needed to implement the `release` default override for unpublishable packages is  just knowing where the `release` field is coming from (workspace vs package-specific config) - kind of like [`git config --show-origin`](https://git-scm.com/docs/git-config#Documentation/git-config.txt---show-origin). 

I imagine this might not be too difficult to achieve - we just need to store some additional metadata for the `release` field (probably by way of storing it in a `ConfigField<bool>` struct or similar), and specify the source when doing a `PackageConfig::merge`. I'll start looking into this and implementing it if it's not too difficult. In the meantime, if you or anyone else have any better ideas, I'm all ears!

> * adding a `release_unpublishable` or similar that you set in the config that you set once in `[[workspace]]` to enable git tags and so on for `publish = false` packages

This is a great idea! Shouldn't be too difficult to add in this PR in addition to the `release` field override, and would make things easier for people with large workspaces with primarily unpublishable packages.

MarcoIeni:
> Shouldn't be too difficult to add in this PR in addition to the `release` field override, and would make things easier for people with large workspaces with primarily unpublishable packages.

I would prefer to have it in a separate PR so we can discuss better the name of the new config, the docs, etc.

This PR is already big enough and I'm sure it will be challenging to review as already is. I'll try to find some time soon to do it!

> Perhaps it's because I envision using release-plz to manage and release versions in many scenarios with unpublishable packages - contrary to the initial inspiration behind release-plz of simplifying the workflow of updating and publishing crates to registries.

Yes, this was my original idea, but it's great when people extend your ideas further. Thanks for working on this!

crates/release_plz_core/src/published_packages/git_tags.rs
PigeonF: Without setting `[workspace.publish]` in the `release-plz.toml`, the `package.publish` from the `Cargo.toml` seems to be respected 👍 (though I can only test the `package.publish = false` case, since I don't have  any packages that publish to a registry)

crates/release_plz_core/src/published_packages/git_tags.rs
```diff
@@ -0,0 +1,459 @@
...
+        crate::tera::render_template(&escaped_partial_template, &context, "release_tag_regex");
+    // Add anchors to ensure regex matches whole string
+    let release_tag_regex =
+        Regex::new(&format!("^{release_tag_regex}$")).expect("invalid rendered version tag regex");
+
+    tags.filter_map(move |tag| {
+        // Check if the tag name matches the regex
+        let (_, [version_str]) = release_tag_regex.captures(tag)?.extract();
+        // Check if the captured version string can be parsed as a package version
+        Some((tag, Version::parse(version_str).ok()?))
```
MarcoIeni, line 166: I wonder if we could move some filters like this one when we retrieve the tags for the first time (e.g. in the `get_all_tags` function).

So that we run this filter only once instead of running it for each package. We could gain a bit of performance in this way.

crates/release_plz/tests/all/release.rs
```diff
@@ -128,6 +129,12 @@ async fn release_plz_releases_a_new_project_with_custom_release() {
     "#;
     context.write_release_plz_toml(config);
 
+    // Test publish = true in release-plz config but publish = false in manifest
+    let cargo_toml_path = context.repo_dir().join("Cargo.toml");
+    let mut cargo_toml = LocalManifest::try_new(&cargo_toml_path).unwrap();
+    cargo_toml.data["package"]["publish"] = false.into();
+    cargo_toml.write().unwrap();
```
MarcoIeni, line 136: could we also test that no packages are published to the registry in this case?

You can download packages from gitea like this:
https://github.com/eviltak/release-plz/blob/843c55c2781bf460a0f1d6e10de08bf57bdbc19c/crates/release_plz/tests/all/changelog.rs#L70

crates/release_plz/tests/all/release_pr.rs
```diff
@@ -315,6 +314,96 @@ async fn release_plz_doesnt_add_invalid_labels_to_release_pr() {
...
+
+    context.merge_release_pr().await;
+    // The commit contains the PR id number
+    let expected_commit = format!("{expected_title} (#1)");
+    assert_eq!(
+        context.repo.current_commit_message().unwrap(),
+        expected_commit
+    );
+
+    context.run_release().success();
```
MarcoIeni, line 352: Similar as the other comment: should we test that the package was not published in the registry?

MarcoIeni (commented):
some first comments 👍

crates/release_plz_core/src/cargo.rs
```diff
@@ -213,3 +215,20 @@ pub async fn wait_until_published(
 
     Ok(())
 }
+
+/// Creates a new Cargo [`GlobalContext`] in the specified working directory `cwd`.
+///
+/// If `cwd` is [`None`], [`GlobalContext::default`] is returned.
+pub fn new_global_context_in(cwd: Option<Utf8PathBuf>) -> anyhow::Result<GlobalContext> {
```
MarcoIeni, line 222: why the ending `_in`? 🤔

crates/release_plz_core/src/clone/mod.rs
MarcoIeni, line 1: I see that this file contains both refactorings and new functionalities.

I feel like it would be easier to review for me if you raise one PR for the refactorings and we merge that before this one.

What do you think? Would it be too tedious to do for you?

Feel free to also include other refactorings you have done in other files 👍

crates/release_plz_core/src/command/release.rs
```diff
@@ -739,17 +748,41 @@ async fn release_package(
...
+                release_info.package.name, release_info.package.version
+            );
+        }
+
+        if !publish && !create_git_tag && !create_git_release {
+            info!(
+                "{} {}: no release method enabled",
+                release_info.package.name, release_info.package.version
+            );
+        }
```
MarcoIeni, line 777: could you extract this into a separate function?

crates/release_plz_core/src/next_ver.rs
```diff
@@ -395,33 +405,40 @@ impl ReleaseMetadataBuilder for UpdateRequest {
...
+    let repo_is_clean_result = repository.repo.is_clean();
     if !input.allow_dirty {
-        repository.repo.is_clean()?;
+        repo_is_clean_result?;
+    } else if repo_is_clean_result.is_err() {
+        // Commit the uncommitted changes in a temporary commit, so we can use `checkout_head`
+        // to restore them
+        repository
+            .repo
+            .add_all_and_commit("release-plz temp commit with uncommitted changes")?;
```
MarcoIeni, line 426: is this change unrelated to the new functionality? If yes, could you move it into its own PR?

Sometimes really weird error happen, so it's important to minimize the amount of changes that go into a single PR.

I squash the PRs as a convention so the git history will be deleted once I merge the PR unfortunately. That's because I prefer small PRs to long PRs with a well organized commit history (I know many people disagree, that's just my preference with this project).

crates/release_plz_core/src/clone/mod.rs
eviltak: I can take a look - I think I did make the commits granular enough so they can be extracted into separate PRs as needed. Given that you prefer to squash merge PRs, I'll wait till this PR is ready to "merge" (i.e. is feature complete and has no bugs) before I split it up into as many smaller PRs as I can, so we can avoid having to make follow-up PRs. Does that work for you?

crates/release_plz_core/src/next_ver.rs
eviltak: All changes (barring one or two minor quality of life changes) I have included in this PR are required to not break functionality. However, it should be possible for me to extract them into separate smaller PRs, since most if not all of them have their own individual commits. 

This particular change is required because the code that generates published packages from git tags first has to check out the git tag to ensure it is valid. If the repo has uncommitted changes, that checkout would fail (or at best, discard the uncommitted changes), hence why I chose to commit them here.

This change has it's own commit, and if you are fine with PRs containing very small few-line changes (which sounds like it would fit with your squash-merge philosophy), then I can extract it out to its own PR. However, as I mentioned above, I would prefer to do this after this PR is done and out of draft status, in case one of these smaller fixes needs to be changed/removed for reasons we haven't discovered yet. Is this okay with you?

crates/release_plz_core/src/published_packages/git_tags.rs
eviltak: This may not be possible with the code as-is - because the git tag template may contain the package name, every package could have a unique `release_tag_regex`. The goal of this filtering is to find the release git tags that correspond to a specific package. 

Perhaps a faster way could be to make the regex capture the package name in addition to the version string - this will let us precompute the list of all release tags for any package in the repository. We will still have to do some per package filtering, but it would simply be a name equality test, which should be faster than compiling a new regex every time.

crates/release_plz_core/src/clone/mod.rs
MarcoIeni: Yes, sure it does, thanks!

crates/release_plz_core/src/published_packages/git_tags.rs
MarcoIeni: Don't bother if it's too complex of it makes the code way more complex. I thought this was a quick win, but if it's not you can leave the code as it is 👍

crates/release_plz_core/src/next_ver.rs
MarcoIeni: Yes, thanks 👍

eviltak:
Having fixed the unpublished package breaking change issue, I think this PR has finally reached a point where I can start splitting it up. There are a couple of tasks yet to be completed, but those are specific to this PR and shouldn't affect the new PRs. At this point I don't need you to fully review this PR, but I would love it if you could at least take a cursory look and/or test it manually, in case I missed something else.

In the meantime, I shall compile below a list of the separate PRs along with the commits they shall include.

### fix(core): Use dunce to canonicalize package paths
- eviltak@800ea3c

### fix: Make release-pr tests Use UTC rather than local timezone
- eviltak@b462f8f

### feat(git_cmd): Add worktree management methods
- eviltak@d17ca60
- TODO: Add method for worktree removal (being done in `release_plz_core::published_packages::git_tags`)

### feat(git_cmd): Add method to retrieve all tags in repo
- eviltak@1923665

### fix(core): Fetch latest package summaries from cargo registry
- eviltak@d9d66c3

### feat: Release non-publishable packages
- eviltak@dc440ae
- eviltak@5718482
- eviltak@5b50b10
- eviltak@a26d930

### fix(core): Make release dry-run more informative
- eviltak@11a1171
- TODO: Extract into method (https://github.com/release-plz/release-plz/pull/1872#discussion_r1912001647)

### fix(core): Refactor published package retrieval
- Depends on "Fetch latest package summaries from cargo registry"
- eviltak@80d16ec
- Remove the git tags published package parts from:
  - eviltak@4a01c78
  - eviltak@442c9e2
- TODO: Add a `Summary::descriptive_source_name` method and log the name of the source of the selected published package

marcoieni:
this feature was implemented in https://github.com/release-plz/release-plz/pull/2512
Thanks anyway for your efforts!

eviltak:
@marcoieni apologies for basically abandoning this PR. Had to prioritize other things at work, and I soon after switched jobs, moved across the country and completely forgot about this.

Glad to see this feature was implemented!

PR #1899, opened 2024-12-09
fix: dev dependencies should affect release order

cfsnyder:
I have a cargo workspace project with multiple packages and encountered build errors when dev dependencies from the same workspace were not released before the packages that depended on them. The original code had a comment indicating that this dependency ordering constraint was not significant:

> Ignore development dependencies. They don't need to be published before the current package...

I'm not sure if the Cargo publish behavior has changed or if it dev dependencies are only significant with certain project configurations. One theory is that dev dependencies must be resolved if they're used by integration tests vs. just unit tests, but I have not verified.

### DIFF
--- a/crates/release_plz_core/src/release_order.rs
+++ b/crates/release_plz_core/src/release_order.rs
@@ -1,4 +1,4 @@
-use cargo_metadata::{Dependency, DependencyKind, Package};
+use cargo_metadata::Package;
 use tracing::debug;
 
 /// Return packages in an order they can be released.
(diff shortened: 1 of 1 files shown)


cfsnyder:
I see there are some failing tests related to this scenario. I will look at those more closely and see if I can create a reproducer for my scenario. Curious if you have any thoughts/feedback on this issue though @MarcoIeni ?

MarcoIeni:
What's the behavior of https://github.com/crate-ci/cargo-release ?

MarcoIeni:
I'm closing this PR because I don't know if this is the correct behavior. Closing PRs that aren't actionable for me helps me reducing the cognitive load of maintaining this project. I hope you understand.

Please open a new issue if the release-plz release order is different wrt `cargo release` or 'cargo publish --workspace`, which is now available on nightly.


[Here](https://github.com/rust-lang/cargo/issues/1169#issuecomment-2540367442) you can find how to test cargo publish 👍

Issue #1881, opened 2024-12-02
Implement PR labels for Gitea


MarcoIeni:
like https://github.com/release-plz/release-plz/pull/1879 but for gitea.

Also add an integration test.

- [x] Note that it might be needed to create the label via API before adding it. This might be needed also for github. If this is true, create the label if it doesn't exist. To minimize the number of API calls needed, you should read the error message or status code of the add_labels API call to understand if creating a label is needed.

EleisonC:
hey @MarcoIeni I don't mind working on this if it's still available

MarcoIeni:
Good choice, this is a fun one!
I assigned it to you 😁

MarcoIeni:
> Note that it might be needed to create the label via API before adding it. This might be needed also for github.

this isn't needed for github

> To minimize the number of API calls needed, you should read the error message or status code of the add_labels API call to understand if creating a label is needed.

we didn't do this but it's not an issue imo. If it is for anyone, pr welcome

PR #1937, opened 2024-12-27
feat: allow adding labels to Gitea release PR

EleisonC:
### Add labels Gitea Enhancement.
Close [#1881](https://github.com/release-plz/release-plz/issues/1881)

1. Fixed Typo in Struct Name:

- Renamed the struct GiteaRelase to GiteaRelease to correct a spelling error.

2. Pull Request Labels Management:

- New Structs and Fields:
    - Introduced a Label struct with name and ID fields.
    - Added a labels field to existing structs like GitPr and GitLabMr to accommodate label information.

- Label Validation:
    - Implemented validation to ensure label names are non-empty and do not exceed 50 characters, providing appropriate error messages for invalid inputs.

- Gitea Integration:
    - Introduced a GiteaLabelResponseStruct to handle responses from Gitea's label API.
    - Enhanced label management for Gitea backends by:
        -   Fetching existing labels.
        -   Creating new labels if they don't exist.
        -   Mapping labels to their respective IDs.
        -   Assigning labels to PRs using these IDs.
        
3. Token Scope Enhancement:

    - Updated token creation to include read:issue and write:issue permissions, facilitating the new label management functionalities.
    
4.  Testing:
Added test cases to verify label functionalities:
release_plz_add_labels_to_release_pull_request: Assesses the addition of labels to a release PR.
release_plz_add_invalid_labels_to_release_pr: Evaluates the validation process for invalid labels.

### DIFF
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -279,7 +279,7 @@ async fn release_plz_doesnt_add_invalid_labels_to_release_pr() {
             [workspace]
             pr_labels = [" "]
         "#,
-            "Empty labels are not allowed",
+            "leading or trailing whitespace is not allowed",
         ), // space label
         (
             r#"
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -294,7 +294,14 @@ async fn release_plz_doesnt_add_invalid_labels_to_release_pr() {
             pr_labels = [""]
             "#,
             "Empty labels are not allowed",
-        ), // empty label
+        ),
+        (
+            r#"
+            [workspace]
+            pr_labels = ["abc", "abc"]
+            "#,
+            "duplicate labels are not allowed",
+        ),
     ];
 
     for test_case in test_cases {
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -218,7 +218,7 @@ async fn changelog_is_not_updated_if_version_already_exists_in_changelog() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
-async fn release_plz_add_labels_to_release_pull_request() {
+async fn release_plz_adds_labels_to_release_pr() {
     let test_context = TestContext::new().await;
 
     // Initial PR setup with two labels
@@ -270,7 +270,7 @@ async fn release_plz_add_labels_to_release_pull_request() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
-async fn release_plz_add_duplicate_labels_to_release_pr() {
+async fn release_plz_adds_deduplicated_labels_in_release_pr() {
     let test_context = TestContext::new().await;
     let duplicate_config_test_cases = r#"
             [workspace]
@@ -294,7 +294,7 @@ async fn release_plz_add_duplicate_labels_to_release_pr() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
-async fn release_plz_add_invalid_labels_to_release_pr() {
+async fn release_plz_doesnt_add_invalid_labels_to_release_pr() {
     let test_context = TestContext::new().await;
     let test_cases: &[(&str, &str)] = &[
         // (label config, expected error message)
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -270,30 +270,6 @@ async fn release_plz_adds_labels_to_release_pr() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
-async fn release_plz_adds_deduplicated_labels_in_release_pr() {
-    let test_context = TestContext::new().await;
-    let duplicate_config_test_cases = r#"
-            [workspace]
-            pr_labels = ["feature", "Feature", "FEATURE"]
-            "#; // initial release pr
-    let duplicate_config2_test_cases = r#"
-            [workspace]
-            pr_labels = ["feature", "FEATURE", "Bug", "urgent"]
-            "#; // update release pr
-
-    test_context.write_release_plz_toml(duplicate_config_test_cases);
-    test_context.run_release_pr().success();
-    let mut prs = test_context.opened_release_prs().await;
-    assert_eq!(prs[0].labels.len(), 3, "Expected 2 labels to get created");
-
-    test_context.write_release_plz_toml(duplicate_config2_test_cases);
-    test_context.run_release_pr().success();
-    prs = test_context.opened_release_prs().await;
-    assert_eq!(prs[0].labels.len(), 5, "Expected 3 labels to get created");
-}
-
-#[tokio::test]
-#[cfg_attr(not(feature = "docker-tests"), ignore)]
 async fn release_plz_doesnt_add_invalid_labels_to_release_pr() {
     let test_context = TestContext::new().await;
     let test_cases: &[(&str, &str)] = &[
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -284,12 +284,12 @@ async fn release_plz_adds_deduplicated_labels_in_release_pr() {
     test_context.write_release_plz_toml(duplicate_config_test_cases);
     test_context.run_release_pr().success();
     let mut prs = test_context.opened_release_prs().await;
-    assert_eq!(prs[0].labels.len(), 1, "Expected 2 labels to get created");
+    assert_eq!(prs[0].labels.len(), 3, "Expected 2 labels to get created");
 
     test_context.write_release_plz_toml(duplicate_config2_test_cases);
     test_context.run_release_pr().success();
     prs = test_context.opened_release_prs().await;
-    assert_eq!(prs[0].labels.len(), 3, "Expected 3 labels to get created");
+    assert_eq!(prs[0].labels.len(), 5, "Expected 3 labels to get created");
 }
 
 #[tokio::test]
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -270,6 +270,30 @@ async fn release_plz_add_labels_to_release_pull_request() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
+async fn release_plz_add_duplicate_labels_to_release_pr() {
+    let test_context = TestContext::new().await;
+    let duplicate_config_test_cases = r#"
+            [workspace]
+            pr_labels = ["feature", "Feature", "FEATURE"]
+            "#; // initial release pr
+    let duplicate_config2_test_cases = r#"
+            [workspace]
+            pr_labels = ["feature", "FEATURE", "Bug", "urgent"]
+            "#; // update release pr
+
+    test_context.write_release_plz_toml(duplicate_config_test_cases);
+    test_context.run_release_pr().success();
+    let mut prs = test_context.opened_release_prs().await;
+    assert_eq!(prs[0].labels.len(), 1, "Expected 2 labels to get created");
+
+    test_context.write_release_plz_toml(duplicate_config2_test_cases);
+    test_context.run_release_pr().success();
+    prs = test_context.opened_release_prs().await;
+    assert_eq!(prs[0].labels.len(), 3, "Expected 3 labels to get created");
+}
+
+#[tokio::test]
+#[cfg_attr(not(feature = "docker-tests"), ignore)]
 async fn release_plz_add_invalid_labels_to_release_pr() {
     let test_context = TestContext::new().await;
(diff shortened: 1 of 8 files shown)


crates/release_plz/tests/all/release_pr.rs
MarcoIeni, line 1: these tests are great 🙌

EleisonC: Thank you. Is there anything else you want me to change, or is this good to go?

MarcoIeni: I'm still reviewing. Sorry for not specifying 👍

crates/release_plz_core/src/git/backend.rs
```diff
@@ -147,6 +160,7 @@ pub struct GitLabMr {
     pub source_branch: String,
     pub title: String,
     pub description: String,
+    pub labels: Vec<Label>,
```
MarcoIeni, line 163: are you sure this is how the gitlab merge request response looks like?
[Here](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr) it looks like it's just a vector of strings.

If you confirm, can you fix this?

crates/release_plz_core/src/git/backend.rs
```diff
@@ -156,10 +170,10 @@ pub struct GitLabAuthor {
...
 impl From<GitPr> for GitLabMr {
     fn from(value: GitPr) -> Self {
-        let description = match value.body {
-            Some(d) => d,
-            None => "".to_string(),
-        };
+        let mut desc = "".to_string();
+        if let Some(body) = value.body {
+            desc = body;
+        }
```
MarcoIeni, line 176: this is just a refactoring, right?

If so, we could use 

```
        let desc = value.body.unwrap_or_default();
```

it should be equivalent, right?

crates/release_plz_core/src/git/backend.rs
```diff
@@ -540,23 +555,37 @@ impl GitClient {
         };
 
         info!("opened pr: {}", git_pr.html_url);
-        self.add_labels(pr, git_pr.number)
-            .await
-            .context("Failed to add labels")?;
+        if pr.labels.is_empty() {
+            warn!("No labels provided for PR #{}", git_pr.number);
```
MarcoIeni, line 559: why emitting a warning if no pr labels are provided?
I think it's just noise for people who don't use pr labels

crates/release_plz_core/src/git/backend.rs
```diff
@@ -579,7 +600,93 @@ impl GitClient {
...
+                } else {
+                    self.client
+                        .post(format!("{}/{}/labels", self.issues_url(), pr_number))
+                        .json(&json!({ "labels": label_ids }))
+                        .send()
+                        .await?
+                        .successful_status()
+                        .await?;
+                }
+
```
MarcoIeni, line 689: can you extract this code in a separate function (or multiple)?

crates/release_plz_core/src/git/backend.rs
```diff
@@ -726,6 +833,19 @@ impl GitClient {
...
+        if l.len() > 50 {
+            anyhow::bail!("Failed to add label `{l}`. It exceeds maximum length of 50 characters.");
+        }
+
+        if l.trim().is_empty() {
+            anyhow::bail!("Failed to add label. Empty labels are not allowed.");
+        }
+    }
+    Ok(())
+}
```
MarcoIeni, line 847: can we move the validation when we read the configuration? So that we fail earlier.

crates/release_plz_core/src/git/backend.rs
EleisonC: I am pushing a fix for this now

crates/release_plz_core/src/git/backend.rs
EleisonC: Yes Just a refactor and i have changed it to  `let desc = value.body.unwrap_or_default();` that looks more idiomatic

MarcoIeni: nevermind, I pushed that change already 👍
Sorry if I also left this comment. But it's good that you checked that this refactor looks correct to you as well 👍

MarcoIeni:
- [ ] update the config docs to say that labels are also available on gitea

Reminder for myself too 👍

crates/release_plz_core/src/git/backend.rs
EleisonC: I moved this to the release_pr command section

crates/release_plz_core/src/git/backend.rs
EleisonC: Pushed refactor for this

crates/release_plz_core/src/git/backend.rs
```diff
@@ -540,49 +563,163 @@ impl GitClient {
...
-                        "labels": pr.labels
-                    }))
-                    .send()
-                    .await?
-                    .successful_status()
+            BackendType::Github => self.post_github_labels(labels, pr_number).await,
+            BackendType::Gitlab => self.post_gitlab_labels(labels, pr_number).await,
+            BackendType::Gitea => {
+                let (labels_to_create, mut label_ids) = self
+                    .get_pr_info_and_categorize_labels(pr_number, labels)
```
MarcoIeni, line 583: here you assume that if a label is not present in PR number `pr_number` we need to create the label.

But what if the label already exists? Does Gitea return an error in this case?
If yes, shouldn't we check if the label already exist via their API?

For example, the user could have created this label for other purposes and they want to reuse it for the first time for the release PR

MarcoIeni: https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#list-labels-for-a-repository

you could try this (maybe it works for gitea too). I haven't found the gitea docs in their swagger

crates/release_plz_core/src/git/backend.rs
```diff
@@ -540,49 +563,163 @@ impl GitClient {
...
+            .post(self.pr_labels_url(pr_number))
+            .json(&json!({ "labels": label_ids }))
+            .send()
+            .await?
+            .successful_status()
+            .await?;
+        Ok(())
+    }
+
+    async fn get_pr_info_and_categorize_labels(
```
MarcoIeni, line 645: can you document the return type of this function?

MarcoIeni: Also, I wonder if we need this function at all.

Does Gitea complaint if we try to add a label that already exists in the issue?
If yes, we should capture the returned status code in post_gitea_labels function and add a context to the error saying that there's a bug in release-plz

crates/release_plz_core/src/git/backend.rs
```diff
@@ -105,13 +106,26 @@ pub struct GitPr {
...
+        self.labels.iter().map(|l| l.name.as_str()).collect()
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Label {
+    pub name: String,
+    /// ID of the label.
+    /// Used by Gitea and GitHub. Not present in GitLab responses.
+    id: Option<u64>,
```
MarcoIeni, line 128: it would be cool if at compile time with the type system we could have u64 for GitHub and Gitea and Option<u64> for GitLab. But now I don't know how. So if you have a good idea we could do it in another pr :)

crates/release_plz_core/src/git/backend.rs
```diff
@@ -540,49 +563,163 @@ impl GitClient {
...
+        let mut label_ids = Vec::new();
+
+        for label in labels_to_create {
+            info!("Backend Gitea Creating label: {}", label);
+            let res = self
+                .client
+                .post(format!("{}/labels", self.repo_url()))
+                .json(&json!({
+                    "name": label.trim(),
+                    "color": "#FFFFFF"
```
MarcoIeni, line 685: is the color mandatory? If no, we can leave it to gitea to pick a random color. Otherwise, we can write a comment saying that the color is mandatory 👍

crates/release_plz_core/src/git/backend.rs
```diff
@@ -540,49 +563,163 @@ impl GitClient {
...
+                }))
+                .send()
+                .await?
+                .successful_status()
```
MarcoIeni, line 689: the successful_status function returns an error in case of NOT_FOUND and UNPROCESSABLE_ENTITY so those match arms are unreachable because in the next line you use ?

I think you should match on the error, and then match on the status.
Then when you match on the error, add context instead of creating a new error, so still show the error returned from Gitea to the user, as it might contain useful info.

MarcoIeni:
I reviewed all the code, so these should be the last comments.
Thank you for your patience and for addressing all my comments, I really appreciate it 🙌 I hope you enjoy the process 😁

crates/release_plz_core/src/git/backend.rs
EleisonC: I see your point and i am going to change to
- getting all labels at the repository level not Pull request 
- this should solve for situations where ```For example, the user could have created this label for other purposes and they want to reuse it for the first time for the release PR```. By the time we are creating new labels it for the repository. 
Also, avoid duplicate labels.

crates/release_plz_core/src/git/backend.rs
EleisonC: We could implement something similar to what was done with GitPr and GitLabMr by introducing GitLabel and GitLabLabel. 
Perhaps GitPr could be extended to handle situations where GitLabel would be used because we are dealing with label data returned from a request for PR info. 
Gitea & Github - returns An Object. with  `id`  as one of its keys.
Gitlab - returns only a vector of strings.
So I think extending GitPr and GitLabMr would be the start of the solution. What do you think

crates/release_plz_core/src/git/backend.rs
EleisonC: Yes the color is mandatory for gitea. [Ref here](https://docs.gitea.com/api/1.22/#tag/issue/operation/issueListLabels). Will also add which color it is

crates/release_plz_core/src/git/backend.rs
EleisonC: Adjusting the system to retrieve label information from the repository level.

> Does Gitea give an error if we attempt to add a label that already exists in the issue?

No, it does not handle duplication. I may need to implement a case check when filtering labels for creation to minimize the risk of duplication. This behaviour is similar to that of GitHub.

crates/release_plz_core/src/git/backend.rs
MarcoIeni: If you know how to remove the with_context error handling, I would say go for it! But maybe let's do it in another pr, and let's first merge this one 👍

crates/release_plz_core/src/git/backend.rs
EleisonC: Nice. Thank you for pointing this out

EleisonC:
> I reviewed all the code, so these should be the last comments. Thank you for your patience and for addressing all my comments, I really appreciate it 🙌 I hope you enjoy the process 😁

Hey @MarcoIeni, I have made the requested changes and added one extra test for duplication behaviour during the release-pr process. Let me know if there is anything else.
Yes, I did enjoy this one. The end product makes me feel proud of myself.

crates/release_plz_core/src/git/backend.rs
```diff
@@ -642,30 +638,79 @@ impl GitClient {
         Ok(())
     }
 
-    async fn get_pr_info_and_categorize_labels(
+    /// Retrieves and categorizes labels for a PR, ensuring case-insensitive deduplication
```
MarcoIeni, line 641: why do you do case-insensitive deduplication?

I just tested on gitea, and it's possible to create labels that are both lower case and upper case.

Why should release-plz forbid this?

image

EleisonC: I was looking at it from the perspective of minimizing duplication.
Assuming the PR already had the label "BUG" created and assigned to a PR.
Let's say I re-ran release-pr and passed in the pr_label - "Bug", "bug".  it won't create a new label because those words are the same.

crates/release_plz/tests/all/release_pr.rs
```diff
@@ -270,7 +270,7 @@ async fn release_plz_add_labels_to_release_pull_request() {
 
 #[tokio::test]
 #[cfg_attr(not(feature = "docker-tests"), ignore)]
-async fn release_plz_add_duplicate_labels_to_release_pr() {
+async fn release_plz_adds_deduplicated_labels_in_release_pr() {
```
EleisonC, line 273: I am changing this to` does not add`. Unless you suggest I change the implementation not to handle duplication from a case-sensitive aspect. @MarcoIeni

crates/release_plz_core/src/git/backend.rs
EleisonC: what do you think

MarcoIeni: I think that's up to gitea/github to decide. We shouldn't limit what users can do.
Plus removing this check will simplify the codebase.
So I would allow users to create labels with different capitalization 👍

EleisonC: Okay. @MarcoIeni I have made the necessary changes. And also updated the test. let me know if there is anything

MarcoIeni (approved):
Thanks 🙏 🙌

Issue #1960, opened 2025-01-07
silent fail instead of error of get_diff if the git folder of repository is invalid


gwen-lg:
## Bug description

If the data in `.git` folder of the local repository aren't valid, or problem during the copy, `release-plz` can report :
> <package> : there are no commits

instead of report the problem of invalid git repository.

- Would you like to work on a fix? y

## To Reproduce

1. clone a crate package
2. introduce error in .git by removing `*.idx` files in `.git/objects/pack/`
3. tried to run `release-plz update`
4. `there are no commits` message is reported instead of the git error.

## Expected behavior

release-plz should report an error.

## Environment

- release-plz version:  0.3.112

## Additional context

in `get_diff` when `checkout_last_commit_at_paths` return an error, the error shouldn't be ignored.

MarcoIeni:
PR welcome 👍

PR #1962, opened 2025-01-07
fix: don't suppress git error when determining commits to release

gwen-lg:
the git error from call to `checkout_last_commit_at_paths` shouldn't be silenced. Fix : #1960
If the git repository files are corrupted, we need to have an error.

I have tried to reproduce a case where we can have an error from `checkout_last_commit_at_paths` because of no commit to include in diff, but without success, to get the `there are no commits` as valid message.


marcoieni:
sorry if it took so long to review this but I just found the time to test this.

I also wasn't able to replicate that info message.

Hopefully when we ship this we don't brake anyone 😁

Issue #1961, opened 2025-01-07
release-plz faile if `.gitignore` include extension used by git (like *.idx)


gwen-lg:
## Bug description

If `.gitignore` inlude `*.idx` in the list of ignored files, the copy of  `.git` folder is incomplete, and `release-plz` don't work.

- Would you like to work on a fix? y

## To Reproduce

1. go in a rust package repository clone
2. add *.idx in `.gitignore` and commit
3. run `release-plz update`
4. changes are not processed and `there are no commits` is reported
Note: if you inspect the copy of repository, you can see than files `.git/*.idx` aren't copied, and the git repository of the copy is invalid.

## Expected behavior

The `.gitignore` rules shouldn't apply to the content of .git repository

## Environment

- release-plz version: 0.3.112

## Additional context

I wonder if the fix should be done in `release-plz` or  the `ignore` crate instead.
I have tried to add `overrides` with `.git/` and `.git/*` to WalkBuilder in `copy_directory` function, but it's doesn't work for files in subfolder of `.git`.

MarcoIeni:
If possible the fix should be in the ignore crate. It would be great to hear what the ignore maintainers think about this.

PR #1965, opened 2025-01-07
whitelist .git folder for copying local repo.

gwen-lg:
This should fix issue #1961
The `.gitignore` rules shouldn't apply for .git directory. Otherwise, we can have a broken git repository,
for example if `*.idx` is present in ignore list.


gwen-lg:
Convert to draft as it's don't work as expected.

marcoieni:
closing for inactivity to keep the issue tracker clean.
Thanks for working on this anyway.
Feel free to open a new PR if you have a new solution 👍

PR #1975, opened 2025-01-12
fix(core): Use dunce to canonicalize package paths

eviltak:
The previous implementation of `PackagePath::canonicalize` was not using `dunce::canonicalize`. This lead to Package::dependencies_to_update comparing a dunce canonicalized path to a std::fs canonicalized path.

On Windows, the former would rarely add the extended length prefix, but the latter always added the extended path length prefix. This led to `next_ver::is_dependency_referred_to_package` always returning false on Windows, causing versioned path dependencies not being updated.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,4 +1,4 @@
 disallowed-methods = [
-    { path = "cargo_metadata::camino::Utf8Path::canonicalize_utf8", reason = "use `canonicalize_utf8` instead" },
-    { path = "camino::Utf8Path::canonicalize_utf8", reason = "use `canonicalize_utf8` instead" },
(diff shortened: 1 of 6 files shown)


marcoieni:
I edited the code to forbid the use of `Utf8Path::canonicalize_utf8`. I replaces all its usages with a new function I created. What do you think?

clippy.toml
```diff
@@ -0,0 +1,4 @@
+disallowed-methods = [
+    { path = "cargo_metadata::camino::Utf8Path::canonicalize_utf8", reason = "use `canonicalize_utf8` instead" },
+    { path = "camino::Utf8Path::canonicalize_utf8", reason = "use `canonicalize_utf8` instead" },
```
eviltak, line 3: Should this be "use `fs_utils::canonicalize_utf8` instead" for more clarity?

eviltak:
Looks good to me! The addition of the lint is a great idea.

PR #1982, opened 2025-01-12
fix(core): Stash uncommited changes in dirty repos during update

eviltak:
When obtaining the next versions of the workspace, if `allow_dirty` is set to true and the repository is dirty, we commit the uncommitted changes so that they can easily be restored without being lost with a `checkout_head` (which happens a couple times during the diff process).

Note that this commit is only made in the `TempRepo` made by `Project::get_repo`, and the user's original repo remains untouched.

As per the recommendation in https://github.com/release-plz/release-plz/pull/1982#issuecomment-2601007848, this was changed to _stash_ the uncommitted changes rather than commit them.

### DIFF
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -418,8 +418,16 @@ pub async fn next_versions(input: &UpdateRequest) -> anyhow::Result<(PackagesUpd
     let repository = local_project
         .get_repo()
         .context("failed to determine local project repository")?;
+
+    let repo_is_clean_result = repository.repo.is_clean();
     if !input.allow_dirty {
-        repository.repo.is_clean()?;
+        repo_is_clean_result?;
+    } else if repo_is_clean_result.is_err() {
+        // Commit the uncommitted changes in a temporary commit, so we can use `checkout_head`
+        // to restore them
+        repository
+            .repo
+            .add_all_and_commit("release-plz temp commit with uncommitted changes")?;
     }
     let packages_to_update = updater
         .packages_to_update(&registry_packages, &repository.repo, input.local_manifest())
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -423,11 +423,14 @@ pub async fn next_versions(input: &UpdateRequest) -> anyhow::Result<(PackagesUpd
     if !input.allow_dirty {
         repo_is_clean_result?;
     } else if repo_is_clean_result.is_err() {
-        // Commit the uncommitted changes in a temporary commit, so we can use `checkout_head`
-        // to restore them
-        repository
-            .repo
-            .add_all_and_commit("release-plz temp commit with uncommitted changes")?;
+        // Stash uncommitted changes so we can freely check out other commits
+        repository.repo.git(&[
+            "stash",
+            "push",
+            "--include-untracked",
+            "-m",
+            "uncommitted changes stashed by release-plz",
+        ])?;
     }
     let packages_to_update = updater
         .packages_to_update(&registry_packages, &repository.repo, input.local_manifest())
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -423,7 +423,9 @@ pub async fn next_versions(input: &UpdateRequest) -> anyhow::Result<(PackagesUpd
     if !input.allow_dirty {
         repo_is_clean_result?;
     } else if repo_is_clean_result.is_err() {
-        // Stash uncommitted changes so we can freely check out other commits
+        // Stash uncommitted changes so we can freely check out other commits.
+        // This function is ran inside a temporary repository, so this has no
+        // effects on the original repository of the user.
         repository.repo.git(&[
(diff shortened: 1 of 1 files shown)


marcoieni:
Will this commit be present in the release PR?
In don't get why committing is useful

eviltak:
> Will this commit be present in the release PR?

No, since the release-pr creation process creates a new `TempRepo`.

> In don't get why committing is useful

It's to avoid the uncommitted changes being lost/overridden when checking out a different commit in the temp repository. It also helps avoid an `error: Your local changes to the following files would be overwritten by checkout`.

eviltak:
Specifically, it will prevent this from ever happening in `get_package_diff`:

https://github.com/release-plz/release-plz/blob/3255ff4495850c70d5de139e1d9c5f2f3e0c3d85/crates/release_plz_core/src/next_ver.rs#L892-L897

marcoieni:
I see. Have you considered `git stash push --include-untracked -m "name_of_stash"` instead of creating a commit?
It feels like a less destructive operation. I find creating a commit is a bit weird. But maybe I'm missing something. 🤔

You basically don't need these changes in the temp repo, so you just want to "revert" them, right?

eviltak:
> You basically don't need these changes in the temp repo, so you just want to "revert" them, right?

The opposite - I want to make sure that they aren't lost and are properly accounted for in the temp repo.

https://github.com/release-plz/release-plz/blob/644331fe32466e9972bbfc4b8a9de59e5b1c39ad/crates/release_plz/src/args/update.rs#L84-L87

Unless I misunderstood something, wouldn't stashing/discarding the uncommitted changes do the opposite of  "The uncommitted changes will be part of the update"? If the only difference between a released version and the current working tree is an uncommitted change, stashing them here would lead to  `next_versions` not returning any changes for that package (and leaving the version for that package untouched).

> this commit could end up in the changelog

Yes, the commit would end up in the changelog, unless it is skipped by a commit parser. But if the uncommitted changes are part of an update, perhaps they should also be mentioned in the changelog? 

In my opinion, the easiest way to include uncommitted changes in an update is to commit them as such in the temp repo - this allows us to `checkout` to different commits, and ensures that the uncommitted changes are never lost. The only downside is that the commit may appear in the changelog; however, if you don't want it in the changelog, it should be pretty easy to filter it out when the changelog is generated in `Updater::update_result`.

I should also add that the motivation for this change was to allow for checking out the repository at potential release tags in #1497 to perform some validation (e.g. the workspace at that tag has a package with the right name and version). Without this change, the current implementation of #1497 would fail to work for _most_ dirty work trees, since the following checkout is pretty much guaranteed to fail if you modify any file belonging to a package:

https://github.com/eviltak/release-plz/blob/92237ca66e310444e4d6bb0f168ae58625329d43/crates/release_plz_core/src/published_packages/git_tags.rs#L62-L64

marcoieni:
> Unless I misunderstood something, wouldn't stashing/discarding the uncommitted changes do the opposite of "The uncommitted changes will be part of the update"?

Yes, you are right.

> If the only difference between a released version and the current working tree is an uncommitted change, stashing them here would lead to next_versions not returning any changes for that package (and leaving the version for that package untouched).

Yes, you are right. I think this is the intended behavior.
Imo if the user wants their changes to be in the changelog they should commit them with a proper error message.
Otherwise, if we filter out that commit message and that's the only change for that package, what do we write in the changelog for that package?

I imagine people use `allow_dirty` for weird use cases, not for updating the rust code of their crates. 🤔

> In my opinion, the easiest way to include uncommitted changes in an update is to commit them as such in the temp repo

Right now uncommitted changes appear in the `relase-plz update` output and in the `release-plz release-pr` commit (tested [here](https://github.com/marcoieni/rust-workspace-example/pull/233/files)).
So technically they are "part of the update" 😅

So maybe stashing is the right solution to your problem because it keeps the current behavior, right? (i.e. the files will still be in the release pr, but not show in changelog, etc)

eviltak:
I understand now, thanks for the clarification! From what you said, I think stashing will achieve the intended effect. It may cause some unforeseen changes to the `allow_dirty` behavior, but as you said, I doubt people are using the option for anything other than one-off operations. 

My main goal is to be able to freely check out any commit in the temp repo, so stashing is perfectly fine. I'll change it to run the `git stash push --include-untracked -m "name_of_stash"` command you recommended.

crates/release_plz_core/src/next_ver.rs
```diff
@@ -418,8 +418,19 @@ pub async fn next_versions(input: &UpdateRequest) -> anyhow::Result<(PackagesUpd
     let repository = local_project
         .get_repo()
         .context("failed to determine local project repository")?;
+
+    let repo_is_clean_result = repository.repo.is_clean();
     if !input.allow_dirty {
-        repository.repo.is_clean()?;
+        repo_is_clean_result?;
+    } else if repo_is_clean_result.is_err() {
+        // Stash uncommitted changes so we can freely check out other commits
```
marcoieni, line 426:
```suggestion
        // Stash uncommitted changes so we can freely check out other commits.
        // This function is ran inside a temporary repository, so this has no
        // effects on the original repository of the user.
```

marcoieni (approved):

PR #1983, opened 2025-01-12
fix(core): Make `release --dry-run` more informative

eviltak:
When `--dry-run` is specified, prints (by adding `tracing` events) the steps that `release` would have taken, such as git tag and release creation.


marcoieni:
I changed the code so that it emits one info line. Hopefully it's less verbose. What do you think?

eviltak:
Works for me!

marcoieni (approved):

Issue #524, opened 2023-02-08
`release-plz update` and `release-plz release-pr` ignores yanked versions


MarcoIeni:

MarcoIeni:
blocked by https://github.com/rust-lang/cargo/issues/11693

MarcoIeni:
Maybe reading the cargo audit codebase might help

MarcoIeni:
Also, maybe calling the registry api directly might help
https://doc.rust-lang.org/nightly/cargo/reference/registry-index.html

marcoieni:
Idea from https://github.com/release-plz/release-plz/pull/1984#issuecomment-2598563145:

> It could also be possible to lookup the latest (yanked or not) published version of a crate via the `crates_index` library and then specifically request that precise version (e.g. `= 1.2.3`) from the registry source.

hunger:
Updated doc link: https://release-plz.dev/docs/extra/yanked-packages

PR #1984, opened 2025-01-12
allow fetching latest package summaries from cargo registry

eviltak:
Refactors the `clone::Cloner` implementation to allow for retrieving the intermediary `IndexSummary`s when looking up the latest version of a package in a Cargo registry. 

This allows us to query published package metadata (such as name and version) from the `IndexSummary` without having to download the entire package, thereby letting us defer the package download.

### DIFF
--- a/crates/release_plz_core/src/clone/mod.rs
+++ b/crates/release_plz_core/src/clone/mod.rs
@@ -22,7 +22,7 @@ use anyhow::{bail, Context};
 use cargo::core::dependency::Dependency;
 use cargo::core::Package;
 use cargo::sources::source::{QueryKind, Source};
-use cargo::sources::{PathSource, SourceConfigMap};
+use cargo::sources::{IndexSummary, PathSource, SourceConfigMap};
 
 use walkdir::WalkDir;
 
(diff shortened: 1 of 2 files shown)


marcoieni:
> This allows us to query published package metadata (such as name and version) without having to download the entire package

Would [this](https://github.com/eviltak/release-plz/blob/d8743ddf6e599d2c71e8fc0b8ff1ebcffee07faa/crates/release_plz_core/src/cargo.rs#L59) do the same for your purposes or do we need this new method?
Indexes are retrieved [here](https://github.com/eviltak/release-plz/blob/2602b119ef46c6187155de72f01bb6601105c217/crates/release_plz_core/src/command/release.rs#L567).

eviltak:
> Would [this](https://github.com/eviltak/release-plz/blob/d8743ddf6e599d2c71e8fc0b8ff1ebcffee07faa/crates/release_plz_core/src/cargo.rs#L59) do the same for your purposes or do we need this new method?

That won't work, because I need the `IndexSummary` so I can pass it to `clone_from_summary_into`. It uses the `PackageId` stored in the `IndexSummary` to download the package, but with future versions of Cargo could need other information. 

> Maybe your solution is better and could solve #524 ?

Unfortunately, Cargo [filters out](https://github.com/rust-lang/cargo/blob/0.85.0/src/cargo/sources/registry/mod.rs#L816-L823) yanked packages (unless they are specifically requested or in the Cargo.lock) before passing them to the callback passed to `Source::query`, so this won't _directly_ fix #524.

However, this two-step (query first, download second) approach enables the the latest released package version algorithm (as already discussed in #1497). As you alluded to in that PR, if a released package version is yanked from the registry, it can still be retrieved from other sources (e.g. git tags). The latest released package version selection algorithm would then choose the yanked package from the other source (e.g. it's git tag), rather than choosing the one in the registry.

It could also be possible to lookup the latest (yanked or not) published version of a crate via the `crates_index` library and then specifically request that precise version (e.g. `= 1.2.3`) from the registry source.

crates/release_plz_core/src/clone/mod.rs
```diff
@@ -140,26 +193,19 @@ impl Cloner {
...
-                    clone_directory(to_utf8_path(pkg.root())?, dest_path)
-                        .context("failed to clone directory")?;
-                }
+        let name = &crate_.name;
+        let vers = crate_.version.as_deref();
+        let latest = query_latest_package_summary(src, name, vers)?;
+
+        let pkg = match latest {
+            Some(l) => {
+                let pkg = self.clone_from_summary_into(&l, dest_path)?;
```
marcoieni, line 202: this function acquires the lock. The lock was already acquired becase this function is called by `clone_single`, which is called by `clone_in`, which is called by `clone` (which acquires the lock)

eviltak: No, this is not a deadlock. The [Cargo cache locks](https://docs.rs/cargo/latest/cargo/util/cache_lock/index.html) provide per-process mutual exclusion, and also allow recursively acquiring locks within a process.

marcoieni: in the next [line](https://github.com/eviltak/release-plz/blob/7aa7059bde44a7396a30012bc8b9e23f87041201/crates/release_plz_core/src/clone/mod.rs#L88) we also call get_source, which invalidates the cache 😵
We already invalidate the cache in the `clone` function.

Why is the function `clone_from_summary_into` public? Can't the function `clone_from_summary_into` be a private functions like the others? You could take source as an argument and avoid acquiring the lock 👍
If the function needs to be public later, we can address this in another PR

I also tried to ensure that the lock is taken via the type system but I failed 😅

eviltak: The `clone_from_summary_into` function needs to be public for the released packages refactor required by git tags feature, and will be needed for changes introduced in a PR I am going to make right after this one is merged. 

After that change, the `clone` function will only be called in `PackageDownloader::download`, which itself would have been made redundant (its usages would have been replaced by calls to `Cloner::query_latest_package` and `Cloner::clone_from_summary_into`) and would only ever be called in the `TestContext::download_package` release-plz integration test utility. 

Though it would only be used in tests, I chose to leave the `PackageDownloader` struct untouched and do this change in two parts to reduce the number of changes required and to be able to split it up into smaller PRs. 

If you feel strongly about it I can move this change along with the `PackageDownloader` and `Cloner::clone` deletion to the "Refactor released package retrieval" PR I was planning on creating after this one, and close this PR.

marcoieni: I did this change in https://github.com/release-plz/release-plz/pull/1984/commits/bb843227498e9c1ce405cf7fe7f7bc3f11255d47
We can change it in another PR 👍

marcoieni (approved):
thanks!

marcoieni:
I don't think this is a `fix`, right?
I removed it from the PR title 👍

Issue #1996, opened 2025-01-18
Wrong next version when there is a whitespace in the scope


orhun:
## Bug description

Originally reported at https://github.com/orhun/git-cliff/issues/1008

`next_version` crate does not work as expected when there is a whitespace in the scope.

- Would you like to work on a fix? inshallah

## To Reproduce

Reproducible test case added in #1997

## Expected behavior

Correct version bump

## Environment

- OS: arch btw
- release-plz version: master

## Additional context

None, thank you!

orhun:
It looks like we need to go one layer deeper: https://github.com/cocogitto/conventional_commits_parser_rs/issues/4

marcoieni:
the fix is https://github.com/release-plz/release-plz/issues/2002 👍

PR #1997, opened 2025-01-18
test(next-version): add test for scope with whitespace

orhun:
See #1996


lgatellier:
@orhun I think you can also add a test with minor a non-breaking change, since the bug also occurs in this case.

crates/next_version/tests/all/normal.rs
```diff
@@ -94,3 +94,23 @@ fn commit_with_custom_minor_increment_regex_increments_minor_version() {
         Version::new(1, 3, 0)
     );
 }
+#[test]
```
marcoieni, line 97:
```suggestion

#[test]
```

marcoieni (approved):

PR #2001, opened 2025-01-20
feat: Release unpublishable packages

eviltak:
`release-plz` currently does not process packages with `package.publish = false` in their Cargo.toml (i.e. unpublishable packages) at all. This PR allows such packages to be processed and released (to places other than Cargo registries) by release-plz.

However, to avoid [breaking workflows](https://github.com/release-plz/release-plz/pull/1872#issuecomment-2577202196) for existing workspaces with unpublishable packages, this PR also introduces a config change to only enable processing of nonpunishable packages if it is explicitly enabled in the release-plz config. In other words, the processing of unpublishable packages must be made opt-in.

This is done by setting `release = true` in the package-specific config for an unpublishable package, _even if_ `release = true` is set for the workspace. In the future, we could [consider adding an option](https://github.com/release-plz/release-plz/pull/1872#issuecomment-2582031148) like `release_unpublishable` to the `[workspace]` config to opt into releasing _all_ unpublishable packages in the workspace.

## TODO
- [ ] Update website documentation to reflect new behavior

### DIFF
--- a/crates/release_plz/src/config.rs
+++ b/crates/release_plz/src/config.rs
@@ -1,12 +1,15 @@
 use cargo_metadata::camino::Utf8Path;
 use cargo_utils::to_utf8_pathbuf;
 use release_plz_core::{
-    fs_utils::to_utf8_path, set_version::SetVersionRequest, GitReleaseConfig, ReleaseRequest,
-    UpdateRequest,
+    fs_utils::to_utf8_path, set_version::SetVersionRequest, GitReleaseConfig, Publishable,
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, path::PathBuf, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+    time::Duration,
+};
 use url::Url;
 
 use crate::changelog_config::ChangelogCfg;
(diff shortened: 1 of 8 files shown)


website/docs/faq.md
```diff
@@ -1,20 +1,29 @@
...
 
-If you want to check which packages release-plz will publish, run
+If you want to check which packages release-plz will release, run
 `release-plz release --dry-run`.
 
+### Making release-plz process unpublished packages
+
+Unpublished packages *can* be updated and released by release-plz if so desired.
+Their processing must be opted into by setting the [package-specific `release` option](config.md#the-release-field-package-section)
+to `true`, even if `release = true` is specified at the workspace level.
```
marcoieni, line 25:
```suggestion
### Release unpublishable packages

Even if a project is unpublishable, you can still use release-plz to update and release it.

To do so, and make release-plz process a package, set the
[package-specific `release` option](config.md#the-release-field-package-section)
to `true`.
Note that you need to set `release = true` at the `[[package]]` level, even if `release = true` is specified at the `[workspace]` level.
```

marcoieni (commented):
comments to the docs. I'll review the code asap.

crates/release_plz/src/config.rs
```diff
@@ -73,37 +54,88 @@ impl Config {
...
+
+    /// Configures `self` with the given package-specific config and overrides for a package with
+    /// the given name. Returns the new configured value.
+    fn configure_package(
+        self,
+        package: impl Into<String>,
+        package_config: PackageSpecificConfig,
+        overrides: &Self::Overrides,
+    ) -> Self;
+}
```
marcoieni, line 77: what's the point of this trait if it doesn't specify any common functions?
Doesn't it make the code only more complex and less flexible? 🤔

Also, I wish this was introduced as a standalone PR as a "refactoring" and then the change of the behavior was done in a PR based on that one.

marcoieni:
this was needed for https://github.com/release-plz/release-plz/issues/1144 I suppose but that feature was implemented in https://github.com/release-plz/release-plz/pull/2512

Issue #2000, opened 2025-01-20
`pr_branch_prefix` does not substitute template variables


sassman:
## Bug description

the configuration `pr_branch_prefix` does not support template variables. e.g. `{{package}}` is not substituted.

- Would you like to work on a fix? [y]

## To Reproduce

1. create a config fire `.release-plz.toml`
2. in the `[workspace]` section add `pr_branch_prefix = "release-plz-{{package}}-"`
3. let the github action trigger, and see that `{{package}}` is still part of the branch name, not the substituted package name

## Expected behavior

well if the package was named `abc` the branch name then should be prefixed `release-plz-abc-`

## Screenshots

<img width="611" alt="Image" src="https://github.com/user-attachments/assets/e9b8207d-05d7-45dc-8442-b46c1dd6538a" />

## Environment

Verified via the github action

- runs-on: ubuntu-latest
- release-plz/action@v0.5

marcoieni:
There's a very similar discussion: https://github.com/release-plz/release-plz/issues/1991

It's not a bug because it's not documented that you can substitute template variables (right?).

My main point for not supporting this is that if you are in a workspace with multiple packages, what should `{{package}}` be?

What if the package is renamed? In that case we need to close the old PR, but release-plz can't do it because it doesn't find the prefix anymore.

So changing the prefix has some complex implications. What do you think?

sassman:
Well would this not be also an issue for the PR name?

But there it is supported:
https://release-plz.dev/docs/config#the-pr_name-field

As a user without the insight knowledge, it is a bit unexpected that same things (PR vs Branch) behave different.

sassman:
> It's not documented that you can substitute template variables

I just realized that I read the docu that way that those placeholders where just examples provided. In my mind it was a context with all variables being available to all config fields. Which was a false assumption. 

Maybe it's worth to indicate on the docs whats the actual context on available variables, and have this distingued from the usage examples of how the variables can be used. 

And if a configuration has no variables and no templating, then this can be also indicated by e.g. showing no context available or similar. 

I know, in cargo-generate could have done this part actually also better.

marcoieni:
sorry for the late reply!

> Well would this not be also an issue for the PR name?

yes, but you can (and we do) easily rename PR names. You can't do the same for branches.

Imagine the following case:
- author edits package A in its workspace
- release-plz releases the release pr.
- author edits package A in its workspace
- the branch name changed if it contained `{package}`. what should release-plz do in this case?

PR #2003, opened 2025-01-21
fix: fix `pr_branch_prefix` accepting same template variables as `pr_branch_name` does

sassman:
- this fixes issue #2000
- tests are yet missing
- doc updates also missing


sassman:
In my context I have a mono repo with several rust crates that are all non workspace projects.

So I only aim for isolation in the branch name, to not have duplicate branch names that overwrite each other.

I guess your suggestion makes sense for workspace projects. So I'm happy with it :)

PR #2005, opened 2025-01-24
feat: Download packages from different registries in parallel to improve performance

LukeMathWalker:
Primarily looking at Pavex here, where downloading packages takes several seconds.

### DIFF
--- a/crates/release_plz_core/src/registry_packages.rs
+++ b/crates/release_plz_core/src/registry_packages.rs
@@ -65,41 +65,8 @@ pub fn get_registry_packages(
             let temp_dir = tempdir().context("failed to get a temporary directory")?;
             let directory = temp_dir.as_ref().to_str().context("invalid tempdir path")?;
 
-            // Find the registry from where to download each package.
-            let packages_grouped_by_registry = local_packages.iter().chunk_by(|p| {
-                // If registry is not provided, fallback to the Cargo.toml `publish` field.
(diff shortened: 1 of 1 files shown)


marcoieni:
Hey Luca 👋

I'm testing this on pavex. It seems like only one spawn is created.

Try adding a print like this:

```rust
            let handle = scope.spawn(move || {
                println!("downloading ");
                downloader.download()
            });
```

marcoieni:
I think this is what we need to parallelize
https://github.com/release-plz/release-plz/blob/2c3c25e43db3f68fec66cdbf44203c57287f00a0/crates/release_plz_core/src/clone/mod.rs#L129

LukeMathWalker:
You're right, although it might useful to other setups to merge this as is.
I'll open a separate PR for the cloning part.

marcoieni (approved):

LukeMathWalker:
It doesn't look like parallelizing cloning will be possible, since we need to produce a `Package` struct from a different thread, and that's Russian doll of `Rc`s, so it won't compile.

marcoieni:
we discard the Package struct here:
https://github.com/release-plz/release-plz/blob/767491eb3cb4f391dad3b089d29a48e97b6d7ec8/crates/release_plz_core/src/download.rs#L76

So maybe there's a way to download the packages without producing the Package struct?
Worst case we can read the manifest 😅

Issue #1331, opened 2024-03-09
remove `cargo` dependency


MarcoIeni:
## Motivations

cargo increases the compile time of release-plz.
We only use it to download crates from the registries.

## Solution

After reading [this](https://doc.rust-lang.org/cargo/reference/registry-index.html) docs, I think we can implement the download of the crate by ourselves.

Example code doing this (not sure if they work with private registries):
- https://github.com/Nullus157/cargo-dl
- https://github.com/Xion/cargo-download
- https://github.com/nakabonne/rhack

Things to check:
- [ ] make sure it works with private registries: both sparse and git protocol
- [ ] support authentication with cargo login. This is from `cargo help login`:

      For the default cargo:token credential provider, the token is saved in $CARGO_HOME/credentials.toml.
      CARGO_HOME defaults to .cargo in your home directory.

      If a registry has a credential-provider specified, it will be used.
      Otherwise, the providers from the config value registry.global-credential-providers will be attempted, starting from the end of the list.
  

## Alternatives

It would be much better if cargo offered this out of the box. I.e. `cargo download my-crate`.
We could open an RFC to add this functionality to cargo.
However, this might take a lot of time until we see it implemented.

mbodmer:
I am not sure whether you would get all the information you need from `cargo search <CRATENAME> [--registry myregistry]`.

Do we need to actually download the crate?

MarcoIeni:
Yes. Release-plz compares the current project with the published version file by file, going back to history commit by commit.

When the two projects match release-plz stops looking back in git history and that is the last commit that goes into the changelog for the new release.

MarcoIeni:
The best would be to have this command built into cargo, i.e. https://github.com/rust-lang/cargo/issues/1861

PR #2013, opened 2025-02-01
Parallel cloning

LukeMathWalker:
Continuining the conversation from #2005 here, rather than on the closed PR cc @marcoieni 

I can discard `cargo`'s package structure early on, but I then have to fight with `self` which contains `GlobalContext`, which is again not `Send`. The only thing inside `self` we seem to be using is the shell, which is a `RefCell`.

### DIFF
--- a/crates/release_plz_core/src/clone/mod.rs
+++ b/crates/release_plz_core/src/clone/mod.rs
@@ -34,6 +34,7 @@ pub use cargo::{
 
 use crate::fs_utils::strip_prefix;
 use crate::fs_utils::to_utf8_path;
+use crate::read_package;
 
 /// Rust crate.
 #[derive(PartialEq, Eq, Debug)]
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
I was battling with the very same issue in Pavex this week, and I chose to create a global `static` with an `Arc<Mutex<Shell>>` inside to simplify things.
That works in my case since I vendor the parts of `cargo` I need, therefore I don't have to obey its API contract.
An alternative here would be to use a `LazyLock` in a thread local, giving each thread its own `Shell` which you can then fetch whenever needed. There seems to be no [configuration](https://github.com/release-plz/release-plz/blob/263b82e16768501b0cb953c70ca223eadba1435b/crates/release_plz_core/src/clone/cloner_builder.rs#L81) for the shell itself, so we wouldn't need any arguments.

LukeMathWalker:
Never mind, we have:

```rust
let pkg = Box::new(src).download_now(summary.package_id(), &self.config)?;
```

in the call stack, so you need an actual `cargo` global context instance.

marcoieni:
> I was battling with the very same issue in Pavex this week, and I chose to create a global `static` with an `Arc<Mutex<Shell>>` inside to simplify things.
> That works in my case since I vendor the parts of `cargo` I need, therefore I don't have to obey its API contract.

that works for me! It would also solve https://github.com/release-plz/release-plz/issues/1331

marcoieni:
btw to speed up CI we could cache `~/.cargo/registry/cache` 🤔

Issue #2002, opened 2025-01-20
Switch to git-conventional


orhun:
## Motivations

[`conventional_commit_parser`](https://crates.io/crates/conventional_commit_parser) hasn't been updated in 3 years at this point and causes some issues like #1996 whereas [`git-conventional`](https://crates.io/crates/git-conventional) is more up-to-date and battle-tested by [`git-cliff`](https://github.com/orhun/git-cliff/blob/main/git-cliff-core/Cargo.toml#L92).

- Would you like to implement this feature? [y/n] y

## Solution

Switch to `git-conventional` for parsing conventional commits.

## Alternatives

None, idk

## Additional context

See #1996

marcoieni:
Having just one dependency for commit parsing in the dependency tree of release-plz makes sense. PR welcome 👍

PR #2021, opened 2025-02-01
chore: use `git-conventional` library to parse commits

marcoieni:
Close https://github.com/release-plz/release-plz/issues/2002

### DIFF
--- a/crates/next_version/src/version_increment.rs
+++ b/crates/next_version/src/version_increment.rs
@@ -51,7 +51,7 @@ impl VersionIncrement {
             if !current_version.pre.is_empty() {
                 return Some(VersionIncrement::Prerelease);
             }
-            // Parse commits and keep only the ones that follow conventional commits specification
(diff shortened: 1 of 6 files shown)


crates/next_version/src/version_increment.rs
```diff
@@ -12,22 +12,8 @@ pub enum VersionIncrement {
...
-        commits
-            .iter()
-            .any(|commit| custom_commit_matches_regex(regex, commit))
-    } else {
-        false
-    }
-}
-
-fn custom_commit_matches_regex(regex: &Regex, commit: &ConventionalCommit) -> bool {
-    if let CommitType::Custom(custom_type) = &commit.commit_type {
```
marcoieni, line 26: we removed this check, so this is a breaking change because you can now override the default behavior of conventional commits types.
The new behavior is documented in the doc tests of this PR

marcoieni: actually it shouldn't be a breaking change I think.

PR #2033, opened 2025-02-05
fix: support env var override for registry index

### LINKED ISSUE #2032: Release request doesn't handle `CARGO_REGISTRIES_<name>_INDEX` env variable properly
## Bug description

Similar to #1506, the `CARGO_REGISTRIES_<name>_INDEX` env var is not respected for providing an index url for private regististries per the [doc](https://doc.rust-lang.org/cargo/reference/config.html#registries).

## Expected behavior

Call `release-plz release --registry foo` against custom registry while having `CARGO_REGISTRIES_FOO_INDEX` `CARGO_REGISTRIES_FOO_TOKEN` env variable set.

## Environment

- OS: macOS 14.5
- release-plz version: 0.3.116

ncatelli:
## Summary
Attempts to loop up a `CARGO_REGISTRIES_<name>_INDEX` env var as the highest priority override for registry indexes falling back to the cargo config hierarchy if not present.

## Related Issues
closes #2032 

## Todo
- [x] Test coverage

### DIFF
--- a/crates/cargo_utils/src/registry.rs
+++ b/crates/cargo_utils/src/registry.rs
@@ -7,6 +7,14 @@ use url::Url;
 const CRATES_IO_INDEX: &str = "https://github.com/rust-lang/crates.io-index";
 const CRATES_IO_REGISTRY: &str = "crates-io";
 
+/// Read index for a specific registry using environment variables.
+/// <https://doc.rust-lang.org/cargo/reference/environment-variables.html>
+pub fn registry_index_url_from_env(registry: &str) -> Option<String> {
+    let env_var = format!("CARGO_REGISTRIES_{}_INDEX", registry.to_uppercase());
+
+    std::env::var(env_var).ok()
+}
+
 /// Find the URL of a registry
 pub fn registry_url(manifest_path: &Path, registry: Option<&str>) -> anyhow::Result<Url> {
     fn read_config(
@@ -31,6 +39,19 @@ pub fn registry_url(manifest_path: &Path, registry: Option<&str>) -> anyhow::Res
     // it's looks like a singly linked list
     // put relations in this map.
     let mut registries: HashMap<String, Source> = HashMap::new();
+
+    // set top-level env var override if it exists.
+    if let Some(registry_name) = registry {
+        if let Some(env_var_override) = registry.and_then(registry_index_url_from_env) {
+            registries
+                .entry(registry_name.to_string())
+                .or_insert(Source {
+                    registry: Some(env_var_override),
+                    replace_with: None,
+                });
+        }
+    }
+
     // ref: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
     for work_dir in manifest_path
         .parent()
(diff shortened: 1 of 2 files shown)


crates/cargo_utils/src/registry.rs
```diff
@@ -31,6 +39,19 @@ pub fn registry_url(manifest_path: &Path, registry: Option<&str>) -> anyhow::Res
     // it's looks like a singly linked list
     // put relations in this map.
     let mut registries: HashMap<String, Source> = HashMap::new();
+
+    // set top-level env var override if it exists.
+    if let Some(registry_name) = registry {
+        if let Some(env_var_override) = registry.and_then(registry_index_url_from_env) {
```
marcoieni, line 45:
```suggestion
        if let Some(env_var_override) = registry_index_url_from_env(registry_name) {
```

crates/release_plz_core/src/command/release.rs
```diff
@@ -1000,6 +1001,26 @@ mod tests {
...
+        V: AsRef<OsStr>,
+        F: FnOnce(),
+    {
+        // Store the previous value of the var, if defined.
+        let previous_val = env::var(key.as_ref()).ok();
+
+        env::set_var(key.as_ref(), value.as_ref());
+        (f)();
+
+        // Reset or clear the var after the test.
```
marcoieni, line 1016: this doesn't work because each test doesn't run in an isolated process. So this env var will be shared with other tests that run at the same time.

crates/release_plz_core/src/command/release.rs
```diff
@@ -1040,19 +1061,42 @@ mod tests {
...
-        if let Ok(old) = old_value {
-            env::set_var(&token_env_var, old);
-        } else {
-            env::remove_var(&token_env_var);
-        }
+    #[test]
+    fn should_reference_env_var_provided_index() {
+        use cargo_utils::registry_url;
+
+        let _guard = NO_PARALLEL.lock().unwrap();
```
marcoieni, line 1077: should we acquire the lock inside the `with_env_var` function?
I think that's the critical part of the code we can't run in parallel

crates/release_plz_core/src/command/release.rs
```diff
@@ -1040,19 +1061,42 @@ mod tests {
...
+            // assert the registry index is properly overriden
+            assert_eq!(maybe_registry_index, mock_index_url);
+        });
+
+        let non_overriden_maybe_registry_index =
+            registry_url(fake_manifest_path, Some(registry_name)).ok();
 
-        assert!(registry_token.is_some());
-        assert_eq!(token, registry_token.unwrap().expose_secret());
+        // assert it's inherited from the workspace otherwise
```
marcoieni, line 1099: is "otherwise" at the end a typo? 🤔

marcoieni, line 1099:
```suggestion
        // assert it's inherited from the workspace since the environment
        // variable is no longer set at this point.
```

marcoieni:
Thanks, I'm merging this. Let me know if it works for you after testing it 👍

marcoieni (approved):

PR #2059, opened 2025-02-09
style: add style to the command line

joshka:
image

### DIFF
--- a/crates/release_plz/src/args/mod.rs
+++ b/crates/release_plz/src/args/mod.rs
@@ -14,7 +14,7 @@ use anyhow::Context;
 use cargo_metadata::camino::{Utf8Path, Utf8PathBuf};
 use cargo_utils::CARGO_TOML;
 use clap::{
-    builder::{styling::RgbColor, Styles},
+    builder::{styling::AnsiColor, Styles},
     ValueEnum,
 };
 use init::Init;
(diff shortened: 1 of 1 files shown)


marcoieni:
I feel like people might have opinions on the colors we pick, whatever they are 😅
Maybe it could also conflict with their color scheme.

E.g. If someone has the nord color scheme I imagine these colors would be a punch in their face.

Do cargo or other relevant tools in the rust ecosystem have colors for their help menu? 🤔

marcoieni:
I see that cargo has some colors and I guess they are the same regarding of the color scheme 🤔

marcoieni:
> I'm not too concerned about the specific colors.

This is how these colors look on a white theme. Yellow is unreadable.

image

Should we use the same colors cargo is using?

joshka:
I think it's worth going with your own colors a little, but using the color mappings from the user's terminal theme. E.g. the fix above uses the standard 16 bit colors - Red and Yellow, which could be any color. It's up to the terminal theme to ensure that the chosen rendering of yellow works correctly on the terminal background color. Feel free to choose colors that make sense to you here and update this PR. I don't really mind which specific colors are chosen, only that there are colors.

marcoieni (approved):

marcoieni:
thanks!

PR #2070, opened 2025-02-15
feat: edit changelog order to respect release order

marcoieni:


crates/release_plz_core/src/project.rs
```diff
@@ -99,6 +107,7 @@ impl Project {
         &self.root
     }
 
+    /// Packages that can be published. They are already ordered by release order.
```
marcoieni, line 110:
```suggestion
    /// Packages that can be published, ordered by release order.
```

Issue #2060, opened 2025-02-10
Update Clap usage approachs to be more idiomatic rust tooling


joshka:
## Motivations

There's a bunch of subtle little things that mean the way the CLI works is a little different from other Rust tooling (specifically cargo). I'm about to start some work on these, but wanted to give a heads up prior to doing so.

## Solution

I intend to make the following quality of life changes 

- [ ] Color / styling: https://github.com/release-plz/release-plz/pull/2059
- [ ] Display tool name in help text
- [ ] Short and long descriptions of commands
- [ ] Command help text should generally be imperative (e.g. "Release packages to the cargo registry" instead of "For each package not published ..."
- [ ] Consistent wording with cargo where parameters are the same
- [ ] Shorter command aliases (e.g. `pr` instead of `release-pr`, `completions` instead of `generate-completions`)
- [ ] When there are more than 4-6 options, group options into related sections
- [ ] Order commands / options from more common to more esoteric
- [ ] Add dry-run option to all release-pr command (and maybe update)
- [ ] Move common options (manifest / registry) to be included / flattened (to avoid repeating the help text in each)
- [ ] Use ArgAction::SetFalse for `--no-<xxx>` options
- [ ] Use clap related crates to implement common behavior rather than implementing it directly (e.g. [clap-verbosity-flag](https://crates.io/crates/clap-verbosity-flag), [clap_complete_command](https://crates.io/crates/clap_complete_command) etc.

Where the visible command / option name changes, I'll leave aliases so that there's no breaking change for anyone using release-plz in scripts

How does this sound?

marcoieni:
It sounds great, thanks!

joshka:
Starting points for comparison later:

![Image](https://github.com/user-attachments/assets/f5a098dc-a0b6-4782-bb09-118533ea1b10)

![Image](https://github.com/user-attachments/assets/fe293b9c-7f12-46f9-a95c-f0e72a7a6400)

PR #2071, opened 2025-02-16
chore: simplify logging setup

joshka:
Addresses parts of https://github.com/release-plz/release-plz/issues/2060

- Use RELEASE_PLZ_LOG env var to control logging level
- ~Use clap-verbosity-flag to add --verbose and --quiet flags to the cli~
- Fallback to RUST_LOG if RELEASE_PLZ_LOG is not set (for backwards
  compatibility)
- Remove "pretty" styled logs - these make the logs harder to read as
  when viewing them in a github action as the information is split over
  multiple lines
- Use simpler tracing_subscriber config (`fmt()` and `init()`)
- Remove tracing-log dependency (this is enabled by default by `init()`)

# Default
image

# Verbose
image

# Trace (can also be enabled using `-vv` or `--verbose --verbose`)
image

---

As a side note, I'd generally suggest that keeping the target on by default is a good thing to do as knowing where a log message comes from is one of the main things a log output should tell you. I haven't made that change here.

### DIFF
--- a/crates/release_plz/src/args/mod.rs
+++ b/crates/release_plz/src/args/mod.rs
@@ -17,11 +17,10 @@ use clap::{
     ValueEnum,
     builder::{Styles, styling::AnsiColor},
 };
-use clap_verbosity_flag::{InfoLevel, Verbosity};
 use init::Init;
 use release_plz_core::fs_utils::current_directory;
 use set_version::SetVersion;
-use tracing::info;
+use tracing::{info, level_filters::LevelFilter};
 
 use crate::config::Config;
 
@@ -49,9 +48,28 @@ const HELP_STYLES: Styles = Styles::styled()
 pub struct CliArgs {
     #[command(subcommand)]
     pub command: Command,
+    /// Print source location and additional information in logs.
+    /// The default log level is INFO.
+    /// `-v` sets the log level to DEBUG.
+    /// `-vv` sets the log level to TRACE.
+    /// To change the log level, you can also use the `RELEASE_PLZ_LOG` environment variable. E.g. `RELEASE_PLZ_LOG=DEBUG`.
+    #[arg(
+        short,
+        long,
+        global = true,
+        action = clap::ArgAction::Count,
+    )]
+    verbose: u8,
+}
 
-    #[command(flatten, next_help_heading = "Logging Options")]
-    pub verbosity: Verbosity<InfoLevel>,
+impl CliArgs {
+    pub fn verbosity(&self) -> LevelFilter {
+        match self.verbose {
+            0 => LevelFilter::INFO,
+            1 => LevelFilter::DEBUG,
+            _ => LevelFilter::TRACE,
+        }
+    }
 }
 
 #[derive(clap::Subcommand, Debug)]
--- a/crates/release_plz/src/args/mod.rs
+++ b/crates/release_plz/src/args/mod.rs
@@ -39,10 +39,7 @@ const HELP_STYLES: Styles = Styles::styled()
 
 /// Release-plz manages versioning, changelogs, and releases for Rust projects.
 ///
-/// See the Release-plz website for more information <https://release-plz.ieni.dev/>.
-///
-/// The default log level is INFO. To change it, set the `RELEASE_PLZ_LOG` environment variable or
-/// use the `--verbose` or `--quiet` flags.
+/// See the Release-plz website for more information <https://release-plz.dev/>.
 #[derive(clap::Parser, Debug)]
 #[command(version, author, styles = HELP_STYLES)]
 pub struct CliArgs {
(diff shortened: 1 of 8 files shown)


marcoieni:
thanks. I'm a but busy in these days since I'm in London for rust nation, but I'll be able to review your changes in 1 week hopefully.

In the meantime, do you think the name RELEASE_PLZ_LOG is too long and maybe we should use something like LOG_LEVEL?

joshka:
> In the meantime, do you think the name RELEASE_PLZ_LOG is too long and maybe we should use something like LOG_LEVEL?

Not really. Prefixing with the app name is somewhat of a convention for CLI apps. RUST_LOG is the main built-in environment variable name from EnvFilter, but as soon as you graduate from the app being a rust CLI run from `cargo run` to being a CLI that could be installed anywhere as a binary, adding a prefix is helpful. It allows you to make these settings for the tool in your shell profile. Using a generic name does not allow this.

joshka:
Oh, I meant to add the rationale for this. I was looking at release-plz logs in github actions, and because they're not really using a standard format, it makes it difficult to work out where each line of the log comes from. Moving towards this more standard format helps that immensely and makes it easy to narrow down more details of what went wrong when something fails.

marcoieni:
I removed the `clap-verbosity-flag` because:
- I didn't find the `-q` variable useful. Is there a use case where you want to show less logs than INFO? If yes, we can discuss in a separate issue.
- I found `-v` trivial to implement manually, so I don't think a dependency is needed. The manual implementation keeps the code simpler.

crates/release_plz/src/args/mod.rs
```diff
@@ -49,9 +48,28 @@ const HELP_STYLES: Styles = Styles::styled()
 pub struct CliArgs {
     #[command(subcommand)]
     pub command: Command,
+    /// Print source location and additional information in logs.
+    /// The default log level is INFO.
```
joshka: In order to use the first line as the short help text (`-h`), and reserve the rest for the long text (`--help`) add a blank line:

```suggestion
    /// Print source location and additional information in logs.
    ///
    /// The default log level is INFO.
```

joshka (commented):
LGTM

joshka:
Any thoughts on this comment

> As a side note, I'd generally suggest that keeping the target on by default is a good thing to do as knowing where a log message comes from is one of the main things a log output should tell you. I haven't made that change here.

I didn't want to make that change without running it by you, but I think it's likely good for making it easier to track down what's happening a bit easier. The log message is good for telling what happened, the target is good for telling which part of the app was responsible for making that happen. E.g. a small example - if you're dealing with a message about downloading a crate, the context of a message coming from the part of the app dealing with controlling the download processes, and from the part which does the actual http requests carries different meanings when interpreting what's happening.

marcoieni:
I would prefer to not have the target and file in the logs since these are implementation details that users should not care about and make logs more verbose. 🤔

> if you're dealing with a message about downloading a crate, the context of a message coming from the part of the app dealing with controlling the download processes, and from the part which does the actual http requests carries different meanings when interpreting what's happening.

In this case, I would like the logs to be more explanatory instead of relying on filename or target

joshka:
> I would prefer to not have the target and file in the logs since these are implementation details that users should not care about and make logs more verbose. 🤔
> 
> > if you're dealing with a message about downloading a crate, the context of a message coming from the part of the app dealing with controlling the download processes, and from the part which does the actual http requests carries different meanings when interpreting what's happening.
> 
> In this case, I would like the logs to be more explanatory instead of relying on filename or target

Let's phrase this a bit more explicitly. I recently saw a failure occur running a release-plz github action. The lack of targets and consistent log messages made it more difficult for me to diagnose which crate each log message came from and to diagnose the problem. I think keeping the target generally doesn't harm the readability of things while removing it by default harms the ability to diagnose failures. The target is shown greyed out at the beginning of the line, which helps make it avoid taking the focus from the main message part. (I forget which repo it was exactly - this was some time back now)

marcoieni:
I opened https://github.com/release-plz/release-plz/issues/2088 👍

marcoieni:
I have a question about this pr. How do I enable files, line numbers and target with log level INFO? 🤔

I expected `RELEASE_PLZ_LOG=INFO cargo run -- release --dry-run -v` to do this

joshka:
> I have a question about this pr. How do I enable files, line numbers and target with log level INFO? 🤔
> 
> I expected `RELEASE_PLZ_LOG=INFO cargo run -- release --dry-run -v` to do this

I would rarely if ever want or need the file and line number in a log message. If modules, spans etc are designed well enough, then the target communicates the right about of information. This PR moves file and line number functionality to the debug level.

marcoieni (approved):
Thanks. Nice improvement!
I made verbosity and log level independent so that users have more control. Feel free to open an issue if you think something I changed needs further discussion 👍

PR #2072, opened 2025-02-16
refactor: simplify command modules

joshka:
Fixes some parts of https://github.com/release-plz/release-plz/issues/2060

- rename the `args` module to `commands`
- Move all the each sub commands into command/xxx.rs
- rename modules to simpler names (completions, schema, pull_request)
- rename init to init_github as there may be an eventual init command
  that is just for release-plz.yml
- push code down into modules by adding run() methods on each command
- move command help text into each command struct
- add aliases for backward compatibility of command names
- reorder commands based on a natural ordering which they'd be likely
  encountered (init first, admin-ish commands last)
- fix up help text generally
- Move CliArgs to main

# Main help
image

# Init
image

# Update
image


crates/release_plz/src/commands.rs
```diff
@@ -0,0 +1,69 @@
...
+            Command::GenerateSchema => schema::generate_schema_to_disk()?,
+            Command::InitGithub(command) => command.run()?,
+            Command::SetVersion(command) => command.run()?,
+        }
+        Ok(())
+    }
+}
+
+#[derive(ValueEnum, Clone, Copy, Debug, Eq, PartialEq)]
+pub enum OutputType {
```
joshka, line 58: I'm not 100% sure of why this exists - guessing it's a future addition where this might be available on the commands to allow the user to choose the output type.

crates/release_plz/src/commands.rs
```diff
@@ -0,0 +1,69 @@
...
+pub enum Command {
+    #[command(alias = "init", long_about)]
+    InitGithub(InitGithub),
+    Update(Update),
+    #[command(name = "pr", aliases = ["pull-request", "release-pr"])]
+    PullRequest(PullRequest),
+    SetVersion(SetVersion),
+    Release(Release),
+    #[command(alias = "generate-completions")]
+    Completions(Completions),
```
joshka, line 31: `completions` is a convention used in a number of commands for this functionality. alias for backwards compat

crates/release_plz/src/commands.rs
```diff
@@ -0,0 +1,69 @@
...
+    set_version::SetVersion, update::Update,
+};
+
+#[derive(clap::Subcommand, Debug)]
+pub enum Command {
+    #[command(alias = "init", long_about)]
+    InitGithub(InitGithub),
+    Update(Update),
+    #[command(name = "pr", aliases = ["pull-request", "release-pr"])]
+    PullRequest(PullRequest),
```
joshka, line 27: Let's type short commands when they make sense like this. release-pr alias for backwards compat, pull-request to avoid abbreviations.

crates/release_plz/src/commands.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl Command {
+    pub async fn run(self) -> anyhow::Result<()> {
+        match self {
+            Command::Update(command) => command.run().await?,
+            Command::PullRequest(command) => command.run().await?,
+            Command::Release(command) => command.run().await?,
+            Command::Completions(command) => command.run(),
+            Command::CheckUpdates => check_updates::check_update().await?,
+            Command::GenerateSchema => schema::generate_schema_to_disk()?,
```
joshka, line 49: These could gain command structs if needed, but it's not necessary right now.

crates/release_plz/src/commands/completions.rs
```diff
@@ -1,20 +1,25 @@
...
+pub struct Completions {
     /// Shell option
     #[arg(default_value = "bash")]
     shell: Shell,
 }
 
-impl GenerateCompletions {
-    pub fn print(&self) {
+impl Completions {
+    pub fn run(&self) {
```
joshka, line 17: Renamed for consistency with the other commands, not a big deal. It could be left as-is.

crates/release_plz/src/commands/init_github.rs
```diff
@@ -1,4 +1,36 @@
...
+use clap::builder::PathBufValueParser;
+
+use super::manifest::ManifestCommand;
+
+/// Initialize release-plz for the current GitHub repository.
+///
+/// Stores the necessary tokens in the GitHub repository secrets and generates the release-plz.yml
+/// GitHub Actions workflow file.
+#[derive(clap::Parser, Debug)]
+pub struct InitGithub {
```
joshka, line 12: Renamed from Init to prepare a possible init command that just initializes the release-plz.yml file instead of the entire github repo. I could go either way about whether this is necessary. (No bug problem changing this back to init and perhaps adding sub fields that specify whether to initialize the repo secrets and actions.

crates/release_plz/src/commands/init_github.rs
```diff
@@ -223,6 +255,68 @@ impl ReleaseMetadataBuilder for NoopReleaseMetadataBuilder {
     }
 }
 
+mod gh {
```
joshka, line 258: Copied this in from the previous module as I didn't want to create a whole init folder and gh.rs in there for this much code. Not 100% wedded to the idea here. Suggest something more preferable if you want it.

crates/release_plz/src/main.rs
```diff
@@ -1,110 +1,38 @@
...
+const SECONDARY_COLOR: AnsiColor = AnsiColor::Yellow;
+const HELP_STYLES: Styles = Styles::styled()
+    .header(MAIN_COLOR.on_default().bold())
+    .usage(MAIN_COLOR.on_default().bold())
+    .placeholder(SECONDARY_COLOR.on_default())
+    .literal(SECONDARY_COLOR.on_default());
+
+#[derive(clap::Parser, Debug)]
+#[command( version, author, styles = HELP_STYLES)]
+pub struct CliArgs {
```
joshka, line 23: Move this into main.rs from args as it's now fairly small and it helps give context to the main function which is also rather anemic now that the command parsing / matching has been pushed down. It could still live in an args.rs perhaps, but it makes a bit of sense from a reading code perspective to have this here. (guiding principal, keep small related things together)

joshka:
This will likely conflict with #2071 happy for either to go first and will update the second to work with the other one.

crates/release_plz/Cargo.toml
```diff
@@ -28,7 +28,7 @@ anyhow.workspace = true
 base64.workspace = true
 cargo_metadata.workspace = true
 chrono = { workspace = true, features = ["clock"] }
-clap = { workspace = true, features = ["derive", "env"] }
+clap = { workspace = true, features = ["derive", "env", "wrap_help"] }
```
marcoieni, line 31: I don't see any difference in the output when enabling this

crates/release_plz/src/commands.rs
marcoieni: yes, we might have other output types in the future

crates/release_plz/src/commands/init_github.rs
marcoieni: I prefer to have a single command for both and then have sub fields 👍

crates/release_plz/src/commands/init_github.rs
marcoieni: I prefer files instead of `mod` 👍

crates/release_plz/src/main.rs
marcoieni: I would prefer to have this in its own file 👍

crates/release_plz/src/commands.rs
marcoieni: adding `pr` is fine, I would avoid pull-request to not have too many aliases which can be confusing

marcoieni:
Closing for inactivity. If you want to work on this in the future, please open a small PR for each change.

PR #2084, opened 2025-02-22
feat: support rust 2024 edition

marcoieni:


brianheineman:
@marcoieni Thanks for the fix and for all the effort you put into `release-plz`; much appreciated!!! 🎉

PR #2105, opened 2025-03-02
fix: use tracing for crate download messages

joshka:
Before (In a release-plz workflow):
image

After:
image

I know these are two different commands, but I'm unsure how to test release-plz changes easily.

I suspect there's some weird interplay with the way that the cargo commands work with the shell and the use of the shell by tracing, as there's some text which gets written while downloading and then overwritten by the trace message without clearing the line.

### DIFF
--- a/crates/release_plz_core/src/clone/mod.rs
+++ b/crates/release_plz_core/src/clone/mod.rs
@@ -11,7 +11,7 @@ use cargo_metadata::camino::Utf8Path;
 use cargo_metadata::camino::Utf8PathBuf;
 pub use cloner_builder::*;
 pub use source::*;
-use tracing::warn;
+use tracing::{debug, info, warn};
 
 use std::collections::HashSet;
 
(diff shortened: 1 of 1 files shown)


marcoieni:
> I know these are two different commands, but I'm unsure how to test release-plz changes easily.

There are integration tests or you can use `release-plz update` locally.

> I suspect there's some weird interplay with the way that the cargo commands work with the shell and the use of the shell by tracing, as there's some text which gets written while downloading and then overwritten by the trace message without clearing the line.

Yeah, not sure, I don't want to merge this as it is because of the bug in the words.

Imo we could delete the `Cloning into` log because it's not helpful.

joshka:
You only see those if the source isn't already downloaded.

marcoieni:
I see. I think cargo does already a good jobs at printing logs for crate downloads. should we just delete those 2 notes? 🤔

marcoieni:
deleting these logs in https://github.com/release-plz/release-plz/pull/2129

PR #2117, opened 2025-03-09
!fix: require repo ID in GitHub Action YAML

### LINKED ISSUE #2116: prevent forks from trying to release
## Bug description

when someone forks a project using release-plz github action, by default it will try to run release-plz in the fork repo and create a release PR / possibly try to release your crate.

## Expected behavior

it shouldn't only create make release PRs / releases in the source repo

## Additional context

it would be great if this could be prevented by default by e.g. adding required settings to the github action for the github user/repo. that way it can check if it matches in the fork.

i can't find any documentation on the current site / README saying we should add this conditional logic to the actions workflow.

andrewbanchich:
fixes #2116 

note: need to make PR into the action repo as well

### DIFF
--- a/crates/release_plz/src/init/mod.rs
+++ b/crates/release_plz/src/init/mod.rs
@@ -2,7 +2,7 @@ mod gh;
 
 use std::io::Write;
 
-use anyhow::Context;
+use anyhow::{Context, anyhow};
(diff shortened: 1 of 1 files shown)


marcoieni:
Thanks for this. I chose to adopt the approach of https://github.com/release-plz/release-plz/pull/2128 instead

PR #2153, opened 2025-03-27
fix: copy git repo in terms of a local git clone

0ax1:
All filters need to be turned off when copying the repository to a temporary directory, in order to guarantee that the repository is unchanged and therefore recognized as such by Git.

### DIFF
--- a/crates/release_plz_core/src/copy_dir.rs
+++ b/crates/release_plz_core/src/copy_dir.rs
@@ -3,6 +3,7 @@ use std::{io, path::Path};
 use anyhow::Context;
 use cargo_metadata::camino::{Utf8Path, Utf8PathBuf};
 use tracing::{debug, trace};
+use walkdir::WalkDir;
 
 use crate::fs_utils::strip_prefix;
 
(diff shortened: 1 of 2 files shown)


marcoieni:
this also disables local `.gitignore`.
which means copying the `target/` directory which can be large.
Plus, at that point, why using the `ignore` library altogether? 😂

I'm fine adding standard filters such as `parents()` for example, but why copying files that are ignored by git?
I don't think that they avoid copying this files makes it impossible to recognize the repository by git.

0ax1:
> this also disables local `.gitignore`. which means copying the `target/` directory which can be large. Plus, at that point, why using the `ignore` library altogether? 😂
> 
> I'm fine adding standard filters such as `parents()` for example, but why copying files that are ignored by git? I don't think that they avoid copying this files makes it impossible to recognize the repository by git.

The problem we see right now is that we have a dep outside of our org as a git submodule which force added files that are also part of their ignore patterns. So the files are committed to the git repo, but also listed in `.gitignore`. Therefore, copying a dir and respecting the repo local `.gitignore` is incorrect. 

I agree with your point though that this is not the common case.

What do you think about adding an env var the user can opt into, in order to disable all filters?

0ax1:
A different way to approach this might be along the lines of `git ls-files --recurse-submodules`.

marcoieni:
> A different way to approach this might be along the lines of `git ls-files --recurse-submodules` + `.git` dir. So rather than excluding files, copy all files which are part of the repo.

that works for me.
when you do the pr, also remove the `ignore` dependency, which is no longer needed.
Also add a comment that explains the submodule use case you described, thanks! 🙏

0ax1:
> > A different way to approach this might be along the lines of `git ls-files --recurse-submodules` + `.git` dir. So rather than excluding files, copy all files which are part of the repo.
> 
> that works for me. when you do the pr, also remove the `ignore` dependency, which is no longer needed. Also add a comment that explains the submodule use case you described, thanks! 🙏

I implemented the copy to a temporary dir in terms of a git clone between 2 local paths. This also made the `copy_dir` obsolete. Currently checking why the test suite not fully passes.

0ax1:
@marcoieni Do you think calling git clone is a sensible strategy? Do you have a hunch why 2 of the test suites don't pass, which is not obvious from the CI output.

crates/release_plz_core/src/copy_dir.rs
```diff
@@ -1,173 +0,0 @@
...
-}
-
-#[cfg(test)]
-mod tests {
-    use crate::fs_utils::Utf8TempDir;
-
-    use super::*;
-
-    #[test]
-    fn is_dir_copied_correctly() {
```
marcoieni, line 116: let's bring back this beautiful tests please 😁
You can adapt them to test your copy that uses `git clone`.

crates/release_plz_core/src/git/github_graphql.rs
```diff
@@ -213,14 +208,6 @@ mod tests {
         // file removal
         fs_err::tokio::remove_file(&removed_path).await.unwrap();
 
-        // type change (replace symlink with a content it pointed to)
-        fs_err::tokio::remove_file(&type_changed_path)
-            .await
-            .unwrap();
-        fs_err::tokio::write(&type_changed_path, b"unchanged")
-            .await
-            .unwrap();
```
marcoieni, line 222: why you changed the test? I expect the new implementation to be compatible with the previous one.

marcoieni:
> Do you think calling git clone is a sensible strategy?

yes, IF previous tests cases keep working.
Also it allows to remove a lot of code 👍

> Do you have a hunch why 2 of the test suites don't pass

in the windows test, I see: 
```
thread 'update::up_to_date_project_is_not_touched' panicked at crates\release_plz_core\tests\all\update.rs:13:5:
assertion failed: comparison_test.are_projects_equal()
```

You can look into the `are_projects_equal` function and check why those projects are not equal anymore.
I guess the new copy isn't working as before

For the Test Suite with docker, I see:

```
fatal: empty ident name (for <runner@fv-az1279-218.yw3jarirvqvunhs3l0rvj1dcqe.phxx.internal.cloudapp.net>) not allowed
```

so maybe the git clone requires more `git config` stuff with respect to the other git commands that the test suite runs. Not sure what.

crates/release_plz_core/src/copy_dir.rs
0ax1: Will do!

0ax1:
Gonna close this for now until I find a solution which gels with the current test suite

PR #2156, opened 2025-03-30
fix: canonicalize paths when comparing package contents

aschey:
I'm using a symlinked `README.md` file in a monorepo with multiple crates and I noticed that release-plz wasn't picking up the changes to the readme file. I also use the `include` setting to only include files necessary for my crates to build, so the source readme file isn't tracked in Cargo.

The fix I made is to canonicalize the file paths when comparing the readme files and the package contents, so symlinked files are accounted for correctly when the source files aren't included.


crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -809,14 +809,20 @@ fn get_package_files(
...
             if is_crate_path_same_as_git_repo {
-                l
+                let prefix = repository.directory();
+                let stripped = normalized
+                    .strip_prefix(prefix)
+                    .with_context(|| format!("failed to strip {prefix} from {normalized}"))?;
+                Ok(stripped.to_path_buf())
             } else {
-                crate_relative_path.join(l)
+                Ok(normalized)
```
marcoieni, line 821: the comment of the function says "The paths are relative to the git repo root". If you see how this function is used you get why: [here](https://github.com/release-plz/release-plz/blob/69805e71822d126803f95245346e2a8da5b859a7/crates/release_plz_core/src/command/update/updater.rs#L765) we compare this set of files with the git output.

Here you are returning the normalized output, so this won't work, right? Or am I missing something?

btw thanks for sending this PR!

aschey: Ah yes, good catch. I think we'll have to strip the prefix on both branches here instead of just the first one. That should be doable.

aschey: Pushed a fix for this. Since we're recreating the absolute (canonicalized) path for each item in the loop, we have to strip the prefix each time and the check for `crate_relative_path == ""` isn't useful anymore.

crates/release_plz/tests/all/release_pr.rs
```diff
@@ -534,6 +534,69 @@ async fn release_plz_honors_features_always_increment_minor_flag() {
...
+#[cfg_attr(not(feature = "docker-tests"), ignore)]
+async fn release_plz_detects_symlink_package_changes() {
+    let context = TestContext::new().await;
+    let readme = "README.md";
+    let readme_symlink = "README_SYMLINK.md";
+    symlink_readme(&context, readme, readme_symlink);
+    let cargo_toml_path = context.repo_dir().join(CARGO_TOML);
+    let mut cargo_toml = LocalManifest::try_new(&cargo_toml_path).unwrap();
+
+    // Exclude irrelevant files from the package to ensure changes are being tracked explicitly
```
marcoieni, line 548: last comment before this pr can be merged. I don't get why we only include `src`. how is this relevant for this test? 

does the test fails if we remove this `include`? If not, why having it?

aschey: Without explicitly setting `include` here, the original README will be included in the package since all files are included by default. This means that this test still succeeds without the changes in this PR. We have to include the symlinked version only in order for this test to verify that the path normalization is working correctly. I can try to say this more succinctly and update the comment.

marcoieni (approved):
thanks!

PR #2166, opened 2025-04-04
fix `last` clippy lint and ignore security vulnerabilities

marcoieni:


marcoieni:
I don't know why `cargo update -p openssl` doesn't work, so I will ignore the audit. PR welcome.

PR #2173, opened 2025-04-08
feat: default "remove_source_branch=true" when creating MR in GitLab

mbodmer:
This change changes the default of the checkbox "Delete source branch" to be set, in Gitlab Merge Requests:

image

see: https://docs.gitlab.com/api/merge_requests/#create-mr


crates/release_plz_core/src/git/backend.rs
```diff
@@ -193,6 +195,9 @@ impl From<GitPr> for GitLabMr {
             title: value.title,
             description: desc,
             labels,
+            // default the flag indicating to remove the source branch when
+            // merging to true. Can be manually overridden before merging.
```
marcoieni, line 199:
```suggestion
            // By default, remove the source branch when merging the PR.
            // The checkbox can be unchecked in the UI before merging.
```

crates/release_plz_core/src/git/backend.rs
```diff
@@ -170,6 +170,7 @@ pub struct GitLabMr {
     pub title: String,
     pub description: String,
     pub labels: Vec<String>,
+    pub remove_source_branch: bool,
```
marcoieni, line 173: this struct is also Deserialize. 

we use the /merge_requests endpoint here:

https://github.com/release-plz/release-plz/blob/fd0a01b9a84a5ac849799636c51b139762113e6b/crates/release_plz_core/src/git/backend.rs#L385 

and we deserialize this struct here:

https://github.com/release-plz/release-plz/blob/fd0a01b9a84a5ac849799636c51b139762113e6b/crates/release_plz_core/src/git/backend.rs#L477

I don't think this field is returned. So imo instead of adding this field to this struct, we should add this field only when we create the gitlab pr

mbodmer:
Hi Marco, thanks for reviewing! This is embarrassing, sorry. I have moved the fix to the correct struct now. Was too much of a side thing :-)

marcoieni:
> Hi Marco, thanks for reviewing! This is embarrassing, sorry. I have moved the fix to the correct struct now. Was too much of a side thing :-)

No worries, thanks for this amazing change 😁

Issue #1896, opened 2024-12-08
Fail early if `publish = false` in Cargo.toml and `publish = true` in release-plz configuration


marcoieni:
In the `run` function, add a validation to fail if the [publish](https://docs.rs/cargo_metadata/0.19.1/cargo_metadata/struct.Package.html#structfield.publish) is `Some([])` (which means in Cargo.toml it was `publish = false`) and [publish](https://release-plz.dev/docs/config#the-publish-field) in release-plz config is explicitly set to `true`.

yinkaolotin:
@marcoieni i got into rust a few weeks and looking for rust projects to contribute to. I am interested in contributing to release-plz. I have been going though the codebase and I think this issue might be a good starting point. Could you please confirm if it's available for me to work on?

marcoieni:
Yes, it is, thanks!

PR #2176, opened 2025-04-10
fix: fail early if publish = false in Cargo.toml and publish = true in release-plz config

yinkaolotin:
This changes ensures that the `publish` field set in packages in the projects metadata is consistent with those in release-plz configuration. 

Close https://github.com/release-plz/release-plz/issues/1896


crates/release_plz_core/src/command/update/update_config.rs
```diff
@@ -20,6 +20,8 @@ pub struct UpdateConfig {
     pub features_always_increment_minor: bool,
     /// Template for the git tag created by release-plz.
     pub tag_name_template: Option<String>,
+    /// If `false`, don't run `cargo publish`.
+    pub publish: bool,
```
marcoieni, line 24: the update command doesn't run cargo publish.

yinkaolotin:
@marcoieni could you please give this a quick look?

yinkaolotin:
> great stuff, thanks! 🙌 I did some minor adjustments such as adding some comments and reusing the `package.is_publishable()` function which was already implemented elsewhere 👍

@marcoieni I'll need to pay better attention to the codebase and write clearer comments, thanks!.

Issue #2194, opened 2025-04-23
Docs suggest default tag is prefixed by main package name


lmmx:
## Bug description

The docs suggest the default tag, given a package foo is `foo-v` when it is just `v` (plus version number).

This is indicated in the docs on "single git tag" which suggest customising it to be just "v{{ version }}" (it already is this) and also in the docs page which shares a (helpful) suggested binary release workflow YAML in which its run is gated on the presence of the "foo-v" prefix (there it's given as `my-bin-v`).

This took a moment to diagnose the cause of because GitHub won't tell you **why** a workflow doesn't run, just that it was skipped, but I can see that both the name of the Action workflow run and the tag are just "v0.1.2" (not `foo-v0.1.2`)

- Would you like to work on a fix? [y/n] Y

## To Reproduce

To reproduce: I ran `release-plz init` and followed all the steps as directed. I did not edit the CI other than to fill in my username and package binary name.

## Expected behavior

The 2 docs pages above certainly seem to suggest that the default tag prefix is supposed to be `my-bin-v` but it looks like it is just `v`.

In the 'single git tag' docs page I would expect this:

```toml
# (Optional) Customize the git tag name to remove the `my_main_package` prefix.
git_tag_name = "v{{ version }}"
```

not to say that if this is indeed already the standard behaviour

In the 'releasing binaries' page I would expect this

```yaml
    if: github.repository_owner == 'MyOwner' && startsWith(github.event.release.name, 'my-bin-v')
```

To just be

```yaml
    if: github.repository_owner == 'MyOwner' && startsWith(github.event.release.name, 'v')
```

If the default git tag name is indeed already `v{{ version }}`

## Screenshots

![Image](https://github.com/user-attachments/assets/e19d49be-4ebe-4d7c-bce3-81fbc42eab81)

## Environment

- OS: Linux Mint 22 (Ubuntu 24.04 base)
- release-plz version: 0.3.133

## Additional context

No other issues, very smooth first use :-)

lmmx:
Ohh this looks like it's my mistake, it **will** be as documented if there are multiple crates

> The git_tag_name field
> 
> [Tera template](https://keats.github.io/tera/docs/#templates) of the git tags that release-plz creates. Use this to customize the git tags name pattern.
> 
> By default, it's:
> 
>     "{{ package }}-v{{ version }}" for workspaces containing more than one public package.
>     "v{{ version }}" for projects containing a single crate or workspaces containing just one public package.
> 
> Where:
[...]

PR #2195, opened 2025-04-23
docs: mention how to modify the example CD workflow YAML

lmmx:
- Close https://github.com/release-plz/release-plz/issues/2194


marcoieni (approved):
Thanks! I removed the duplicate explanation because it was already present in the configuration field 👍

Issue #2164, opened 2025-04-04
Dependent crates in same repo still not updated when upstream crates are updated


scouten-adobe:
## Bug description

Imagine a series of crates in the same workspace: A > B > C > D

A through C are library crates, and D is a binary crate.

In a recent run using release-plz 0.3.129, which purports to fix #2034, B and C had changes and were properly updated. D had no changes, but depends on B and C, but no new release was generated.

- Would you like to work on a fix? [y/n]

Yes, but I'll be at a conference and unavailable to look into it deeply until 14 April.

## To Reproduce

See GitHub Actions run here https://github.com/contentauth/c2pa-rs/actions/runs/14248999683/job/39937004113.

When I return from conference, I can try to build a simpler repro case.

marcoieni:
PR welcome 👍

marcoieni:
https://github.com/release-plz/release-plz/pull/2196 tests this scenario and shows that release-plz supports it.

If you experience other erros where dependencies are not updated correctly, open a PR with a tests that fails 👍

PR #2196, opened 2025-04-23
test complex scenarios where dependent crates in same repo are updated

cdmurph32:
* Check for dependency changes even if there are no release commits
* Breaking changes in dependencies result in minor version update, not a point release update.
* Added tests for both scenarios.

### DIFF
--- a/crates/release_plz_core/src/command/update/package_dependencies.rs
+++ b/crates/release_plz_core/src/command/update/package_dependencies.rs
@@ -1,9 +1,8 @@
 use cargo_metadata::{Package, camino::Utf8Path, semver::Version};
 use cargo_utils::LocalManifest;
-use std::collections::HashMap;
 use toml_edit::TableLike;
 
-use crate::{PackagePath as _, semver_check::SemverCheck};
+use crate::PackagePath as _;
 
 pub trait PackageDependencies {
     /// Returns the `updated_packages` which should be updated in the dependencies of the package.
@@ -12,7 +11,7 @@ pub trait PackageDependencies {
         updated_packages: &'a [(&Package, Version)],
         workspace_dependencies: Option<&dyn TableLike>,
         workspace_dir: &Utf8Path,
-    ) -> anyhow::Result<HashMap<&'a Package, SemverCheck>>;
+    ) -> anyhow::Result<Vec<&'a Package>>;
 }
 
 impl PackageDependencies for Package {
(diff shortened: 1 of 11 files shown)


cdmurph32:
@scouten for you review

crates/release_plz_core/src/command/update/package_dependencies.rs
```diff
@@ -58,16 +62,34 @@ impl PackageDependencies for Package {
...
+    {
+        // Breaking change: major version bump or minor version bump in 0.x series
+        Some(SemverCheck::Incompatible(new_version.to_string()))
+    } else if new_version.major == old_version.major && new_version.minor > old_version.minor {
+        Some(SemverCheck::Compatible)
+    } else if new_version.major == old_version.major
+        && new_version.minor == old_version.minor
+        && new_version.patch > old_version.patch
+    {
+        //TODO: Should this be None?
```
cdmurph32, line 89: This is an open question. I'll change the function based on the answer.

scouten-adobe (approved):

scouten-adobe:
@marcoieni FYI this PR comes from a team mate of mine.

scouten-adobe:
This should fix #2164.

crates/release_plz_core/src/command/update/package_dependencies.rs
```diff
@@ -20,17 +21,20 @@ impl PackageDependencies for Package {
...
 
-        let mut deps_to_update: Vec<&Package> = vec![];
-        for (p, next_ver) in updated_packages {
-            let canonical_path = p.canonical_path()?;
-            // Find the dependencies that have the same path as the updated package.
-            let matching_deps = package_manifest
+        updated_packages
+            .iter()
+            .filter_map(|(p, next_ver)| {
+                let Ok(canonical_path) = p.canonical_path() else {
```
marcoieni, line 33: the code used `?` to report the error to the user. 
Instead here you are discarding the error and you don't even log it.
Please keep throwing the error. E.g. using `?` again. Same later in `should_update_dependency(dep, next_ver)?`

marcoieni:
In https://github.com/release-plz/release-plz/pull/2196/commits/c722d7e955cb61a59db8bb8e319f73a673d356f2 I reverted most of the changes because I didn't get them and without those changes tests pass.
Please open a new PR and explain why those changes are needed.

crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -82,6 +82,11 @@ impl Updater<'_> {
 
         let mut old_changelogs = OldChangelogs::new();
         for (p, diff) in packages_diffs {
+            if diff.is_version_published {
+                // We need to check if one of the dependencies of this package changed.
+                packages_to_check_for_deps.push(p);
+            }
```
marcoieni, line 88: I think moving this block of code here is wrong. We just need to check if a package needs to be updated because of its dependencies if `next_version != current_version`.
Otherwise we are going to update it anyway.

Imo the real issue is that the function `dependent_packages_update` doesn't detect that `binary` needs to be updated when running `TEST_LOGS=1 ENABLE_LOGS=1 RELEASE_PLZ_LOG=info cargo t release_plz_updates_binary_with_no_commits_and_dependency_change -- --nocapture`.

I spent two hours on this tonight and I'm too tired now to continue now.

PR #2205, opened 2025-05-02
docs: explain zizmor warning

### LINKED ISSUE #2200: Harden recommended CI pipeline (`persist-credentials`)
## Motivations

The pipeline described on https://release-plz.dev/docs/github/quickstart has some issues according to the [zizmor](https://github.com/woodruffw/zizmor) Github CI auditing tool:

```
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/release-plz.yml:24:9
   |
24 |         - name: Checkout repository
   |  _________-
25 | |         uses: actions/checkout@v4
26 | |         with:
27 | |           fetch-depth: 0
   | |________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/release-plz.yml:50:9
   |
50 |         - name: Checkout repository
   |  _________-
51 | |         uses: actions/checkout@v4
52 | |         with:
53 | |           fetch-depth: 0
   | |________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low
```

If possible, it would be good to set `persist-credentials: false`. See [this section](https://woodruffw.github.io/zizmor/audits/#artipacked) for the motivation of why this is a good idea.

- Would you like to implement this feature? [y/n]: Sure, it is trivial to add. What I don't know if it breaks anything that release-plz relies on.

## Solution

Ideally `persist-credentials: false` 

## Alternatives

Explicitly setting `persist-credentials: true` to document that it is actually needed.

marcoieni:
Close #2200

### DIFF
--- a/website/docs/github/security.md
+++ b/website/docs/github/security.md
@@ -88,4 +88,16 @@ by the `actions/checkout` action to run git commands like `git tag` and `git pus
 Note that this token is different with respect to the one used to call the
 GitHub API.
 
-To solve the warning, you can set `persist-credentials: true` in the `with` section.
+To solve the warning, set `persist-credentials: true` in the `with` section
+of the `actions/checkout` action:
+
+```yaml
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+# highlight-start
+          persist-credentials: true
+# highlight-end
+```
--- a/website/docs/github/security.md
+++ b/website/docs/github/security.md
@@ -58,3 +58,34 @@ jobs:
 
 This is the same approach used in the crates.io
 [repository](https://github.com/rust-lang/crates.io/blob/7e52e11c5ddeb33db70f0000bbcdfb01e9b43b0d/.github/workflows/ci.yml#L30C32-L31C1).
+
+## `zizmor` warning
+
+[zizmor](https://github.com/woodruffw/zizmor) is a static analysis tool for GitHub Actions.
+When you run it on the release-plz [workflow](./quickstart.md#3-setup-the-workflow), it will
+emit the [artipacked](https://woodruffw.github.io/zizmor/audits/#artipacked) warning:
+
+```text
+warning[artipacked]: credential persistence through GitHub Actions artifacts
+  --> .github/workflows/release-plz.yml:24:9
+   |
+24 |         - name: Checkout repository
+   |  _________-
+25 | |         uses: actions/checkout@v4
+26 | |         with:
+27 | |           fetch-depth: 0
+   | |________________________- does not set persist-credentials: false
+   |
+   = note: audit confidence → Low
+```
+
+This warning is emitted because the `actions/checkout` action does not set
+`persist-credentials: false` in the `with` section.
+
+Unfortunately, `persist-credentials` needs to be set to `true` (which is the default)
+for the release-plz action to work because release-plz needs the token generated
+by the `actions/checkout` action to run git commands like `git tag` and `git push`.
+Note that this token is different with respect to the one used to call the
+GitHub API.
+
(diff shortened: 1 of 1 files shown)


website/docs/github/security.md
```diff
@@ -58,3 +58,46 @@ jobs:
...
+```
+
+This warning is emitted because the `actions/checkout` action does not set
+`persist-credentials: false` in the `with` section.
+
+Unfortunately, `persist-credentials` needs to be set to `true` (which is the default)
+for the release-plz action to work because release-plz needs the token generated
+by the `actions/checkout` action to run git commands like `git tag` and `git push`.
+Note that this token is different with respect to the one used to call the
+GitHub API.
```
woodruffw, line 89: (Author of `zizmor` here)

I might be misunderstanding here, but this isn't necessarily true: `actions/checkout` persists the default `github.token` by default, so it's often the same one used to call the GitHub REST APIs.

You can see that here: 

https://github.com/actions/checkout/blob/85e6279cec87321a52edac9c87bce653a07cf6c2/action.yml#L24

(and then follow `authToken` through the code to see where it's used for `git` auth)

Since the [recommended workflow](https://release-plz.dev/docs/github/quickstart#3-setup-the-workflow) sets `pull-requests: write` and `contents: write` as global permissions, this means that a pretty powerful credential gets dropped on disk silently. In particular, if that credential is accidentally leaked (e.g. via an artifact), someone could pretty easily push malicious changes directly to the repo.

My recommendation would be to update the recommended workflow itself to set `persist-credentials: true` -- that doesn't fix the problem but at least it makes the choice explicit, and would make `zizmor` shut up about it :slightly_smiling_face:

marcoieni: Hi, thanks for this great tool!

I removed this phrase in https://github.com/release-plz/release-plz/pull/2206 thanks!



> My recommendation would be to update the recommended workflow itself to set `persist-credentials: true`

I thought about this, but I prefer to not set this option because it will confuse users (especially people who aren't familiar with GitHub Actions).
I prefer having a "quicker" quickstart and then document more details in other sections of the docs 👍

woodruffw: > I thought about this, but I prefer to not set this option because it will confuse users (especially people who aren't familiar with GitHub Actions).
> I prefer having a "quicker" quickstart and then document more details in other sections of the docs 👍

I understand not wanting to confuse users, but I might not be understanding why `persist-credentials: true` would be confusing -- that would be the same as the current implicit default, but would also silence `zizmor` (since it doesn't emit any finding if the user clearly communicates their intention to have credentials persisted).

Or in other words: you can preserve the current behavior (no user confusion) *and* silence `zizmor` by setting `persist-credentials: true` (**not** `persist-credentials: false`).

(P.S. Thank you for `release-plz` as well! I'm a huge fan of release automation, and evaluating `release-plz` for simplifying `zizmor`'s releases is on my TODO list :slightly_smiling_face:)

marcoieni: confusing might be the wrong word. I think it slows down the onboarding and adds cognitive load because almost no one knows about this option.



> (P.S. Thank you for `release-plz` as well! I'm a huge fan of release automation, and evaluating `release-plz` for simplifying `zizmor`'s releases is on my TODO list 🙂)

great, let me know if I can help! :)
You can find me at many Rust conferences in case you want to chat 😁

marcoieni: Hey William, FYI release-plz now works with `persist-credentials: false` 👍
This setting is now present everywhere in the docs: https://release-plz.dev/docs/github/quickstart

woodruffw: That's awesome!

PR #2210, opened 2025-05-06
fix: align registry env var name generation with Cargo

nickj:
Ensures `registry_index_url_from_env` correctly forms environment variable names for Cargo registry configurations, even when registry names contain dashes.

The sanitization logic now mirrors Cargo's behavior:
1. Replace sequences of dashes with an underscore.
2. Any non-alphanumeric character that is not a dash or underscore is invalid.

This fixes a bug where names like `my-registry` would not resolve to the expected `CARGO_REGISTRIES_MY_REGISTRY_INDEX`.

### DIFF
--- a/crates/cargo_utils/src/registry.rs
+++ b/crates/cargo_utils/src/registry.rs
@@ -10,7 +10,9 @@ const CRATES_IO_REGISTRY: &str = "crates-io";
 /// Read index for a specific registry using environment variables.
 /// <https://doc.rust-lang.org/cargo/reference/environment-variables.html>
 pub fn registry_index_url_from_env(registry: &str) -> Option<String> {
-    let Ok(env_var) = get_registry_env_var_name(registry) else { return None };
+    let Ok(env_var) = get_registry_env_var_name(registry) else {
+        return None;
+    };
     std::env::var(env_var).ok()
 }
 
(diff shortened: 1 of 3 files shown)


crates/cargo_utils/src/registry.rs
```diff
@@ -145,3 +172,36 @@ pub fn cargo_home() -> anyhow::Result<PathBuf> {
...
+            get_registry_env_var_name("registry1"),
+            "CARGO_REGISTRIES_REGISTRY1_INDEX"
+        );
+        assert_eq!(
+            get_registry_env_var_name("UPPERCASE"),
+            "CARGO_REGISTRIES_UPPERCASE_INDEX"
+        );
+        assert_eq!(
+            get_registry_env_var_name("with-special-chars!@#$%^&*()"),
+            "CARGO_REGISTRIES_WITH_SPECIAL_CHARS__INDEX"
```
marcoieni, line 200: this test case looks wrong, what do you think?

nickj: Sorry, it appears I was mistaken about multiple non-alphanumeric characters not repeating the `_` replacement.

I just tested a simple name locally like this:

```
CARGO_REGISTRIES_DEMO_INDEX="ssh://git@some.registry/crate-index.git" cargo login --registry demo
```

then, adding multiple dashes like this:
```
CARGO_REGISTRIES_DEMO___TWO_INDEX="ssh://git@some.registry/crate-index.git" cargo login --registry demo---two
```

Also it turns out the characters are much more restricted:

```
❯ CARGO_REGISTRIES_DEMO___TWO_INDEX="ssh://git@some.registry/crate-index.git" cargo login --registry demo@-@two
error: invalid character `@` in registry name: `demo@-@two`, characters must be Unicode XID characters (numbers, `-`, `_`, or most letters)
```

nickj: Updated now!

marcoieni (approved):
thanks!

PR #2221, opened 2025-05-11
chore: Indicate when diff calculation begins to avoid appearing stalled

joshka:
Previously, release-plz could appear unresponsive for a noticeable
period (e.g., 15 seconds on a reasonably sized repository) while
calculating the diff for a package. This change adds a log message at
the start of the diff calculation to inform the user about this
potentially long-running operation.

image

### DIFF
--- a/crates/release_plz_core/src/command/update/updater.rs
+++ b/crates/release_plz_core/src/command/update/updater.rs
@@ -486,7 +486,10 @@ impl Updater<'_> {
         registry_packages: &PackagesCollection,
         repository: &Repo,
     ) -> anyhow::Result<Diff> {
-        info!("determining next version for {} {}", package.name, package.version);
+        info!(
+            "determining next version for {} {}",
+            package.name, package.version
+        );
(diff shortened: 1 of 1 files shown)


crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -486,6 +486,7 @@ impl Updater<'_> {
         registry_packages: &PackagesCollection,
         repository: &Repo,
     ) -> anyhow::Result<Diff> {
+        info!("calculating diff for {} {}", package.name, package.version);
```
marcoieni, line 489:
```suggestion
        info!("determining next version for {} {}", package.name, package.version);
```

marcoieni:
Thanks, good idea!
I changed the words a bit, although it could be better probably 😅

joshka:
No problem on the wording change. Once you've seen a log message once, the actual words don't matter too much, as the majority of the value is from the presence of the log message not the content.

PR #2232, opened 2025-05-16
add max analyze commits

### LINKED ISSUE #2228: Stop analyzing git history after 100 commits by default
## Motivations
`release-plz update` on big projects using release-plz for the first time takes a long time because release-plz needs to analyze the entire git history.

- Would you like to implement this feature? [y/n]

## Solution
Introduce a new config file field called `max_commits` which by default is 100.
In [this](https://github.com/release-plz/release-plz/blob/fc8b9f360bd2a10affa549dcba85ae09bedf1a68/crates/release_plz_core/src/command/update/updater.rs#L544) function stop if we already analyzed 100 commits.

TroyKomodo:
Adds a limit to the maximum analyze commits 

fixes #2228


crates/release_plz/src/config.rs
```diff
@@ -164,6 +164,9 @@ pub struct Workspace {
     ///   `release-plz-`. So if you want to create a PR that should trigger a release
     ///   (e.g. when you fix the CI), use this branch name format (e.g. `release-plz-fix-ci`).
     pub release_always: Option<bool>,
+    /// Maximum number of commits to analize.
+    /// - If not provided: will analize all commits up until the previous release.
```
marcoieni, line 168:
```suggestion
    /// Maximum number of commits to analyze.
    /// If not provided, analyze all commits up until the previous release.
```

marcoieni:
The issue indicated that the config field should be 100 by default (Btw, I changed my mind. let's make it 1000). Can you please make this change?

Furthermore, this field should only be considered if there are no previous releases of the crate. I.e. if [registry_package](https://github.com/release-plz/release-plz/blob/4665f3cdf44a87429310202ad15df8f74b17fa6a/crates/release_plz_core/src/command/update/updater.rs#L548) is None.


Also, can you document this in https://github.com/release-plz/release-plz/blob/main/website/docs/config.md ?

PR #2233, opened 2025-05-16
add check-only & exit-status

TroyKomodo:
* adds a `check-only` flag to the `update` command which does not update the manifests and simply outputs the summary.
* adds a `exit-status` flag to the `update` command which returns a exit code of `1` when there are updates to be done.

### DIFF
--- a/crates/release_plz/src/args/update.rs
+++ b/crates/release_plz/src/args/update.rs
@@ -105,6 +105,8 @@ pub struct Update {
     /// Kind of git host where your project is hosted.
     #[arg(long, visible_alias = "backend", value_enum, default_value_t = GitForgeKind::Github)]
     forge: GitForgeKind,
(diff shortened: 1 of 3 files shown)


crates/release_plz/src/args/update.rs
```diff
@@ -105,6 +105,10 @@ pub struct Update {
     /// Kind of git host where your project is hosted.
     #[arg(long, visible_alias = "backend", value_enum, default_value_t = GitForgeKind::Github)]
     forge: GitForgeKind,
+    #[arg(long)]
+    check_only: bool,
+    #[arg(long)]
+    exit_status: bool,
```
marcoieni, line 111: these are missing docs

marcoieni:
Is `exit_status` used at all? 🤔

PR #2265, opened 2025-06-12
update git-cliff-core to v2.9.1

marcoieni:


marcoieni:
- [] I don't know if filter_commits should be true or false

marcoieni:
yolo

PR #2268, opened 2025-06-13
refactor: tidy up config path handling

joshka:
Replace trait-based config handling with a dedicated `ConfigPath`
struct that is flattened into command structs.

Motivation: prior to this change, this implemented a rust version of inheritance using traits, the reading order kind of looks like:
- main()
  - ConfigCommand::config()
    - Update::config_path()
    - args::parse_config()
    - first_file_contents()
- main()
  - Update::update_request()

It's now instead:
- main()
  - Update::update_request()


marcoieni (approved):
cool. thanks!

PR #2270, opened 2025-06-13
Add some more logging related to package publishing

tgross35:
Make it easier to track via logs what logic is being used to determine whether or not a package should be published.

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -14,7 +14,7 @@ use crates_index::{GitIndex, SparseIndex};
 use git_cmd::Repo;
 use secrecy::{ExposeSecret, SecretString};
 use serde::Serialize;
-use tracing::{debug, info, instrument, warn};
+use tracing::{debug, info, instrument, trace, warn};
 use url::Url;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


tgross35:
This helped me debug https://github.com/rust-lang/libc/pull/4481 (though doesn't fix it), the output is now:

```
2025-06-13T16:41:40.478102Z DEBUG should release: Yes
2025-06-13T16:41:40.478133Z TRACE package ctest is publishable: false
2025-06-13T16:41:40.478143Z TRACE package ctest-next is publishable: false
2025-06-13T16:41:40.478151Z TRACE package libc is publishable: false
2025-06-13T16:41:40.478158Z TRACE package libc-test is publishable: false
2025-06-13T16:41:40.478169Z  INFO nothing to release
```

It seems like the check at https://github.com/release-plz/release-plz/blob/7fa0085c8dde4a9f979eb8766f1cd1d1baf4a6b4/crates/release_plz_core/src/next_ver.rs#L167-L171 isn't correct for some reason; crate `libc` should be publishable, but `publish` is `[]`. That logic seems to have existed for a while, @marcoieni any idea why things might be different here?

tgross35:
Omg 🤦 I have a comment `publish = false # On the main branch, we don't want to publish anything` in the Cargo.toml file, this accidentally made it to the branch where we _do_ want to publish things. Easy fix :D

I guess the log output here is probably still helpful to have.

marcoieni (approved):
thanks!

PR #2308, opened 2025-07-10
fix: 🐛 Notify a user if the config file is wrongly formatted

### LINKED ISSUE #2306: Wrongly formated config files give a wrong error message
## Bug description

When giving the configuration file a wrong format (which for example doesn't match the TOML spec) a wrong error message is displayed. When reading the file that the message comes from it seems that the error should be `invalid config file {}` and not `release-plz config file not found, using default configuration` this is very misleading for a user since even when using the verbose mode (with the -v flag) the more informative error is not emited.

- Would you like to work on a fix? [**y**/n]

## To Reproduce

1. Create a wrongly formatted toml file:
```toml
[sadf]]

[package]
```

2. Run `release-plz release-pr -v`
3. The error is:
```
2025-07-09T21:40:35.530419Z  INFO release_plz::args::config_path: crates/release_plz/src/args/config_path.rs:51: release-plz config file not found, using default configuration
2025-07-09T21:40:35.530514Z  INFO release_plz::args::config_path: crates/release_plz/src/args/config_path.rs:51: release-plz config file not found, using default configuration
```

## Expected behaviour
The more informative error should be emitted.

## Environment

- OS: Arch Linux
- release-plz version: 0.3.137

kamack38:
I've added a warning when the config file is wrongly formatted. Whether it should error is up for a discussion.

Fixes #2306

### DIFF
--- a/crates/release_plz/src/args/config_path.rs
+++ b/crates/release_plz/src/args/config_path.rs
@@ -3,10 +3,10 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{Context as _, bail};
+use anyhow::{bail, Context as _};
 use clap::Args;
 use fs_err::read_to_string;
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::config::Config;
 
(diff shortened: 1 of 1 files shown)


kamack38:
@marcoieni  I've tried adding a test for it, but I don't know how to check the stdout for a certain output. Maybe we should just return an error? What do You think?

marcoieni:
I merged https://github.com/release-plz/release-plz/pull/2311 because the CI was green there 👍
I will have a look at the test when I find time!

kamack38:
Okay, so this was superseded by #2311

Issue #1668, opened 2024-09-10
Could not release on artifactory


davidB:
## Bug description
We used jfrog/artifactory (cloud) as registry for private crates.
We can `cargo fetch`  ... `cargo publish`. 

But when we try to use `release-plz` we got error to retrieve metadata

On github runner  (ubuntu-latest):

```txt
Caused by:
    0: timeout while publishing axum-extra-derive
    1: failed fetching sparse metadata
    2: error sending request for url (https://mycompany.jfrog.io/artifactory/api/cargo/mycompany-cargo-local/index/ax/i-/axum-extra-derive)
    3: client error (Connect)
    4: error:0A000460:SSL routines:ssl3_read_bytes:tlsv1 alert no application protocol:ssl/record/rec_layer_s3.c:907:SSL alert number 120
    5: error:0A000460:SSL routines:ssl3_read_bytes:tlsv1 alert no application protocol:ssl/record/rec_layer_s3.c:907:SSL alert number 120
```

On local (macos x86):

```txt
  2024-09-09T10:58:34.754858Z ERROR  can't determine if package is published

Caused by:
    0: timeout while publishing axum-extra-derive
    1: failed fetching sparse metadata
    2: error sending request for url (https://mycompany.jfrog.io/artifactory/api/cargo/mycompany-cargo-local/index/ax/i-/axum-extra-derive)
    3: client error (Connect)
    4: Fatal alert

Stack backtrace:
   0: std::backtrace::Backtrace::create
   1: std::backtrace::Backtrace::capture
   2: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   3: <tokio::time::timeout::Timeout<T> as core::future::future::Future>::poll
   4: release_plz_core::cargo::is_published::{{closure}}
   5: release_plz_core::command::release::release_package_if_needed::{{closure}}
   6: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
   7: release_plz::run::{{closure}}
   8: tokio::runtime::park::CachedParkThread::block_on
   9: release_plz::main
  10: std::sys::backtrace::__rust_begin_short_backtrace
  11: std::rt::lang_start::{{closure}}
  12: std::rt::lang_start_internal
  13: _main
  14: <unknown>
```

Timeout is not the root cause, because it failed immediately, the wrong error message/context is another topic.

I guess both errors are related because the request uses `native-tls` that imply a different behavior on mac & linux regarding SSL and HTTP 2.0

[...]

PR #2314, opened 2025-07-14
fix: Retry HTTP/2 GOAWAY from registry with HTTP/1.1

sagebind:
With some private registries that do not support HTTP/2, or perhaps falsely advertise as supporting it, reqwest does not always identify the cause of failure as a connection error. In this case, the underlying `h2` library may return a premature `GOAWAY` error from either the client or server. If this happens, we should use HTTP/1.1 instead.

This PR implements this by walking up the chain of errors to see if a "go away" error is a cause. This is a fairly explicit check, but the downside is that we have to add `h2` as a direct dependency in order to pull in the `h2::Error` type. (It is already a transitive dependency through `reqwest`.)

Related to #1668 and #2261.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4859,6 +4859,7 @@ dependencies = [
  "git-cliff-core",
  "git-url-parse",
  "git_cmd",
+ "h2",
  "http",
  "ignore",
  "itertools",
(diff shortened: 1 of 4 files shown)


marcoieni:
this looks good, thanks. Can you add the first paragraph of the PR body as a comment? I.e.

> With some private registries that do not support HTTP/2, or perhaps falsely advertise as supporting it, reqwest does not always identify the cause of failure as a connection error. In this case, the underlying h2 library may return a premature GOAWAY error from either the client or server. If this happens, we should use HTTP/1.1 instead.

sagebind:
@marcoieni Good suggestion, added some comments to the code.

marcoieni (approved):

Issue #2299, opened 2025-07-09
New option request: Always trigger release of downstream crates when a dependent crate in same repo will be published


scouten-adobe:
## Motivations

In the [content authenticity project](https://github.com/contentauth/c2pa-rs/), we have a core library (`c2pa`) which is reflected in two downstream crates (`c2pa-c-ffi` and `c2patool`). All three of these are hosted in the same repo.

We would like to always release the downstream crates whenever the core library is updated.

Currently release-plz has a hard policy of not releasing if there are no changes to the source code of those crates, which is at odds with the use case stated above.

- Would you like to implement this feature? _yes, if agreed to_

## Solution

I'd like to see a new option which would override the check for direct changes to the source code of a crate if a downstream dependency in the same repo is being updated/re-published. In other words, if `Cargo.toml`'s `dependencies` section is being updated, the `version` field should also be updated and a corresponding changelog entry should be generated as part of the `release-pr` workflow.

## Alternatives

See https://github.com/contentauth/c2pa-rs/pull/1222 for a hacky workaround to the problem we're trying to solve.

marcoieni:
I don't see the c2patool in https://github.com/contentauth/c2pa-rs/blob/main/Cargo.toml

btw if crates have the same workspace versions or you use the [version_group](https://release-plz.dev/docs/config#the-version_group-field) field, they should be released together, if the crates are unchanged.

Is this enough?

scouten-adobe:
c2patool is known as cli in the workspace Cargo.toml file.

I ran some experiments with version_group in a similar usage over in the https://github.com/scouten-adobe/rp-sandbox/ repo and did not observe the result you described. (In other words, a crate that was unchanged had an updated _dependency_ in Cargo.toml, but did not get an updated version, so it wasn't re-released.)

I'll re-run the experiment later today or tomorrow and show you what I'm seeing.

scouten-adobe:
@marcoieni I don't think this is working as described:

> btw if crates have the same workspace versions or you use the [version_group](https://release-plz.dev/docs/config#the-version_group-field) field, they should be released together, if the crates are unchanged.

I've set up a sample repo to try things out over here (https://github.com/scouten-adobe/rp-sandbox) that demonstrates what I think is missing. Since the previous release, I made a change in `rp-sandbox-a`. The other two crates (`rp-sandbox-b` and `rp-sandbox-c`) share a `version_group` with `rp-sandbox-a`.

[...]

PR #2338, opened 2025-08-14
fix: Trigger release of downstream with commit regex

cdmurph32:
This test fails because of the introduction of a commit regex, it is otherwise essentially identical to release_plz_updates_binary_when_library_changes

The issue is with the following code which does not add the crate to `packages_to_check_for_deps` if none of the commits match the regex. https://github.com/release-plz/release-plz/blob/dd42b2e11ad5df764fe5c3c29b68322eebb1ec30/crates/release_plz_core/src/command/update/updater.rs#L88-L93

The fix is to check packages for updated dependencies when there is a commit regex that doesn't match any of the commits.

### DIFF
--- a/crates/release_plz_core/src/command/update/updater.rs
+++ b/crates/release_plz_core/src/command/update/updater.rs
@@ -89,6 +89,7 @@ impl Updater<'_> {
                 && !diff.any_commit_matches(release_commits_regex)
             {
                 info!("{}: no commit matches the `release_commits` regex", p.name);
+                // We need to update this package only if one of its dependencies has changed.
                 packages_to_check_for_deps.push(p);
                 continue;
             };


cdmurph32:
This addresses https://github.com/release-plz/release-plz/issues/2299

marcoieni:
I released the new release-plz github action with this change!

cdmurph32:
Awesome. Thank you!

scouten-adobe:
Confirmed that this worked as intended with our most recent release today. See https://github.com/contentauth/c2pa-rs/releases/tag/c2patool-v0.20.1 if curious. (No direct changes in c2patool, but we wanted the new version of c2pa-rs to be included in a new release of c2patool.)

Issue #2340, opened 2025-08-16
HashKind is determined by cargo version


ncatelli:
## Bug description

I am receiving `400` errors from gix when attempting to publish a release to a private git-indexed registry. This appears to be related to the change introduced in https://github.com/release-plz/release-plz/pull/2084/files#diff-7c1b52109b073d495c93cb95833d0732a1ee80659ab31d903bf7a24417cb8124R31-R48 to support the new stable hash kind introduced with rust 1.85.

Currently release-plz determines the `HashKind` based on the version of cargo on the system.
 
- Would you like to work on a fix? [y/n] y

## To Reproduce

Below I've included a small isolated example that is dependent on my own registry. I'll attempt to introduce a more specific and reproducible example if I can.

```rust
use crates_index::{GitIndex, HashKind};

const INDEX_URL: &str = "<MY PRIVATE REGISTRY INDEX>";

fn main() -> Result<(), anyhow::Error> {
    for hash_kind in [HashKind::Legacy, HashKind::Stable].iter() {
        let index = GitIndex::from_url_with_hash_kind(INDEX_URL, hash_kind)?;
        let crates_in_index = index.crates().count();
        println!("{hash_kind:?}: {crates_in_index}");
    }

    Ok(())
}
```

```bash
> cargo run
   Compiling gix-debugging v0.1.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.57s
     Running `target/debug/gix-debugging`
Legacy: 346
Error: "gix" crate failed. If problems persist, consider deleting `~/.cargo/registry/index/github.com-1ecc6299db9ec823/`

Caused by:
    0: An IO error occurred when talking to the server
    1: Received HTTP status 400
> rustc --version
rustc 1.86.0 (05f9846f8 2025-03-31)
```

## Expected behavior

If unable to update the index with the new `Stable` hash, I think it should fallback to attempting the `Legacy` hash.

## Environment
- OS: Mac OSX
- release-plz version: 0.3.139+
- rust 1.85+

## Additional context

https://github.com/release-plz/release-plz/issues/2334#issuecomment-3192380167

marcoieni:
> This PR determines the `HashKind` based on the version of cargo on the system.

?? This is a GitHub issue, not a PR 🤔

> If unable to update the index with the new `Stable` hash, I think it should fallback to attempting the `Legacy` hash.

sounds good to me. PR welcome. Add some comments please. No tests needed I think given the nature of the change

[...]

PR #2341, opened 2025-08-16
feat: add a fallback index to attempt the legacy hash if the primary index fails

ncatelli:
## General

This PR introduces a fallback index field to the `CargoRegistry` struct. This fallback index is the same index using the `HashKind::Legacy` hash kind if the cargo version is 1.85+. When releasing a crate, the cli will attempt the fallback index if the primary index fails.

relates to #2340

### DIFF
--- a/crates/release_plz_core/src/cargo_hash_kind.rs
+++ b/crates/release_plz_core/src/cargo_hash_kind.rs
@@ -33,6 +33,7 @@ pub fn try_get_fallback_hash_kind(
     use crates_index::HashKind;
 
     match hash_kind {
+        // Some registries still use the legacy hash kind, so we fallback to it.
         HashKind::Stable => Some(HashKind::Legacy),
         HashKind::Legacy => None,
     }
--- a/crates/release_plz_core/src/cargo_hash_kind.rs
+++ b/crates/release_plz_core/src/cargo_hash_kind.rs
@@ -27,6 +27,17 @@ pub fn get_hash_kind() -> anyhow::Result<crates_index::HashKind> {
     Ok(hash_kind)
 }
 
+pub fn try_get_fallback_hash_kind(
+    hash_kind: &crates_index::HashKind,
+) -> Option<crates_index::HashKind> {
+    use crates_index::HashKind;
+
+    match hash_kind {
+        HashKind::Stable => Some(HashKind::Legacy),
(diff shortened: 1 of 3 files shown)


ncatelli:
I may rework this a bit as I don't love the introduction of a new field for the index.

As for functionality, I'll test this branch off a known impacted private registry on Monday.

ncatelli:
I've tested this change against a known impacted registry and have confirmed this resolves the issue with rust versions > 1.85.0.

crates/release_plz_core/src/cargo_hash_kind.rs
```diff
@@ -78,4 +89,21 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_registry_fallback() {
+        use crates_index::HashKind;
+
+        for (stdout, _expected_fallback_hash_kind) in [
+            (
+                "cargo 1.85.0 (d73d2caf9 2024-12-31)",
+                Some(HashKind::Stable),
```
marcoieni: if you write `Some(HashKind::Legacy),` here, the test pass.
Probably `assert!(matches!` doesn't work as expected.

crates/release_plz_core/src/command/release.rs
```diff
@@ -741,28 +779,59 @@ fn registry_indexes(
...
+            SparseIndex::from_url_with_hash_kind(index_url, hash_kind).map(CargoIndex::Sparse);
+        let maybe_fallback = fallback_hash.map(|hash_kind| {
+            SparseIndex::from_url_with_hash_kind(index_url, &hash_kind).map(CargoIndex::Sparse)
+        });
+
+        (maybe_primary, maybe_fallback)
+    } else {
+        let index_url = format!("registry+{u}");
+        let maybe_primary =
+            GitIndex::from_url_with_hash_kind(&index_url, hash_kind).map(CargoIndex::Git);
```
marcoieni, line 817: `(_, Some(Err(e))` worries me. Maybe it can happen that the primary index works just fine but we fail because of weird errors on the fallback?

ncatelli: Good catch. I agree. I think there is a missing case in that pattern like `(Ok(primary_index), Some(Err(_))) => Ok((primary_index, None))` that works off the assumption that we should always try to push through a successful primary index.

PR #2353, opened 2025-08-26
docs: don't declare action outputs as boolean

robjtede:
GHA outputs are untyped, this was a slight head-scratcher without debug logs:

When using for example:
```
  if: ${{ steps.release-plz.outputs.prs_created == true }}
```

Debug logs show:
```
##[debug]......Evaluating String:
##[debug]......=> 'prs_created'
##[debug]....=> 'true'
##[debug]....Evaluating Boolean:
##[debug]....=> true
##[debug]..=> false
##[debug]=> false
##[debug]Expanded: (true && ('true' == true))
                             ^^^^^^ release-plz output
##[debug]Result: false
                 ^^^^^  no implicit type-cast
```

Proposed solution to just not declare these as bools. Happy to come up with some other way of clarifying this.

### DIFF
--- a/website/docs/github/output.md
+++ b/website/docs/github/output.md
@@ -16,7 +16,7 @@ When the action runs with `command: release-pr`, it outputs the following proper
     Each entry is an object containing:
     - `package_name`: The name of the edited package.
     - `version`: The next version of the package. The PR updates the package to this version.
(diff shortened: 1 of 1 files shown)


marcoieni:
maybe we can specify what the possible values are? E.g. "true" or "false"?

marcoieni:
thanks!

Issue #2354, opened 2025-08-27
Use GitHub API to push tags/branches instead of `git` CLI


risu729:
## Motivations

I don't want to set `persistent-credentials: true` for `actions/checkout` for security reasons, but it's necessary now.
https://release-plz.dev/docs/github/security#zizmor-warning.

Also, passing a PAT or a GitHub App token to `actions/checkout` just for saving credentials doesn't seem ideal.
https://release-plz.dev/docs/github/token#use-a-personal-access-token

I know I'm paranoid, but I believe it's better for most use cases.

- Would you like to implement this feature? [y/n]

Yes!

## Solution

Pushing commits was already migrated to the API in https://github.com/release-plz/release-plz/pull/1201 (and https://github.com/release-plz/release-plz/pull/1513).

So, I think the Git operations we need to migrate to the API are:
- tag creation
- branch creation
- branch update (force push)
- branch deletion

These are all available from the GitHub API.

## Additional context

I'm not sure how this affects private repositories, as they require authentication even for read access.
I think it's possible to migrate all git CLI usages to GitHub API, but it might be fine to stick with `persistent-credentials: true` for them.

marcoieni:
PRs welcome! Please create one PR at a time (for each operation), so we keep PRs small.
It would be nice to do the same for gitlab and gitea, but you can also fallback on the git cli for those.
Please make sure that when creating the tag, the tag has the same characteristics of the tag created by the cli.

risu729:
Thank you!
I'll work on it later.

marcoieni:
your changes are now live with the latest github action release! Does that mean that this is done and we just need to update the docs? 😮
If yes, can you test it in one of your repos and share the results?

marcoieni:
tested in https://github.com/marcoieni/rust-workspace-example and this now works, thanks!

risu729:
Sorry for the late reply, thank you for testing!
I also confirmed it's working well.

risu729:
Just a small thing, but we don't need `release-plz/git-config` on GitHub unless we're using tag signing.
https://github.com/release-plz/action/blob/acb9246af4d59a270d1d4058a8b9af8c3f3a2559/action.yml#L68-L69

marcoieni:
Oh, right, should we remove that action completely and assume that people configured the git user by themselves if they want to sign tags? I'm not sure if it's a safe assumption.

Also, some people might run commits and push frok git after running release-plz to commit on the release pr.
[...]

PR #1513, opened 2024-06-09
fix(#1487): don't close release PR on force-push

zvolin:
Fixes #1487 

Force pushing a branch that links to an open PR with the content of a base branch automatically closes it. To prevent that we create a new temporary branch for release, commit there and then force push original one.

Issue #1487, opened 2024-05-23
Release PR is closed when updating


baszalmstra:
## Bug description

In [this repository](https://github.com/mamba-org/rattler) Im making extensive use of release-plz. I absolute love it. However, for a little while now it appears that when release-plz _updates_ the release PR it immediately closes it. I cannot find anything in the logs that suggests that release-plz is doing this. 

- Would you like to work on a fix? [n]

## To Reproduce

I dont have isolated steps to reproduce this issue but I can show steps that exhibit this behavior in the aformentioned repository.

If you take a look at all the workflow runs of release-plz for the main branch: https://github.com/mamba-org/rattler/actions/workflows/release-rust.yaml

You can see that when PR 666 was merged release-plz create a new release PR. This is the workflow run: https://github.com/mamba-org/rattler/actions/runs/9186703934/job/25262873395

Release-plz notices changes in some of the crates and opens a release PR as can be seen in the logs:

```
2024-05-22T07:03:29.666711Z  INFO  opened pr: https://github.com/mamba-org/rattler/pull/668
    in open_pr
    in release_pr

release_pr_output: {"prs":[{"base_branch":"main","head_branch":"release-plz-2024-05-22T07-03-26Z","html_url":"https://github.com/mamba-org/rattler/pull/668","number":668}]}
```

However the [next run of release-plz](https://github.com/mamba-org/rattler/actions/runs/9193700451/job/25285384837) (the release pr was not merged), seems to update the release PR.

```

  2024-05-22T15:06:28.471231Z  INFO  updated pr https://github.com/mamba-org/rattler/pull/668
    in release_pr

release_pr_output: {"prs":[{"base_branch":"main","head_branch":"release-plz-2024-05-22T07-03-26Z","html_url":"https://github.com/mamba-org/rattler/pull/668","number":668}]}
```

Indeed the [release PR](https://github.com/mamba-org/rattler/pull/668) also includes the changes of the last merged PR. However, it was also immediately closed. Reading the logs I can see no indication as to why this would happen.

## Expected behavior

I would expect the release PR to not be closed. Manually reopening and merging the PR seems to be the workaround.

## Environment

- release-plz version: 0.3.68

Pr0methean:
I'm experiencing this too; examples are https://github.com/zip-rs/zip2/pull/147 and https://github.com/zip-rs/zip2/pull/152, both of which I had to manually reopen.

MarcoIeni:
[...]

PR #2355, opened 2025-08-28
feat: use api to create branches

risu729:
As in https://github.com/release-plz/release-plz/issues/2354, use the API to create a new branch on the remote instead of the local Git CLI.

Currently, the implementations for Gitea/GitLab seem useless, but I'll migrate commits to Gitea/GitLab, as well as GitHub, so we can merge the implementation.

ref:
- https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#create-a-reference
- https://docs.gitlab.com/api/branches/#create-repository-branch
- https://gitea.com/api/swagger#/repository/repoCreateBranch

### DIFF
--- a/crates/release_plz_core/src/command/release_pr/mod.rs
+++ b/crates/release_plz_core/src/command/release_pr/mod.rs
@@ -304,7 +304,7 @@ async fn handle_opened_pr(
 }
 
 async fn create_pr(git_client: &GitClient, repo: &Repo, pr: &Pr) -> anyhow::Result<ReleasePr> {
-    if matches!(git_client.forge, ForgeType::Github) {
+    if git_client.forge == ForgeType::Github {
         github_create_release_branch(git_client, repo, &pr.branch, &pr.title).await?;
     } else {
         create_release_branch(repo, &pr.branch, &pr.title)?;
@@ -329,7 +329,7 @@ async fn update_pr(
             repository.original_branch()
         )
     })?;
-    if matches!(git_client.forge, ForgeType::Github) {
+    if git_client.forge == ForgeType::Github {
         github_force_push(git_client, opened_pr, repository).await?;
     } else {
         force_push(opened_pr, repository)?;
--- a/crates/release_plz_core/src/command/release_pr/mod.rs
+++ b/crates/release_plz_core/src/command/release_pr/mod.rs
@@ -304,6 +304,7 @@ async fn handle_opened_pr(
 }
 
 async fn create_pr(git_client: &GitClient, repo: &Repo, pr: &Pr) -> anyhow::Result<ReleasePr> {
+    repo.checkout_new_branch(&pr.branch)?;
     if git_client.forge == ForgeType::Github {
         github_create_release_branch(git_client, repo, &pr.branch, &pr.title).await?;
     } else {
@@ -482,7 +483,6 @@ fn create_release_branch(
     release_branch: &str,
     commit_message: &str,
 ) -> anyhow::Result<()> {
-    repository.checkout_new_branch(release_branch)?;
     add_changes_and_commit(repository, commit_message)?;
     repository.push(release_branch)?;
     Ok(())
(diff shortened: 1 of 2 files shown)


risu729:
I'm not familiar with GitLab and Gitea, but it seems GitLab and Gitea don't support commit signing via API.
I found one for GitLab, but it requires a setting, and not sure what happens with a repository access token.
https://docs.gitlab.com/user/project/repository/signed_commits/web_commits/

Since I guess some users might be running `release-plz` locally or with GPG keys configured, I think we shouldn't use the commit API for GitLab and Gitea, but migrating commits to use the API will make the implementation simpler.
What do you think about this?

risu729:
Additionally, if something similar to https://github.com/release-plz/release-plz/issues/1487 happens on GitLab or Gitea, we need a workaround like https://github.com/release-plz/release-plz/pull/1513.
However, it seems Gitea doesn't support updating ref via API, so the same workaround cannot be applied.
ref: https://github.com/go-gitea/gitea/pull/21328, https://github.com/go-gitea/gitea/issues/35368

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -304,10 +304,10 @@ async fn handle_opened_pr(
 }
 
 async fn create_pr(git_client: &GitClient, repo: &Repo, pr: &Pr) -> anyhow::Result<ReleasePr> {
-    repo.checkout_new_branch(&pr.branch)?;
     if matches!(git_client.forge, ForgeType::Github) {
         github_create_release_branch(git_client, repo, &pr.branch, &pr.title).await?;
     } else {
+        repo.checkout_new_branch(&pr.branch)?;
```
marcoieni: why you made this change?

marcoieni: the rest looks good. We can use the API to create branches in gitlab and gitea, too as you did, just for consistency. 
For simplicity, let's not worry about migrating gitlab and gitea to committing via API since there are some concerns. Let's first migrate github and evaluate later 👍

risu729: Thanks!
It was not to create a branch locally, because it's not required for GitHub.
I modified it to create branches using the API in GitLab and Gitea. It doesn't fetch the created branches, but it should be fine, as we can assume that both the local and remote branches point to the same head.

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -494,7 +496,6 @@ async fn github_create_release_branch(
     release_branch: &str,
     commit_message: &str,
 ) -> anyhow::Result<()> {
-    repository.push(release_branch)?;
```
marcoieni, line 497: I wonder if we should create the branch here, instead of creating before calling this function.
Maybe the function `github_force_push` would benefit from creating the branch here?

I wonder if there's a bug in the current implementation because here we commit to a branch that doesn't exist
https://github.com/risu729/release-plz/blob/33c7b88db6422c40c7333e29683ac2b5bb39f3b1/crates/release_plz_core/src/command/release_pr/mod.rs#L442

marcoieni: btw this function is only one line now, so if we resolve this comment without any code change, we can delete this function

risu729: Sorry, that was my mistake. I moved the branch creations into `create_release_branch` and `github_create_release_branch` functions.
For now, `create_release_branch` doesn't create a branch remotely using the API, but if it's better for consistency, I will refactor so.

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -304,8 +304,7 @@ async fn handle_opened_pr(
 }
 
 async fn create_pr(git_client: &GitClient, repo: &Repo, pr: &Pr) -> anyhow::Result<ReleasePr> {
-    repo.checkout_new_branch(&pr.branch)?;
```
marcoieni, line 307: what's the issue in checking out the new branch in github? 🤔 So that we keep the same behavior across all forges.

You don't checkout because we use the API to commit in github, right? 

I'm worried about bugs where stuff like `git checkout -` brings you back to different branches 😅

risu729: Sorry, I misunderstood that to use API for GitLab and Gitea to create branches.
Yeah, I think we can keep local checkout here.

marcoieni (approved):

PR #2356, opened 2025-08-29
feat: use api to delete branches

risu729:
As in https://github.com/release-plz/release-plz/issues/2354, use the API to delete a new branch on the remote instead of the local Git CLI.

ref:
- https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#delete-a-reference
- https://docs.gitlab.com/api/branches/#delete-repository-branch
- https://gitea.com/api/swagger#/repository/repoDeleteBranch

I added `urlencoding` crate to encode branch names for GitLab/Gitea, but if it's better to stick to replacing `/` with `%2F`, I'll refactor so.

Also, since `Drop` trait doesn't support `async`, it looks messy. I think we can remove `TmpBranch` and manually handle the cleanup.

### DIFF
--- a/crates/release_plz_core/src/command/release_pr/mod.rs
+++ b/crates/release_plz_core/src/command/release_pr/mod.rs
@@ -427,36 +427,40 @@ async fn github_force_push(
     let tmp_release_branch = format!("{}-tmp-{}", pr.branch(), rand::random::<u32>());
     repository.checkout_new_branch(&tmp_release_branch)?;
 
-    let run = async {
-        // Push the "Verified" commit in the temporary branch using
-        // the GitHub API.
-        // We push the release-plz changes to the temporary branch instead of the release PR branch because:
-        // - You can't force-push with the GitHub API, so we can't commit to the release PR branch
-        //   directly if we want a "Verified" commit.
-        // - If we revert the last commit of the release PR branch, GitHub will close the release PR
-        //   because the branch is the same as the default branch. So we can't revert the latest release-plz commit and push the new one.
-        // To learn more, see https://github.com/release-plz/release-plz/issues/1487
-        github_create_release_branch(client, repository, &tmp_release_branch, &pr.title).await?;
-
-        repository.fetch(&tmp_release_branch)?;
-
-        // Rewrite the PR branch so that it's the same as the temporary branch.
-        repository.force_push(&format!(
-            "{}/{}:{}",
-            repository.original_remote(),
-            tmp_release_branch,
-            pr.branch()
-        ))?;
-        Ok::<(), anyhow::Error>(())
-    };
-
-    let result = run.await;
-
+    let result = execute_github_force_push(client, pr, repository, &tmp_release_branch).await;
+    // Ensure the temporary branch is deleted even if the push fails.
     if let Err(e) = client.delete_branch(&tmp_release_branch).await {
         tracing::error!("cannot delete branch {tmp_release_branch}: {e:?}");
     }
-
     result
+}
+
+async fn execute_github_force_push(
+    client: &GitClient,
+    pr: &GitPr,
+    repository: &Repo,
+    tmp_release_branch: &str,
+) -> anyhow::Result<()> {
+    // Push the "Verified" commit in the temporary branch using
+    // the GitHub API.
+    // We push the release-plz changes to the temporary branch instead of the release PR branch because:
+    // - You can't force-push with the GitHub API, so we can't commit to the release PR branch
+    //   directly if we want a "Verified" commit.
+    // - If we revert the last commit of the release PR branch, GitHub will close the release PR
+    //   because the branch is the same as the default branch. So we can't revert the latest release-plz commit and push the new one.
+    // To learn more, see https://github.com/release-plz/release-plz/issues/1487
+    github_create_release_branch(client, repository, tmp_release_branch, &pr.title).await?;
+
+    repository.fetch(tmp_release_branch)?;
+
+    // Rewrite the PR branch so that it's the same as the temporary branch.
+    repository.force_push(&format!(
+        "{}/{}:{}",
+        repository.original_remote(),
+        tmp_release_branch,
+        pr.branch()
+    ))?;
+    Ok(())
 }
 
 fn create_release_branch(
--- a/crates/release_plz_core/src/command/release_pr/mod.rs
+++ b/crates/release_plz_core/src/command/release_pr/mod.rs
@@ -234,7 +234,7 @@ async fn open_or_update_release_pr(
             )
             .await
         }
-        None => create_pr(&git_client, repo, &new_pr).await,
+        None => create_pr(git_client, repo, &new_pr).await,
     }?;
     let release_pr = ReleasePr {
         releases: packages_to_update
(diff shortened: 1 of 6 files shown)


crates/release_plz_core/src/git/forge.rs
```diff
@@ -902,6 +902,31 @@ impl GitClient {
         };
         format!("{}/{commits_api_path}{commit}", self.repo_url())
     }
+
+    /// Delete a branch.
+    pub async fn delete_branch(&self, branch_name: &str) -> Result<(), anyhow::Error> {
+        let url = match self.forge {
+            ForgeType::Github => format!("{}/git/refs/heads/{}", self.repo_url(), branch_name),
```
marcoieni, line 966: should we encode the branch name here as well?

marcoieni:
> I think we can remove `TmpBranch` and manually handle the cleanup.

if this PR works, let's do this refactoring in a separate PR if really want to do this. Personally the code of this PR looks fine to me, but I'm open to review any refactor you propose :)

risu729:
> if this PR works, let's do this refactoring in a separate PR if really want to do this. Personally the code of this PR looks fine to me, but I'm open to review any refactor you propose :)

If the current implementation is fine, it's fine. I thought using `tokio::spawn` might not be ideal because it doesn't wait until the branch is deleted.

crates/release_plz_core/src/git/forge.rs
risu729: I tested it now, but GitHub doesn't support an encoded ref name. I confirmed it works without encoding.

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -456,25 +458,33 @@ async fn github_force_push(
...
-        if let Err(e) = self.repository.delete_branch_in_remote(&self.name) {
-            tracing::error!("cannot delete branch {}: {:?}", self.name, e);
-        }
+        let client = self.client.clone();
+        let name = self.name.clone();
+        tokio::spawn(async move {
+            if let Err(e) = client.delete_branch(&name).await {
+                tracing::error!("cannot delete branch {name}: {e:?}");
+            }
+        });
```
marcoieni:
```suggestion
        // Create a new runtime to block on the async operation
        match tokio::runtime::Runtime::new() {
            Ok(rt) => {
                rt.block_on(async move {
                    if let Err(e) = client.delete_branch(&name).await {
                        tracing::error!("cannot delete branch {name}: {e:?}");
                    }
                });
            }
            Err(e) => {
                tracing::error!("failed to create runtime for branch deletion: {e:?}");
                return;
            }
        }
```
what do you think about doing this to wait for the completion?

risu729: Thanks! However, I found that it's not necessary. In the previous implementation, it waits, but it doesn't handle the error either.
I'm sorry for the confusion.

marcoieni: mmmh, why do you say it waits?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0943eec06529e6fcf467fbfd139edac1
Image

risu729: Oh sorry, I meant the previous implementation with Git CLI.

marcoieni: ok, what do you think about block_on vs spawn?
With block_on we wait so we make sure that the branch is deleted, right? 🤔
With spawn it might happen that release-plz exits because main finishes, but we don't log a possible error for example.

risu729: I checked the `block_on` implementation, but it panics.

```
thread 'main' panicked at /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/runtime/scheduler/multi_thread/mod.rs:86:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b329a7104b937a9c3b090566af69b822

The `spawn` one works fine, but yeah, it might not log errors if the main thread exits earlier.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=973c19f768259c71a13a10e26569e0a7

risu729: If I remove the `TmpBranch`, it would be like this.

```rust
async fn github_force_push(
    client: Arc<GitClient>,
    pr: &GitPr,
    repository: &Repo,
) -> anyhow::Result<()> {
    // Create a temporary branch.
    let tmp_release_branch = format!("{}-tmp-{}", pr.branch(), rand::random::<u32>());
    repository.checkout_new_branch(&tmp_release_branch)?;

    let run = async {
        // Push the "Verified" commit in the temporary branch using
        // the GitHub API.
        // We push the release-plz changes to the temporary branch instead of the release PR branch because:
        // - You can't force-push with the GitHub API, so we can't commit to the release PR branch
        //   directly if we want a "Verified" commit.
        // - If we revert the last commit of the release PR branch, GitHub will close the release PR
        //   because the branch is the same as the default branch. So we can't revert the latest release-plz commit and push the new one.
        // To learn more, see https://github.com/release-plz/release-plz/issues/1487
        github_create_release_branch(&client, repository, &tmp_release_branch, &pr.title).await?;

        repository.fetch(&tmp_release_branch)?;

        // Rewrite the PR branch so that it's the same as the temporary branch.
        repository.force_push(&format!(
            "{}/{}:{}",
            repository.original_remote(),
            tmp_release_branch,
            pr.branch()
        ))?;
        Ok::<(), anyhow::Error>(())
    };

    let result = run.await;

    if let Err(e) = client.delete_branch(&tmp_release_branch).await {
        tracing::error!("cannot delete branch {tmp_release_branch}: {e:?}");
    }

    result
}
```

marcoieni:
there are some merge conflicts unfortunately

crates/release_plz_core/src/command/release_pr/mod.rs
marcoieni: sure, let's remove TmpBranch. Please extract the "run" async block to a function because it looks a bit weird imo

marcoieni: maybe this would allow to remove the Arc?

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -424,58 +424,39 @@ async fn github_force_push(
...
+
+        // Rewrite the PR branch so that it's the same as the temporary branch.
+        repository.force_push(&format!(
+            "{}/{}:{}",
+            repository.original_remote(),
+            tmp_release_branch,
+            pr.branch()
+        ))?;
+        Ok::<(), anyhow::Error>(())
+    };
```
marcoieni: what's the point of the async block? We are already inside an async function so it shouldn't be necessary, right?

crates/release_plz_core/src/command/release_pr/mod.rs
risu729: Refactored! I didn't have a good idea for the function name

crates/release_plz_core/src/command/release_pr/mod.rs
risu729: It's outdated, but it was required to use `await`. Otherwise, the compiler complains.

marcoieni (approved):

Issue #1773, opened 2024-10-15
Sign github tags created by release command


LNSD:
## Motivations

The creation (and push) of Git tags is automated by the `release-plz release` command. The newly created tags are not signed, so they appear in the GitHub UI as "unverified": 

![image](https://github.com/user-attachments/assets/b0c06576-361e-443d-a5a9-bb181d796ee3)

- Would you like to implement this feature? **Yes**

## Proposed Solution

Currently, there's no GitHub API that supports signing git tags (see this [GitHub discussion](https://github.com/orgs/community/discussions/27016)), although a REST API exists to create signed/verified commits using the GitHub token.

To sign tags in the CI using git's `--sign` option, we could use an environment variable (e.g., `GPG_SIGNING_KEY`) to pass a _GPG sub-key_ ID to the `release` command. This environment variable can be optional and keep the current behavior if not provided.

Of course, this would require the GPG key to be imported in the action container's GPG keyring using, for instance, the [Import GPG action](https://github.com/marketplace/actions/import-gpg).

MarcoIeni:
> a REST API exists to create signed/verified commits using the GitHub token.

We are already using this, right?
Is this useful to create signed tags, too?

LNSD:
> We are already using this, right?
> Is this useful to create signed tags, too?

I meant that there is no API to create signed tags. The only way to do this is to use the regular signing mechanism, i.e., GPG, to sign the commits.

MarcoIeni:
Have you checked if we can create a signed tag with [this](https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object) API?

LNSD:
> I see.
> 
> I wonder how release-please sings tags. [googleapis/release-please/tags](https://github.com/googleapis/release-please/tags) 🤔

I think they are using an external _in-house-built_ service. At least, that is what [the `.kokoro/` directory files](https://github.com/googleapis/release-please/blob/main/.kokoro/trampoline_v2.sh) suggests to me.

MarcoIeni:
I see.

> Of course, this would require the GPG key to be imported in the action container's GPG keyring using, for instance, the [Import GPG action](https://github.com/marketplace/actions/import-gpg).

Can't this be done by the user before running the release-plz action?
Release-plz can add the `-s` to the `git tag` if it detects that a gpg key is configured (I guess if the following is present in the .gitconfig:

```
[tag]
    gpgSign = true
```

what do you think?

[...]

PR #2362, opened 2025-08-30
feat: use api to create tags

risu729:
As in https://github.com/release-plz/release-plz/issues/2354, use the API to create a new tag on the remote instead of the local Git CLI.

ref:
- https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object
- https://docs.gitlab.com/api/tags/#create-a-new-tag
- https://gitea.com/api/swagger#/repository/repoCreateTag

This creates an annotated tag, so it should be the same as `git tag -m <message> <name>`.
GitLab and Gitea create an annotated tag if `message` isn't empty.

The tags are not signed in the current implementation either.
I think they don't support tag signing via API, at least GitHub doesn't.


marcoieni:
> I think they don't support tag signing via API, at least GitHub doesn't.

is this a regression, right? See https://github.com/release-plz/release-plz/issues/1773
If yes, probably we should document it and make it configurable so that people can decide whether they want to use the git cli to create a commit or not? Not sure 🤔
Some repos might require this I suppose.
This is annoying.

risu729:
> is this a regression, right?

I didn't notice it, but yes, it's a regression.
We can also check `git config tag.gpgSign`, but I think it's better to add an option to use Git CLI, as an implicit behaviour might cause some confusion.

marcoieni:
> We can also check `git config tag.gpgSign`, but I think it's better to add an option to use Git CLI, as an implicit behaviour might cause some confusion.

actually you are making a good point. We could check if `git config tag.gpgSign` is set, at least as a starting point. If someone complains we can add the config option! I think the less config options, the better, there are too many already 😅

risu729:
> actually you are making a good point. We could check if `git config tag.gpgSign` is set, at least as a starting point. If someone complains we can add the config option! I think the less config options, the better, there are too many already 😅

That's right. I fixed to check `git config tag.gpgSign`.
I used `--default false` to avoid `unwrap_or(false)`, in case `git config` failed for some reasons.

risu729:
The test is failing because the tag doesn't exist locally, but I'm not sure how to solve it.

marcoieni:
> The test is failing because the tag doesn't exist locally, but I'm not sure how to solve it.

maybe you can use the API to check if the tag exists remotely, OR you can also run a git command to download the tags available in the remote locally.

risu729:
Thanks, I added `git fetch --tags`.

risu729:
I tested this too on a real repository.

marcoieni (approved):

PR #2363, opened 2025-08-30
chore: update release-plz workflow permissions

risu729:
This PR is to show my goal in https://github.com/release-plz/release-plz/issues/2354.

After all related PRs are merged, this should work. I will update the docs later.


.github/workflows/release-plz.yml
```diff
@@ -12,32 +12,30 @@ jobs:
     if: ${{ github.repository_owner == 'release-plz' }}
     environment: release-plz
     permissions:
-      contents: write
+      contents: read
```
marcoieni, line 17: can't we just remove this permissions block? Isn't `contents: read` the default?

marcoieni:
Btw I solved the merge conflicts and pushed a few commits to restore your changes and fixes the mistakes I made when merging the conflict.

.github/workflows/release-plz.yml
```diff
@@ -12,32 +12,30 @@ jobs:
...
           toolchain: stable
+      # Generating a GitHub token, so that PRs and tags created by
+      # the release-plz-action can trigger actions workflows.
+      - name: Generate GitHub token
+        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
+        id: generate-token
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          permission-contents: write
```
marcoieni, line 41: why is this required now if before wasn't?
The permissions of the app are managed in the app settings anyway, why do we need to repeat them here? 🤔

.github/workflows/release-plz.yml
risu729: `read: all` is the default, but it's fine to remove the whole block. I think read permissions are safe.

.github/workflows/release-plz.yml
risu729: Yes, it's not strictly required. However, we don't need the `pull-requests` permission for `release-plz release`, so this would narrow the permissions.
Also, even if the app has broad permissions, the `permission` settings safely restrict them from being used. However, this is just my preference; it's not required.

marcoieni: oh, that's actually nice. I'm fine leaving this, can you add a comment that explains this?

.github/workflows/release-plz.yml
marcoieni: oh you mean that `permissions: {}` has more permissions than `permissions: contents: read`?
If that's true, than let's keep `contents: read`. I wasn't aware about this. Can you link some docs or discussions that explains this?

If that's false, let's use `permissions: {}` (zizmor suggest to have an empty permission block instead of having none)

risu729: Oh, sorry, I was wrong.
Not setting `permissions` means `contents: read` and `packages: read` (it wasn't `read: all`).

The document is here. It might be different depending on your repository/organisation settings, but the default is that.
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions

Also, zizmor recommends `permissions: {}` to remove all permissions in the workflow level, and give required permissions in the job levels.
https://docs.zizmor.sh/audits/#excessive-permissions
`permissions: {}` means no permissions at all, not the default permission.

marcoieni: ok, then let's use `permissions: {}` and let's see if we have any issues related to permissions 👍

risu729: I added it, only for `release-plz.yml`.

marcoieni: why not replacing it here, too? 🤔
I mean `permissions: {}` instead of `permissions: contents: read`

risu729: Oh I see, it's required to check out the repository.

PR #2366, opened 2025-09-01
docs: update docs about actions/checkout

risu729:
See https://github.com/release-plz/release-plz/issues/2354.
I think we should document tag signing, but I haven't tested it, so I'm not really sure how it works.
Also, GitHub recently added immutable releases, which are more reliable than tag signing.
https://github.blog/changelog/2025-08-26-releases-now-support-immutability-in-public-preview/


marcoieni:
I converted this to draft because we need to merge more PRs before this becomes ready to be merged 👍 
I.e. we need to finish the implementation.
btw I just released a new version of the binary and the action, so your first changes are already live :)

marcoieni:
> GitHub recently added immutable releases, which are more reliable than tag signing.
> https://github.blog/changelog/2025-08-26-releases-now-support-immutability-in-public-preview/

Aren't tag signing and immutability both useful in different ways?

From my understanding, tag signing ensure WHO signed the tag, while immutability ensures the tag didn't change.
So probably we should support both?

risu729:
> Aren't tag signing and immutability both useful in different ways?
> 
> From my understanding, tag signing ensure WHO signed the tag, while immutability ensures the tag didn't change. So probably we should support both?

That's right, we should support both.

website/docs/github/security.md
```diff
@@ -102,21 +102,11 @@ warning[artipacked]: credential persistence through GitHub Actions artifacts
...
-# highlight-end
-```
+You can safely set it to `false` to avoid the warning, as release-plz does not need
+the local credential generated by the `actions/checkout` action to run.
+
+:::info
+If you use tag signing with release-plz, you need the local credential to push the signed tag to the
+remote repository.
+In this case, you can suppress the warning by setting `persist-credentials: true` explicitly.
+:::
```
marcoieni: imo we could delete this section entirely and set explicitly `persist-credentials: false` in the docs and in `release-plz init`

risu729: I added `persist-credentials: false` and removed the whole `zizmor` section, but let me know if we should mention why we can set `persist-credentials: false` somewhere in the docs.

risu729: For the `init` command, I opened https://github.com/release-plz/release-plz/pull/2417

website/docs/github/output.md
```diff
@@ -231,6 +233,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
+          persist-credentials: false
```
marcoieni: for this example where we use git push, we probably need persist-credentials: true, right? Maybe we can explicitly write `persist-credentials: true` and add a comment explaining why it's needed and that it's the default.

website/docs/github/token.md
```diff
@@ -135,13 +134,6 @@ steps:
       ...
 ```
 
-:::warning
-As shown in the example below,
-you need to add the `token` field to the `actions/checkout` step, too.
-This allows release-plz to use the PAT also when spawning `git` commands,
-such as `git tag`.
-:::
```
marcoieni, line 143: maybe here we should explain that the token is still needed in case of signed tags.

website/docs/github/output.md
risu729: Sorry, I didn't notice that, thanks.
Fixed but I think the comments are not good enough.

marcoieni (approved):
thanks a lot!

risu729:
Thank you so much for your support and the document improvements!

Issue #2390, opened 2025-09-09
Older changelogs have the headline added second time, resulting in duplicate sections


zvolin:
## Bug description

Hey, we've noticed recently that there has been some change to the changelog format, that causes release-plz to add the headline and first paragraph again.

We've been using release-plz for quite a time already and it managed our changelogs entirely. There was some time with no new releases for some of the crates, and we found out running release-plz now resulted in diffs like [this](https://github.com/eigerco/blockstore/pull/36/commits/b17dd1b6bd7c42909f73fd8a0c4a5362e52a7efb):

```diff
 # Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.8.0](https://github.com/eigerco/blockstore/compare/v0.7.1...v0.8.0) - 2025-09-09
+
+### Added
+
+- [**breaking**] upgrade dependencies, update ci ([#35](https://github.com/eigerco/blockstore/pull/35))
+# Changelog
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```

I suspect it is because the new line after the `# Changelog` headline, that was missing in the past

- Would you like to work on a fix? [if I have spare time, but for now just reporting for awareness]

## To Reproduce

Most likely running an old version of release plz and then the newest, but I haven't tracked which actual versions need to be used. You can see those reproductions:
- https://github.com/eigerco/blockstore/pull/36/commits/b17dd1b6bd7c42909f73fd8a0c4a5362e52a7efb
  https://github.com/eigerco/blockstore/actions/runs/17553156574/job/49931944443#step:4:252
- https://github.com/eigerco/leopard-codec/pull/11/commits/036b05e1374e7b06f41f056bc19240be33a16bbc
  https://github.com/eigerco/leopard-codec/actions/runs/17553627455/job/49939113905#step:4:254

## Expected behavior

Inserting just the new line, or ignoring that it is missing. Adding notes about new release at the correct place of changelog.

## Environment

- OS: ubuntu-latest runner (Current runner version: '2.328.0')
- release-plz version: 0.5 action, release-plz 0.3.143

PR #2391, opened 2025-09-09
make changelog parsing more flexible

marcoieni:
fix https://github.com/release-plz/release-plz/issues/2390

### DIFF
--- a/crates/release_plz_core/src/changelog.rs
+++ b/crates/release_plz_core/src/changelog.rs
@@ -71,8 +71,16 @@ impl Changelog<'_> {
             return Ok(old_changelog);
         }
         let old_header = changelog_parser::parse_header(&old_changelog);
-        let config = self.changelog_config(old_header);
+        let config = self.changelog_config(old_header.clone());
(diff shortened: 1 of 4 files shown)


crates/release_plz_core/src/changelog.rs
```diff
@@ -103,6 +111,33 @@ impl Changelog<'_> {
...
+    } else {
+        header.to_string()
+    };
+    // Remove the header to get the changelog body
+    let generated_body = generated
+        .strip_prefix(&header_to_strip)
+        .unwrap_or(generated.as_str());
+    let old_body = old_changelog.strip_prefix(header).unwrap_or(old_changelog);
+    Ok(format!("{header}{generated_body}{old_body}"))
+}
```
marcoieni, line 139: AI generated this. It worked but I wonder if it makes sense 😅
Maybe there's a simpler approach.

marcoieni: I guess it's fine 😅

PR #2417, opened 2025-09-14
fix: set actions/checkout persist-credentials to false in init

risu729:


marcoieni:
I wonder if in the init command we should ask the user if they want to sign tags, and based on that write token or persist credentials false.

Probably we should put a (recommended) on the option to not sign them.

Wdyt?

risu729:
I think that's good, but I'm not too sure how to configure the gpg key on CI.
Should we support it too? Or just ask for the release-plz config and let the users configure the gpg key?

marcoieni:
Let's just provide the release-plz config and let users know they should configure tag signing by themselves for now.

risu729:
I'm sorry for taking a long time. I forgot about this.
Could you take a look when you have a time?

risu729:
It seems zizmor doesn't support YAML anchors. https://github.com/zizmorcore/zizmor/issues/1176

Issue #2159, opened 2025-04-02
Support for Multiple Active Release Branches with Different Version Streams


c12i:
## Motivations

I'm exploring how to use `release-plz` in our repository which maintains multiple concurrent release branches:

- `main`: Dev release branch (currently supporting `v0.5.x-dev.x`)
- `main-0.4`: Patch and backport branch for `v0.4`
- `main-0.3`: Patch and backport branch for `v0.3`

- Would you like to implement this feature? [y/n]

## Solution

Ideally, release-plz would support creating and managing release PRs across multiple active branches with different version streams.

## Alternatives

[`release-please`](https://github.com/googleapis/release-please-action?tab=readme-ov-file#action-inputs) supports setting a target branch

marcoieni:
how would you solve this problem with release-please using target branch?

tgross35:
I ran into a couple of issues attempting to do this, documented better at https://github.com/release-plz/release-plz/issues/2163. I think release-plz probably needs to figure out what the previous release in a series is. As-is it seems like the latest version on crates.io is picked, but I mentioned at https://github.com/release-plz/release-plz/issues/2163#issuecomment-2777242235 that maybe `release-plz release-pr` could do something like:

1. Look at the current Cargo.toml version
2. Ensure that version has been published
3. Make the next version a bump from that version, e.g. `0.2.0 -> 0.2.1`, `1.0.0 -> 1.0.1`, or `0.3.0-alpha.1 -> 0.3.0-alpha.2`
4. Error if that version already exists on crates.io

Another option could be to make this customizable in release-plz.toml, e.g.:

```toml
[package.foo]
series = "0.2.{minor}`

[package.bar]
series = "1.{minor}.{patch}"

[package.baz]
series = "0.3.0-alpha.{minor}"
```

marcoieni:
> 3\. Make the next version a bump from that version, e.g. `0.2.0 -> 0.2.1`, `1.0.0 -> 1.0.1`, or `0.3.0-alpha.1 -> 0.3.0-alpha.2`

So the next version is always a patch bump? What about `0.2.0 -> 0.3.0` because for example cargo semver checks detected a breaking change?

I was thinking that maybe if your Cargo.toml specifies 0.2.0 and crates.io already contains 0.3.0, then Release-plz automatically understands that you already published 0.3.0 and so it bumps to 0.2.1?
But it's unclear on which branch to open this PR.

[...]

PR #2438, opened 2025-09-30
feat: release-plz can now publish backport PRs

LukeMathWalker:
Previously, it would error out when the registry contained a newer version than the local one. 

Related to: https://github.com/release-plz/release-plz/issues/2159

### DIFF
--- a/crates/release_plz_core/src/command/update/updater.rs
+++ b/crates/release_plz_core/src/command/update/updater.rs
@@ -616,9 +616,9 @@ impl Updater<'_> {
                     // as part of the release.
                     // We can process the next package.
                     break;
-                } else if registry_package.package.version < package.version {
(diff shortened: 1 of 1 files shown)


marcoieni:
The test fails for me locally:

```
Caused by:
    0: failed to determine next versions
    1: failed to retrieve difference of package eqfdgnur
    2: package `eqfdgnur` has a different version (0.1.0) with respect to the registry package (0.2.0), but the git tag v0.1.0 exists. Consider running `cargo publish` manually to publish the new version of this package.


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    release_pr::release_plz_can_do_backport_prs
```

marcoieni:
hopefully this fix is correct. No idea why CI and local tests behaved differently and I've no time to investigate 😅

PR #2445, opened 2025-10-06
feat: Add support for postprocessors in changelog

LukeMathWalker:
Without support for post-processors, there is no way to format a changelog ahead of opening the release PR.

If the CHANGELOG.md file is in scope for formatting checks, you're then forced to clone the release PR branch, fix the formatting, edit history, force push... It's been a consistent annoyance for me in Pavex, so I finally decided to do something about it 😁


marcoieni:
> It's been a consistent annoyance for me in Pavex, so I finally decided to do something about it

my strategy to annoy users until they contribute works, then!

PR #2489, opened 2025-11-13
update the security.md

Lori-Shu:
fix issue#2488.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -41,6 +41,15 @@
       // Every weekend
       schedule: '* * * * 0,6',
     },
+    "regexManagers": [
+    {
+      "fileMatch": ["security\\.md$"],
+      "matchStrings": [
+        "uses: release-plz/action@\\s*(?<currentValue>[a-f0-9]{40})"
+      ],
+      "datasourceTemplate": "github-tags",
+    }
+  ]
   ],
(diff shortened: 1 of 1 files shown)


website/docs/github/security.md
```diff
@@ -52,13 +54,10 @@ jobs:
       - ...
       - name: Run release-plz
 # highlight-next-line
-        uses: release-plz/action@63ab0c2746bedc448370bad4b0b3d536458398b0 # v0.5.50
+        uses: renovatebot/github-action@v39.0.3
```
marcoieni: with this you run renovate instead of release-plz. You should use renovate to update the action hash in the release-plz github actions.

Lori-Shu: Do you mean I should merely change back the example? And the renovate is configured well already.

marcoieni: renovate doesn't update the hash in this md file

Lori-Shu:
Sir, is it right to alter this file, I'm just too unfamiliar with this repo. I turned to ai for advice.

marcoieni:
I would like the renovate of release-plz github actions to update the hash in the markdown file `security.md`

Lori-Shu:
I guess I start to head to the right direction.

PR #2497, opened 2025-11-20
Limit first-release commit scan and document max_analyze_commits

LindonAliu:
This continues from PR #2232 by @TroyKomodo and addresses @marcoieni ’s comments.

- Default max_analyze_commits to 1000 and apply it only when the crate has no registry release yet.
- Update CLI/help + docs and regenerate the schema.
- Keep upstream main rebased.

Closes: #2232

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -912,7 +912,7 @@ async fn release_package(
                     *trusted_publishing_client = Some(tp);
                 }
                 Err(e) => {
-                    warn!("Failed to use trusted publishing: {e}. Proceeding without it.");
+                    warn!("Failed to use trusted publishing: {e:#}. Proceeding without it.");
                 }
             }
         }
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -17534,9 +17534,9 @@
             }
         },
         "node_modules/prettier": {
-            "version": "3.7.3",
-            "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.7.3.tgz",
-            "integrity": "sha512-QgODejq9K3OzoBbuyobZlUhznP5SKwPqp+6Q6xw6o8gnhr4O85L2U915iM2IDcfF2NPXVaM9zlo9tdwipnYwzg==",
+            "version": "3.7.4",
+            "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.7.4.tgz",
+            "integrity": "sha512-v6UNi1+3hSlVvv8fSaoUbggEM5VErKmmpGA7Pl3HF8V6uKY7rvClBOJlH6yNwQtfTueNkGVpOv/mtWL9L4bgRA==",
             "license": "MIT",
             "bin": {
                 "prettier": "bin/prettier.cjs"


crates/release_plz/src/args/set_version.rs
```diff
@@ -24,6 +24,10 @@ pub struct SetVersion {
     /// Path to the release-plz config file.
     #[command(flatten)]
     pub config: ConfigPath,
+
+    /// In a workspace update all member's who reference the crate's version.
+    #[arg(long)]
+    disable_dependant_updates: bool,
```
marcoieni: this is unrelated to the PR title. Can you edit this PR to only include changes from the commit scan and max_analyze_commits?

LindonAliu: I rebased the branch on latest main and kept only the max_analyze_commits/commit-scan changes. This is building on @TroyKomodo’s earlier work; I just rebased/cleaned it up.

crates/release_plz/src/config.rs
```diff
@@ -110,7 +112,7 @@ impl Config {
 }
 
 /// Config at the `[workspace]` level.
-#[derive(Serialize, Deserialize, Default, PartialEq, Eq, Debug, JsonSchema)]
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, JsonSchema)]
```
marcoieni, line 115: why removing the derive Default from here?

crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -566,7 +566,13 @@ impl Updater<'_> {
...
         let pathbufs_to_check = pathbufs_to_check(package_path, package)?;
         let paths_to_check: Vec<&Path> = pathbufs_to_check.iter().map(|p| p.as_ref()).collect();
-        loop {
+        let max_analyze_commits = if registry_package.is_none() {
+            self.req.max_analyze_commits()
+        } else {
+            u32::MAX
+        };
+
+        for _ in 0..max_analyze_commits.max(1) {
```
marcoieni: rather than using `max(1)` here, I would return an error if `self.req.max_analyze_commits()` is negative.

We could also add that if `self.req.max_analyze_commits()` is 0, we use `u32::MAX`.

crates/release_plz/src/config.rs
LindonAliu: I removed `Default` from the derive because there's a manual `impl Default for Workspace` at line 177 that sets `max_analyze_commits` to `default_max_analyze_commits()` (which returns `Some(1000)`) instead of `None`.

If I keep `#[derive(Default)]`, it conflicts with the manual implementation, causing a compile error `E0119: conflicting implementations of trait Default`.

crates/release_plz_core/src/command/update/updater.rs
LindonAliu: The type is `u32`, so negative values are rejected at the TOML parsing level by serde. There's no way for a negative value to reach this code.

I did implement the second part of your suggestion: when `max_analyze_commits` is `0`, we now use `u32::MAX`.

crates/release_plz/src/config.rs
marcoieni: Ok, so you can't set that value without removing the derive Default, I see 👍

marcoieni (approved):

Issue #2334, opened 2025-08-12
Error while waiting for package to be published


beeb:
## Bug description

The CI run for my latest release failed with the following error:

```
-- Running release-plz release --
2025-08-12T19:40:36.877762Z  INFO using release-plz config file release-plz.toml
2025-08-12T19:44:05.281669Z  INFO waiting for the package lintspec to be published...
2025-08-12T19:58:46.872890Z ERROR failed to release package

Caused by:
    0: timeout while publishing lintspec
    1: failed to update git index
    2: "gix" crate failed. If problems persist, consider deleting `~/.cargo/registry/index/github.com-1ecc6299db9ec823/`
    3: The slotmap turned out to be too small with 32 entries, would need 1 more
Error: failed to release package
```

The package was correctly published to crates.io using the trusted publisher token I provided, but it got stuck at `waiting for the package lintspec to be published` (as you can see from the timestamps) then errored out with an error about `gix`.

I managed to work around the issue by creating the tag manually but that's not ideal and I would like to understand the cause for this failure.

- Would you like to work on a fix? n

## To Reproduce

1. https://github.com/beeb/lintspec/actions/runs/16919093254/job/47940360826

## Expected behavior

A git tag is created for the version.

## Environment

Github runner

ncatelli:
FWIW I've been running into this internally as well. It seems to have manifested in rust 1.85+ and we have not had the issue when pinning to <1.85.

ncatelli:
> FWIW I've been running into this internally as well. It seems to have manifested in rust 1.85+ and we have not had the issue when pinning to <1.85.

Additionally I think it may be bubbling up from https://github.com/frewsxcv/rust-crates-index/issues/181.

beeb:
Thanks for the info @ncatelli . Unfortunately I'm on 1.88 MSRV at the moment which makes the workaround unfeasible.

beeb:
Curious why this doesn't use the sparse index? Looks like it's trying to use the git index because of the `gix` error.

marcoieni:
I don't know what the issue is and don't have time to debug or fix, but PR welcome 👍

ncatelli:
> Curious why this doesn't use the sparse index? Looks like it's trying to use the git index because of the `gix` error.

I think that error might be a red herring as it's a static error message from crates-index https://github.com/frewsxcv/rust-crates-index/blob/master/src/error.rs#L10.

[...]

PR #2499, opened 2025-11-21
feat: retry github publishing even if crates.io publish succeeded

danieleades:
the purpose of this PR is to allow the "publish to github" step to retry even if the crates.io publish has previously succeeded.

see https://github.com/release-plz/release-plz/issues/2334#issuecomment-3557543558

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -1232,38 +1232,7 @@ async fn release_package(
             );
         }
 
-        let contributors = get_contributors(release_info, git_client).await;
-
-        // TODO fill the rest
-        let remote = Remote {
-            owner: String::new(),
-            repo: String::new(),
-            link: String::new(),
-            contributors,
-        };
-        if should_create_git_release {
-            let release_body =
-                release_body(input, release_info.package, release_info.changelog, &remote);
-            let release_config = input
-                .get_package_config(&release_info.package.name)
-                .git_release;
-            let is_pre_release = release_config.is_pre_release(&release_info.package.version);
-            let git_release_info = GitReleaseInfo {
-                git_tag: release_info.git_tag.to_string(),
-                release_name: release_info.release_name.to_string(),
-                release_body,
-                draft: release_config.draft,
-                latest: release_config.latest,
-                pre_release: is_pre_release,
-            };
-            git_client.create_release(&git_release_info).await?;
-        }
-
-        info!(
-            "published {} {}",
-            release_info.package.name, release_info.package.version
-        );
-        Ok(true)
+        Ok(should_publish)
     }
 }
 


marcoieni:
mmmh, I see tests failing in CI

danieleades:
> mmmh, I see tests failing in CI

i don't see that- only pending checks

Screenshot 2025-11-21 110738

marcoieni:
Click on the ci of the previous commits. Or push a new empty commit to trigger ci again

danieleades:
> Click on the ci of the previous commits. Or push a new empty commit to trigger ci again

Oh you meant the previous commits had failing CI? well, yeah. Hence the subsequent commits.

You can see the CI passing in the equivalent PR in my fork

https://github.com/danieleades/release-plz/pull/1

danieleades:
@marcoieni i believe this is ready for review

crates/release_plz_core/src/git/forge.rs
```diff
@@ -927,14 +962,27 @@ impl GitClient {
...
         if response.status() == StatusCode::UNPROCESSABLE_ENTITY {
-            // Try to capture the body for extra diagnostics.
+            // Capture the body to determine which case this is
             let body = response
                 .text()
                 .await
                 .unwrap_or_else(|_| "<failed to read response body>".to_string());
+
+            // Check if this is a "ref already exists" error (idempotent case)
+            if body.contains("Reference already exists") || body.contains("already exists") {
```
GideonBear:
```suggestion
            if body.contains("already exists") {
```
This does nothing; `"already exists"` is a substring of `"Reference already exists"`, so the first check can only succeed if the second succeeds.

crates/release_plz_core/src/command/release.rs
```diff
@@ -961,54 +1119,6 @@ async fn release_package(
             wait_until_published(index, release_info.package, input.publish_timeout, token).await?;
         }
 
-        if should_create_git_tag {
```
GideonBear, line 964: The `should_create_git_tag` and `should_create_git_release` variables are now unused (I think) in this function, only debug-logged; that debug log should probably moved to the function where these variables are actually used, and the duplicate definition removed.

crates/release_plz_core/src/command/release.rs
```diff
@@ -874,13 +882,164 @@ struct ReleaseInfo<'a> {
...
+/// This function is idempotent - it checks if each resource exists before creating.
+/// Returns `true` if any git operation was performed, `false` if everything already existed.
+async fn create_git_tag_and_release(
+    input: &ReleaseRequest,
+    repo: &Repo,
+    git_client: &GitClient,
+    release_info: &ReleaseInfo<'_>,
+) -> anyhow::Result<bool> {
+    if input.dry_run {
+        return Ok(false);
```
GideonBear, line 951: I don't know exactly how dry runs work in release-plz, should this be `true`?

crates/release_plz_core/src/command/release.rs
```diff
@@ -874,13 +882,164 @@ struct ReleaseInfo<'a> {
...
+    if should_create_git_tag {
+        // Check if tag exists locally
+        let tag_exists_locally = repo.tag_exists(release_info.git_tag)?;
+
+        if tag_exists_locally {
+            info!(
+                "skipping creation of git tag {}: already exists locally",
+                release_info.git_tag
+            );
+        } else {
```
GideonBear: This (combined with other stuff you're doing below) is creating ungodly layers of nesting; can this be extracted to another function? Then you can use the guard pattern.

crates/release_plz_core/src/command/release.rs
```diff
@@ -961,54 +1119,6 @@ async fn release_package(
             wait_until_published(index, release_info.package, input.publish_timeout, token).await?;
         }
 
-        if should_create_git_tag {
-            // Use same tag message of cargo-release
```
GideonBear, line 965: You didn't move this comment

crates/release_plz_core/src/command/release.rs
```diff
@@ -961,54 +1119,6 @@ async fn release_package(
...
-            let message = format!(
-                "chore: Release package {} version {}",
-                release_info.package.name, release_info.package.version
-            );
-            let should_sign_tags = repo
-                .git(&["config", "--default", "false", "--get", "tag.gpgSign"])
-                .map(|s| s.trim() == "true")?;
-            // If tag signing is enabled, create the tag locally instead of using the API
```
GideonBear, line 973: You didn't move this comment

crates/release_plz_core/src/command/release.rs
```diff
@@ -961,54 +1119,6 @@ async fn release_package(
...
-                let sha = repo.current_commit_hash()?;
-                git_client
-                    .create_tag(release_info.git_tag, &message, &sha)
-                    .await?;
-            }
-        }
-
-        let contributors = get_contributors(release_info, git_client).await;
-
-        // TODO fill the rest
```
GideonBear, line 987: You didn't move this comment

GideonBear (commented):
This is about as far as I can get without knowing the codebase. Looks good, I like the idempotent design!

crates/release_plz_core/src/command/release.rs
GideonBear: Resolved

crates/release_plz_core/src/command/release.rs
GideonBear: Resolved

crates/release_plz_core/src/command/release.rs
GideonBear: resolved

crates/release_plz_core/src/git/forge.rs
GideonBear: resolved

crates/release_plz_core/src/command/release.rs
GideonBear: resolved

crates/release_plz_core/src/command/release.rs
GideonBear: Needs a maintainer

crates/release_plz_core/src/command/release.rs
GideonBear: Still applicable, I would personally split this up into multiple functions (`create_git_tag` and `create_git_release` for example), and use the guard pattern within those functions. But that's a matter of style.

danieleades: I hope to find some time in the next couple of days to give this a proper scrub

danieleades: could you take another look please @GideonBear ?

GideonBear: Looks much cleaner!

danieleades:
the `release.rs` module is doing a *lot*. Outside the scope of this PR, i think, but it would be good to break that down into submodules that handle the registry publishing and the github/tea publishing separately.

danieleades:
ping @marcoieni

crates/release_plz_core/src/command/release.rs
```diff
@@ -651,6 +644,69 @@ async fn release_package_if_needed(
...
+    // OPERATION ORDERING RATIONALE:
+    // We perform operations in order of reversibility, from most-reversible to least-reversible.
+    // This ensures failures leave the system in a recoverable state.
+    //
+    // Reversibility hierarchy (most to least):
+    //   1. Git tags (highly reversible: can be deleted/recreated)
+    //   2. Git releases (highly reversible: can be deleted)
+    //   3. Registry publishing (irreversible: crates.io doesn't allow unpublishing)
+    //
+    // Therefore: Git ops FIRST, registry publish SECOND
```
marcoieni, line 657: I don't like this.
Git releases should be done after registry publishing in my opinion because:
- registry publishing could fail
- people could have automation that relies on the fact that when the git release is created, the package is already published

crates/release_plz_core/src/command/release.rs
```diff
@@ -631,13 +631,6 @@ async fn release_package_if_needed(
 ) -> anyhow::Result<Option<PackageRelease>> {
     let git_tag = project.git_tag(&package.name, &package.version.to_string())?;
     let release_name = project.release_name(&package.name, &package.version.to_string())?;
-    if repo.tag_exists(&git_tag)? {
-        info!(
-            "{} {}: Already published - Tag {} already exists",
-            package.name, package.version, &git_tag
-        );
-        return Ok(None);
-    }
```
marcoieni, line 640: if we publish to the registry first, we can keep this check and skip checking if the package is already published in the registry.
We could of course still check if the release was created.

crates/release_plz_core/src/command/release.rs
```diff
@@ -949,104 +1216,37 @@ async fn release_package(
...
-            };
-            git_client.create_release(&git_release_info).await?;
         }
 
-        info!(
-            "published {} {}",
-            release_info.package.name, release_info.package.version
-        );
-        Ok(true)
+        Ok(should_publish)
```
marcoieni, line 1235: please split all refactorings in a separate PR before changing the behavior, so that the diff is easier to review

PR #2512, opened 2025-11-27
feat: add `git_only` config option to use git tags to determine the latest version

### LINKED ISSUE #1144: Add a "git only" mode

jdeinum:
Added the functionality to use git tags as the source of truth for package releases. Aims to close  #1144.
 
I just jammed all of the commits together. There isn't a crazy amount of code and my hope is its easy enough to follow along that I don't need to split it. If you prefer them in pieces, I could maybe do (1) config (2) all other src (3) tests ?  

Happy to make changes as required, just wanted to get this out!

### DIFF
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -108,7 +108,8 @@ fn process_git_only_package(
         .get_package_tag_name(&package.name)
         .unwrap_or_else(|| default_tag_name_template(is_multi_package));
 
-    let release_regex = release_regex::get_release_regex(&template, &package.name).context("get release regex")?;
+    let release_regex =
+        release_regex::get_release_regex(&template, &package.name).context("get release regex")?;
     debug!(
         "looking for tags matching pattern: {}",
         release_regex.to_string()


jdeinum:
I have more tests failing on my own machine than CI, so clearly there are some differences. I think I did introduce a regression so I am gonna look at that during lunch.

marcoieni:
I also once had failing tests locally while they were passing in CI, not sure why.

.schema/latest.json
```diff
@@ -293,6 +296,30 @@
             "null"
           ]
         },
+        "git_only": {
+          "title": "Git Only",
+          "description": "If true, use git tags to determine the latest version instead of the registry.\nIf unspecified at package level, inherits from workspace config.",
```
marcoieni: "if unspecified.." that's true for all the configurations, so I wonder if it makes sense to specify it.

Maybe we can delete that sentence

jdeinum:
There was a period where there were failing tests on both upstream and locally, but they seem resolved now. 

I did run into a problem with dropping temp dirs too early, but my last commit looks like it took care of that.

crates/git_cmd/src/lib.rs
```diff
@@ -37,6 +37,27 @@ impl Repo {
...
+                path.as_ref()
+            );
+            return Ok(());
+        }
+
+        // create the repo
+        git_in_dir(path.as_ref(), &["init"]).context("init git directory")?;
+        debug!("initialized repo at {}", path.as_ref());
+        Ok(())
+    }
```
marcoieni: this function is unused

crates/git_cmd/src/lib.rs
```diff
@@ -216,9 +237,20 @@ impl Repo {
...
+        path: impl AsRef<Utf8Path>,
+        object: Option<&str>,
+    ) -> anyhow::Result<()> {
+        if let Some(commit) = object {
+            self.git(&["worktree", "add", "-f", path.as_ref().as_str(), commit])
+                .context(format!("create git worktree at commit {commit}"))?;
+        } else {
+            self.git(&["worktree", "add", "-f", path.as_ref().as_str()])
+                .context("create git worktree at HEAD")?;
+        }
```
marcoieni: this function is unused, why editing it?

crates/git_cmd/src/lib.rs
```diff
@@ -231,6 +263,38 @@ impl Repo {
...
+        // now we split each line into their components
+        let split_lines: Vec<(String, String)> = lines
+            .iter()
+            .filter(|line| line.contains(" ")) // if something doesn't have a space, we just filter
+            .map(|x| x.split_at(x.find(" ").unwrap())) // unwrap should be safe because of ^
+            .map(|(y, z)| (y.to_string(), z.to_string()))
+            .collect();
+
+        Ok(split_lines)
+    }
```
marcoieni: this function is unused

crates/release_plz/src/config.rs
```diff
@@ -167,6 +219,24 @@ pub struct Workspace {
...
+    /// Default: false
+    ///
+    /// If true, release-plz will use git tags to determine what the latest version of the package
+    /// is (i.e newest version is v0.1.3 and is associated with commit ac83762)
+    /// If false, release-plz will use crates.io release information to get the latest version
+    pub git_only: Option<bool>,
+
+    /// Literal string prefix for release tags when `git_only` is enabled.
+    /// Optional. For example: "v" matches "v1.2.3", "package-v" matches "package-v1.2.3".
+    /// Default: empty string (tags like "1.2.3")
```
marcoieni: doesn't release-plz creates `v1.2.3` tags? Why the default should be `1.2.3`?

crates/release_plz/src/args/release_pr.rs
```diff
@@ -15,6 +15,8 @@ pub struct ReleasePr {
 }
 
 impl ReleasePr {
+    // combines the user provided configuration and package / workspace metadata into an internal
+    // representation that can be passed around internally.
```
marcoieni:
```suggestion
    /// combines the user provided configuration and package / workspace metadata into an internal
    /// representation that can be passed around internally.
```

crates/release_plz/src/args/update.rs
```diff
@@ -176,15 +176,25 @@ impl Update {
         config: &Config,
         cargo_metadata: cargo_metadata::Metadata,
     ) -> anyhow::Result<UpdateRequest> {
+        // if we provided a manifest, use that instead of the current dir
         let project_manifest = self.manifest_path();
+
+        // Check lock file
```
marcoieni:
```suggestion
```
the function name gives more info than the comment

crates/release_plz/src/args/update.rs
```diff
@@ -176,15 +176,25 @@ impl Update {
...
         check_if_cargo_lock_is_ignored_and_committed(&project_manifest)?;
+
+        // create the update request
```
marcoieni:
```suggestion
```
`UpdateRequest::new` already explains this

crates/release_plz/src/args/update.rs
```diff
@@ -176,15 +176,25 @@ impl Update {
...
                 format!("Cannot find file {project_manifest:?}. Make sure you are inside a rust project or that --manifest-path points to a valid Cargo.toml file.")
             })?
             .with_dependencies_update(self.dependencies_update(config))
             .with_max_analyze_commits(self.max_analyze_commits(config))
             .with_allow_dirty(self.allow_dirty(config));
+
+        // the repository url is used for links within the changelogs to the release
+        // normally you don't need to populate it because release-plz will default to the remote
+        // url inside of the Cargo.toml , but users can optionally provide it in the config file /
+        // on the command line
```
marcoieni:
```suggestion
        // The repository url is used for links within the changelogs to the release.
        // Normally you don't need to populate it because release-plz will default to the remote
        // url, but users can optionally provide it in the config file or
        // in the command line argument.
```
Actually we get this information from git, not from the Cargo.toml (iirc).

crates/release_plz/src/args/update.rs
```diff
@@ -203,8 +215,12 @@ impl Update {
                     format!("cannot find project manifest {registry_manifest_path:?}")
                 })?;
         }
-        update = config.fill_update_config(self.no_changelog, update);
+
+        // override workspace configuration with package specific configuration
+        update = config.fill_update_config(self.no_changelog, update)?;
+
         {
+            // update the release date
```
marcoieni:
```suggestion
```
the code says `let release_date =` so it's clear what we are doing

crates/release_plz/src/args/update.rs
```diff
@@ -213,13 +229,18 @@ impl Update {
                         .context("cannot parse release_date to y-m-d format")
                 })
                 .transpose()?;
+
+            // create the pr link
```
marcoieni:
```suggestion
```
`let pr_link =` is already explicit

crates/release_plz/src/main.rs
```diff
@@ -35,15 +35,34 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
...
+
+        // At a high level, the release PR command does the following:
+        //
+        // 1. Get the package / workspace you are looking to update
+        // 2. Get the most recent version of that package according to *some source*
+        // 3. Get the list of commits since the most recent version of that package
+        // 4. Determine the next version number based on those commits
+        // 5. If the next version a.b.c is different from the current version x.y.z (i.e there are
+        //    changes), then we open a new PR for updating to a.b.c. If release-plz already has a
+        //    PR open for updating to an older version, then we replace that PR with a new one
```
marcoieni:
```suggestion
        //    PR open for updating to an older version, then we update that PR
```
We only close and reopen the PR if it contains commits from non-bots

crates/release_plz/src/main.rs
```diff
@@ -35,15 +35,34 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
...
                 "please provide the git token with the --git-token cli argument."
             );
+
+            // The cargo metadata contains information on package versions
             let cargo_metadata = cmd_args.update.cargo_metadata()?;
+
             let config = cmd_args.update.config.load()?;
+
+            // Combine the config + cargo metadata into an internal representation of a release
+            // request that we can pass around
```
marcoieni:
```suggestion
```
this comment is already present in the function definition

crates/release_plz/src/main.rs
```diff
@@ -35,15 +35,34 @@ async fn run(args: CliArgs) -> anyhow::Result<()> {
...
             let request = cmd_args.release_pr_req(&config, cargo_metadata)?;
+
+            // release pr
```
marcoieni:
```suggestion
```
this comment just repeats the name of the variable

crates/release_plz/tests/all/git_only.rs
```diff
@@ -0,0 +1,753 @@
...
+
+    // Make a fix commit
+    let readme = context.repo_dir().join("README.md");
+    fs_err::write(&readme, "# Updated README").unwrap();
+    context.push_all_changes("fix: update readme");
+
+    // Run release-pr
+    let outcome = context.run_release_pr().success();
+
+    // Verify no PR was created
```
marcoieni: If you repeat the same test with the cargo registry, a PR will be created to update the changelog iirc.
Why in this case you don't create the PR?
The behaviors of cargo registry and git tags should be the same.

crates/release_plz/tests/all/git_only.rs
```diff
@@ -0,0 +1,753 @@
...
+    fs_err::write(&readme, "# Updated README").unwrap();
+    context.push_all_changes("fix: update readme");
+
+    // Run release-pr should fail due to validation error
+    let error = context.run_release_pr().failure().to_string();
+    assert!(
+        error.contains("git_only")
+            && error.contains("publish")
+            && error.contains("mutually exclusive"),
+        "Expected validation error about git_only and publish being mutually exclusive, got: {error}"
```
marcoieni, line 665: are you sure `{error}` is expanded in the variable content?

crates/release_plz/tests/all/git_only.rs
marcoieni, line 1: are you testing the `run_release()` function to test that the `release` command creates the tags as expected and asserting that it doesn't publish to the cargo registry?

crates/release_plz_core/src/next_ver.rs
```diff
@@ -56,10 +61,118 @@ impl ReleaseMetadataBuilder for UpdateRequest {
     }
 }
 
+// Build regex: ^{escaped_prefix}(\d+\.\d+\.\d+){escaped_suffix}$
+// The semantic version is captured in group 1
```
marcoieni: in Rust, function comments are with `///`, please change this for all functions

crates/release_plz_core/src/command/release_pr/git.rs
```diff
@@ -0,0 +1,348 @@
...
+use regex::Regex;
+use std::path::Path;
+use tempfile::TempDir;
+use tracing::{debug, error, info, instrument, warn};
+
+use crate::fs_utils::to_utf8_path;
+
+pub struct CustomRepo {
+    repo: Repository,
+}
```
marcoieni, line 13: what's the reason for creating a new representation of the repository, different from `git_cmd::Repo`?
Probably you already wrote this somewhere.
Is there a plan to unify the two? Or they should be different? It could be confusing to have two separate repository representations.

If there's a plan to unify the two, can you open an issue after this pr is merged?

crates/release_plz_core/src/command/release_pr/git.rs
```diff
@@ -0,0 +1,348 @@
+use anyhow::{Context, Result, bail};
```
marcoieni:
```suggestion
use anyhow::{Context, bail};
```
please don't import `anyhow::Result`, I don't like it.

crates/release_plz_core/src/command/update/mod.rs
```diff
@@ -5,7 +5,8 @@ mod update_config;
 pub mod update_request;
 pub mod updater;
 
-use crate::{PackagePath, tmp_repo::TempRepo};
+use crate::PackagePath;
+use crate::tmp_repo::TempRepo;
```
marcoieni: ??
Is this a rustfmt rule? Why making this change?

marcoieni:
> If you prefer them in pieces, I could maybe do (1) config (2) all other src (3) tests ?

that's not how I like to split PRs, Ideally PRs should contain both src and tests for a single functionality. I.e. PRs should be self-contained.
So for example you do a first PR to refactor and add comments, a second one to introduce git_only feature, another to introduce the prefix functionality, etc.
Anyway, I can review this PR as it is, no need to split 👍

jdeinum:
Will go through all of these and make everything consistent once I am back from Europe! Thanks for the feedback.

jdeinum:
Back now, will start on it today.

crates/release_plz/src/config.rs
jdeinum: Fair point, will default to "v"

crates/release_plz_core/src/command/release_pr/git.rs
jdeinum: I made the wrapper to hide implementation from the callers, only expose the functionality that we needed (which was built on the original repo), and add some logging so we didn't need to put those throughout the codebase. TempWorktree does have a unique drop impl to clean up after its done, so that one does more of a purpose. 

I am not particularly attached to either approach, if you think its better to have a single struct I can just move the rest of the functionality into functions that take a Repository and give us back our custom resources.

crates/release_plz_core/src/command/update/mod.rs
jdeinum: Likely just zooming through imports trying to clean, will fix.

crates/release_plz/tests/all/git_only.rs
jdeinum: I added one that validates that it throws an error if both `git_only` and `publish` are true for a package. I think that covers that use case?

jdeinum:
I think this should be a bit cleaner overall. Let me know if you have anything else you'd like changed.

crates/release_plz/tests/all/git_only.rs
jdeinum: Hmmm. I did not think of this. I skipped updating packages for any package without a matching git release regex because I assume we wouldn't want a PR. I will go back and make it consistent.

crates/release_plz/tests/all/git_only.rs
jdeinum: Clippy complains here if you don't, so I imagine so?

crates/release_plz/tests/all/git_only.rs
```diff
@@ -0,0 +1,791 @@
...
+[workspace]
+git_only = true
+
+[[package]]
+name = "lib1"
+git_only_release_tag_prefix = "lib1-v"
+
+[[package]]
+name = "lib2"
+git_only_release_tag_prefix = "lib2-v"
```
marcoieni: can we make this the default? So that people don't need to specify it for every package?
I.e. `git_only_release_tag_prefix` follows the same defaults of the tag names created by release-plz. See https://release-plz.dev/docs/config#the-git_tag_name-field

marcoieni:
Can you update the docs in https://github.com/release-plz/release-plz/blob/main/website/docs/config.md ?

crates/release_plz/tests/all/git_only.rs
marcoieni: shouldn't we also test that `git_only` is true, `publish` is false, and release-plz creates the tags as expected?

crates/release_plz_core/src/command/release_pr/mod.rs
```diff
@@ -145,15 +148,18 @@ pub async fn release_pr(input: &ReleasePrRequest) -> anyhow::Result<Option<Relea
         .update_request
         .git_client()?
         .context("can't find git client")?;
+
+    let unreleased_package_worktree_repo =
+        Repo::new(&tmp_project_root).context("create new repo")?;
```
marcoieni: why you moved this outside of the if?
Since this variable is only used inside the if.

crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -549,6 +554,17 @@ impl Updater<'_> {
             tag_commit.as_deref(),
             &mut diff,
         )?;
+
+        // Clean up cargo package artifacts created during diff calculation.
+        // `cargo package --list` creates files in target/package/ which can interfere with checkout.
+        // The target directory is created in the package directory, not the repository root.
+        let target_package_dir = package_path.join("target").join("package");
```
marcoieni: what if `CARGO_TARGET_DIR` is set? The target dir is not in that directory, right?

Also, I tried:
- `unset CARGO_TARGET_DIR`
- `cargo clean`
- `cargo package --list`

The target directory is not even created.
And even after running `cargo build`, the `target/package` directory doesn't exist for me. I'm using rust 1.92

crates/release_plz/src/config.rs
```diff
@@ -167,6 +219,24 @@ pub struct Workspace {
...
+    ///
+    /// If true, release-plz will use git tags to determine what the latest version of the package
+    /// is (i.e newest version is v0.1.3 and is associated with commit ac83762)
+    /// If false, release-plz will use crates.io release information to get the latest version
+    pub git_only: Option<bool>,
+
+    /// Literal string prefix for release tags when `git_only` is enabled.
+    /// Optional. For example: "v" matches "v1.2.3", "package-v" matches "package-v1.2.3".
+    /// Default: "v" (tags like "v1.2.3")
+    pub git_only_release_tag_prefix: Option<String>,
```
marcoieni: I was thinking..Instead of having two config options (one for the prefix and one for the suffix), what do you think about having only one that is a template?

I.e. `git_only_release_tag`.

And you can use the same config variables of https://release-plz.dev/docs/config#the-git_tag_name-field (`package` and `version`).

I.e. it could be `"{{ package }}-v{{ version }}"` so that you can configure it at the workspace level instead of specifying the package name for every crate of your workspace like you did in the tests.

crates/release_plz/tests/all/git_only.rs
jdeinum: Sure, I think that should be fine.

crates/release_plz_core/src/command/release_pr/mod.rs
jdeinum: I think in the original version I just replaced the existing call to create the temp directory with this one, but you are correct, I'll move that down.

crates/release_plz/src/config.rs
jdeinum: Yes that would be far nicer. I'll do that.

crates/release_plz/tests/all/git_only.rs
jdeinum: Will not do this if doing the single `tag` so marking as resolved

crates/release_plz_core/src/command/update/updater.rs
jdeinum: I hadn't even considered `CARGO_TARGET_DIR`, oops.

I could have sworn cargo package --list was generating a `target/package` for me a little while ago... but upon checking now I also don't see it. I just removed the cleanup code for that then.

crates/release_plz/tests/all/git_only.rs
jdeinum: Added.

marcoieni:
- [x] what do you think about reusing `git_tag_name` as `git_only_release_tag_name` instead of introducing the new config option `git_only_release_tag_name`? 🤔

In theory these two config fields should always have the same value, right?

marcoieni:
> * [x]  what do you think about reusing `git_tag_name` as `git_only_release_tag_name` instead of introducing the new config option `git_only_release_tag_name`? 🤔
> 
> In theory these two config fields should always have the same value, right?

I made this change in the latest commit. let me know what you think 👍

crates/release_plz_core/src/command/update/update_request.rs
```diff
@@ -47,6 +47,10 @@ pub struct UpdateRequest {
     /// Prepare release only if at least one commit respects a regex.
     release_commits: Option<Regex>,
     git: Option<GitForge>,
+
+    /// Use git tags to determine latest package release
+    git_only: Option<bool>,
```
marcoieni: this struct contains this field twice. Both here and in `packages_config.default` 🤔
This doesn't seem right. Can you fix it?

website/docs/config.md
```diff
@@ -332,6 +334,35 @@ Where:
 - `{{ package }}` is the name of the package.
 - `{{ version }}` is the new version of the package.
 
+#### The `git_only` field
+
+Enable git-only mode, which determines package versions from git tags instead of the cargo registry.
+
+- If `true`, release-plz will look for existing release tags in git to determine the current
```
marcoieni:
```suggestion
- If `true`, release-plz will look for existing git tags to determine the current
```
Let's not introduce the concept of "release tags" since they are just git tags and there's no need to introduce a new name for them, right?

crates/release_plz/tests/all/git_only.rs
```diff
@@ -0,0 +1,823 @@
...
+This PR was generated with [release-plz](https://github.com/release-plz/release-plz/)."
+        )
+        .trim(),
+        pr_body.trim()
+    );
+}
+
+#[tokio::test]
+#[cfg_attr(not(feature = "docker-tests"), ignore)]
+async fn git_only_with_lightweight_tags() {
```
marcoieni, line 427: this test runs infinitely from me in my laptop. I think the reason is that it waits for stdin to write the tag message.

Because when I run `git tag tag_name` in the terminal, it opens vim waiting for me to write the tag message.

Why this doesn't happen in CI? Do we need to set the tag message to avoid this? Is it due to my git config?
Is there something we can do to avoid this from happening in tests?

jdeinum:
I like it! The simpler the better :)

May have a chance for those fixes later today, otherwise I'll get to it tomorrow.

crates/release_plz/tests/all/git_only.rs
jdeinum: It doesn't happen for me either, I'm using `nightly-x86_64-unknown-linux-gnu (default)` , but my guess is its a platform issue? I'm running on Arch `6.18.2-arch2-1`. Are you running on windows or mac by chance? There is a possibility I overlooked some platform behavior.

jdeinum: Wait I see the docker tests for mac and windows too. Could you please share your git config? Mine uses primarily defaults for most things other than diff display and credential management stuff.

jdeinum: Based on what I've read in the docs, I think the issue might be that when creating a lightweight tag when signing is enabled, it actually creates an annotated tag rather than a lightweight one, and prompts you to enter either a message (because `-m` isn't present), or for your gpg key. Try turning off signed commits and see if that does it.

crates/release_plz_core/src/command/update/update_request.rs
marcoieni: I feel like the `git_only` field should be in `packages_config` because it is a field that can be customized per package, right? Instead the other fields of this struct (eg `release_commits` or `allow_dirty`) are for project wide configurations.

marcoieni: btw it's still in packages_config.default, so my previous comment wasn't resolved, right? 🤔

crates/release_plz/tests/all/git_only.rs
marcoieni: I'm using mac. You are right. If I run `git config tag.gpgsign false` in the repository, it doesn't open the text editor 👍

https://github.com/release-plz/release-plz/pull/2561 should fix this issue.

So now you can restore lightweight tags in tests 👍

crates/release_plz/tests/all/helpers/test_context.rs
```diff
@@ -278,6 +278,8 @@ fn configure_repo(repo_dir: &Utf8Path, gitea: &GiteaContext) -> Repo {
     // set email
     repo.git(&["config", "user.email", &gitea.user.email()])
         .unwrap();
+    // disable GPG signing for tests
+    repo.git(&["config", "commit.gpgsign", "false"]).unwrap();
```
marcoieni: oh, I didn't see this. You can remove it as I added it in https://github.com/release-plz/release-plz/pull/2561
```suggestion
```

jdeinum: Will remove!

crates/release_plz_core/src/command/update/update_request.rs
jdeinum: Will fix this, my apologies.

jdeinum:
Will get to this tomorrow morning :)

jdeinum:
Or tomorrow morning 😬

marcoieni:
I suggest to update the branch with main 👍

marcoieni:
In https://github.com/release-plz/release-plz/pull/2512/changes/59624fd75776567017c935ae7003f7e3b60cd6c9 I simplified the `get_tag_commit` function. Apparently `find_reference` works both for annotated and lightweight tags.

crates/release_plz_core/src/next_ver.rs
```diff
@@ -56,9 +61,134 @@ impl ReleaseMetadataBuilder for UpdateRequest {
...
+    // using that path
+    let repo = GitRepo::open(worktree.path()).context("open repo for package")?;
+
+    Ok((repo, worktree))
+}
+
+/// run cargo publish within a worktree
+fn run_cargo_publish(worktree: &GitWorkTree) -> anyhow::Result<()> {
+    // run cargo package so we get the proper format
+    let output = std::process::Command::new("cargo")
```
marcoieni: please reuse the function
https://github.com/jdeinum/release-plz/blob/72b404c54ad7a84635705e5986aa9ed7c7bc6876/crates/release_plz_core/src/cargo.rs#L35

crates/release_plz_core/src/next_ver.rs
```diff
@@ -56,9 +61,134 @@ impl ReleaseMetadataBuilder for UpdateRequest {
...
+    let rust_package = MetadataCommand::new()
+        .manifest_path(format!(
+            "{}/Cargo.toml",
+            worktree
+                .path()
+                .to_str()
+                .context("convert worktree path to str")?
+        ))
+        .exec()
+        .context("get cargo metadata")?;
```
marcoieni: why not reusing
https://github.com/jdeinum/release-plz/blob/72b404c54ad7a84635705e5986aa9ed7c7bc6876/crates/cargo_utils/src/workspace_members.rs#L5

crates/release_plz_core/src/next_ver.rs
```diff
@@ -56,9 +61,134 @@ impl ReleaseMetadataBuilder for UpdateRequest {
...
+
+    let package_details = rust_package
+        .packages
+        .iter()
+        .find(|x| x.name == package_name)
+        .ok_or(anyhow::anyhow!("Failed to find package {package_name:?}"))?;
+
+    let new_path = format!(
+        "{}/target/package/{}-{}",
```
marcoieni: you are assuming people don't set the `CARGO_TARGET_DIR` environment variable, right? 🤔

marcoieni:
I only have one file left to review: `crates/release_plz_core/src/next_ver.rs` :)
Hopefully I'll get to that soon.

crates/release_plz_core/src/next_ver.rs
jdeinum: After reading through the docs, `Metadata::target_directory` handles this for us so I'll switch to using that :)

crates/release_plz_core/src/next_ver.rs
```diff
@@ -56,9 +63,119 @@ impl ReleaseMetadataBuilder for UpdateRequest {
     }
 }
 
+/// Build a regex from a Tera template for matching release tags.
+/// The template supports `{{ package }}` and `{{ version }}` variables.
+/// - `{{ package }}` is replaced with the escaped package name
```
marcoieni: what do you mean with "escaped"? Can you make an example?

elpiel:
I actually can't wait to try it out!

jdidion:
This is awesome! Thank you so much for implementing this feature!

PR #2516, opened 2025-11-28
Show full error message for trusted publishing token failure

joshka:
Currently the message doesn't contain enough details to actually diagnose problems with issuing a trusted publishing token. This makes it difficult to configure things correctly.

```
2025-11-28T21:26:57.890714Z  WARN Failed to use trusted publishing: Failed to retrieve token from Cargo registry. Proceeding without it.
```

### DIFF
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -912,7 +912,7 @@ async fn release_package(
                     *trusted_publishing_client = Some(tp);
                 }
                 Err(e) => {
-                    warn!("Failed to use trusted publishing: {e}. Proceeding without it.");
+                    warn!("Failed to use trusted publishing: {e:#}. Proceeding without it.");
                 }
             }
         }


joshka:
for context:
```rust
    #[tokio::test]
    async fn trusted_pub_fail() {
        let client = crate::http_client::http_client_builder().build().unwrap();
        let base_url = super::CRATES_IO_BASE_URL.to_string();
        let jwt = "invalid-jwt-token";
        let err = super::request_trusted_publishing_token(&client, &base_url, jwt)
            .await
            .unwrap_err();
        assert_eq!(
            format!("{err}"),
            "Failed to retrieve token from Cargo registry"
        );
        assert_eq!(
            format!("{err:#}"),
            r#"Failed to retrieve token from Cargo registry: Response body:
{
  "errors": [
    {
      "detail": "Failed to decode JWT"
    }
  ]
}: HTTP status client error (400 Bad Request) for url (https://crates.io/api/v1/trusted_publishing/tokens)"#
        );
    }
```

And this in the wild:

https://github.com/joshka/askama-derive-axum/actions/runs/19774130749/job/56663804471#step:4:251

```
2025-11-28T21:26:57.890714Z  WARN Failed to use trusted publishing: Failed to retrieve token from Cargo registry. Proceeding without it.
```

(The underlying error was that the environment was not set in the github workflow job - unsure if the eventual error message will have this information, but it's still worth noting why token request failed here)

marcoieni (approved):

PR #2574, opened 2026-01-11
Add `custom_minor_increment_regex`

### LINKED ISSUE #2564: Bump minor version for non-conventional commits
## Motivations

In zbus, we don't do conventional commits but something even cooler: emoji prefixes. 😎 However, I don't see a way to tell release-plz to decide whether to bump minor or micro version, based on these prefixes.

- Would you like to implement this feature? n

## Solution

Two solutions come to mind:

Allow configuring this, so I can give a regex pattern to detect if a release only contains bug fixes or other changes.

## Alternatives

I guess that would be to manually update the version with the `set-version` subcommand after `update` command but that's a bit too much work, especially if I want to `release-pr`.

zeenix:
Allow a custom regex setting that triggers a minor version update.

Unfortunately this does not currently work for the case where it's most
useful: non-conventional commits. The reason is that `next_version`
filters out non-conventional commits before checking the regex.

Fixes #2564.

### DIFF
--- a/website/docs/config.md
+++ b/website/docs/config.md
@@ -223,7 +223,7 @@ Example:
 
 ```toml
 [workspace]
-custom_minor_increment_regex = "minor|enhancement|🎉"
+custom_minor_increment_regex = "^minor|^enhancement|^🎉"
 ```
 
 With this configuration, commits like `minor: add feature`, `enhancement: new capability`,


website/docs/config.md
```diff
@@ -202,6 +206,28 @@ This field can be overridden in the [`[package]`](#the-package-section) section.
...
 
+#### The `custom_minor_increment_regex` field
+
+A custom regex pattern to match commit types that should trigger a minor version increment.
+This is useful when you use non-conventional commit prefixes (like emoji prefixes) and want to
+control which commits bump the minor version.
+
+- If the commit message is a conventional commit, the regex is matched against the commit type
+  (the part before the `:` in a commit message).
+- If the commit message is not a conventional commit, the regex is matched against the entire commit message.
```
marcoieni: I changed the PR to follow this behavior but I'm not too sure about it. It can be confusing that we randomly match on the word "minor" in the entire commit.
Maybe we should advertise the use of the character `^` at the beginning of the regex in the docs. Wdyt?

crates/next_version/src/version_updater.rs
```diff
@@ -198,7 +198,11 @@ impl VersionUpdater {
     }
 
     /// Configures a custom regex pattern for minor version increments.
-    /// This will check only the type of the commit against the given pattern.
+    ///
+    /// - For conventional commits, this will check only the type of the commit against the given pattern.
+    /// - For non-conventional commits, this will check the entire commit message against the given pattern.
+    ///
+    /// If you want to match only the beginning of the commit message, use `^` at the start of your regex.
```
marcoieni, line 205: This is a breaking change for the users of the library, so we probably need to bump accordingly, right?

website/docs/config.md
zeenix: > I changed the PR to follow this behavior

Thanks Marco!

> but I'm not too sure about it. It can be confusing that we randomly match on the word "minor" in the entire commit.

Yeah, I agree that matching against the entire commit doesn't sound like a good idea.

> Maybe we should advertise the use of the character `^` at the beginning of the regex in the docs. Wdyt?

I would think that the regex should be matched against the whole commit shortlog/summary (which is what is being picked up for the changelog entry anyway) and not the details section.

crates/next_version/src/version_updater.rs
zeenix: Indeed. Can't we keep the existing behaviour for conventional commits? As I wrote in another comment, I'm not sure we need this change.

marcoieni: for conventional commits we are keeping the existing behavior, I think, yes

zeenix: but the docs here say that this is true for all cases (since it's a paragraph after both bullets points above it). 🤔

website/docs/config.md
zeenix: oh, I only saw #2584 now.

crates/next_version/src/version_updater.rs
marcoieni: I think I changed this on the latest main 👍

marcoieni: I think I changed this on the latest main 👍

PR #2577, opened 2026-01-12
Generate changelog even when version has already been bumped

### LINKED ISSUE #2565: `update` shouldn't give up if version is already bumped
## Bug description

In zbus, we run `cargo-semver` to ensure we don't accidentally break the API and one of the caveats (or one could think of this as a bonus feature too) of that, is that if you add new API, the workflow will expect you to bump the minor version as well. This is also true for deliberate API breaks requiring major version bump. Hence why the version is always already bumped before a PR even merges.

However, `releaze-plz update` sees the version change and seems to give up on generating the changelog for it:

```
2026-01-07T18:01:26.062792Z  INFO determining next version for zbus 5.13.0
2026-01-07T18:01:26.124201Z  INFO zbus: the local package has already a greater version (5.13.0) with respect to the registry package (5.12.0), so release-plz will not update it
```

If there is a way to do this, I didn't find it (yet).

- Would you like to work on a fix? n

## To Reproduce

> Make sure you are able to reproduce the bug in the main branch, too.

As I wrote in #2563, the main branch isn't exactly working for me. I had to install `release-plz` using `cargo binstall`.

> Please try to reproduce the bug in a small example repository.

I'm not sure this is necessary, given that this can be reproduced on any repo (including `release-plz` itself) but if that's not the case, I can provide it.

## Environment

- OS: Fedora 42
- release-plz version: 0.3.150

zeenix:
Currently `release-plz update` gives up on a package completely if its
version has already been bumped. This can be problematic for projects that
only use release-plz for releasing but bump versions already when they
merge an API addition or breakage. In fact, `cargo-semver-checks` even
mandate this and a lot of projects run that in their CI (as they should).

So this commit changes that and now `release-plz` still generates the
changelog (and update the dependencies too) even if version has already
been updated.

Fixes #2565.

### DIFF
--- a/crates/release_plz_core/src/diff.rs
+++ b/crates/release_plz_core/src/diff.rs
@@ -64,6 +64,12 @@ impl Diff {
     }
 
     pub fn should_update_version(&self) -> bool {
+        // `is_version_published = true`` means the current local version equals the registry version
+        // (e.g., local is 0.1.0, registry is 0.1.0).
+        // In this case, if there are new commits, we should bump the version to create a new release.
+        //
+        // If `is_version_published = false`, it means the local version is already bumped,
+        // so we only need to update the changelog, if necessary.
         self.registry_package_exists && !self.commits.is_empty() && self.is_version_published
     }
 


crates/release_plz/tests/all/release_pr.rs
```diff
@@ -1428,6 +1428,91 @@ This PR was generated with [release-plz](https://github.com/release-plz/release-
...
+    // The PR title should mention v0.2.0 (the already-bumped version, not v0.2.1)
+    assert_eq!(opened_prs[0].title, "chore: release v0.2.0");
+
+    let username = context.gitea.user.username();
+    let package = &context.gitea.repo;
+    let pr_body = opened_prs[0].body.as_ref().unwrap().trim();
+
+    // Verify the PR body contains the changelog with both commits.
+    // Note: when version is already bumped, it shows just the version (not a transition like 0.1.0 -> 0.2.0)
+    // and uses releases/tag URL instead of compare URL.
```
marcoieni, line 1469: this is not ideal. We should fix it in a later pr ideally. We should open an issue after this is merged

PR #2593, opened 2026-01-19
Feat/dependency order

secana:


marcoieni:
image

In case you are interested:
- we recently added the git-only feature to release-plz which let you use git releases instead of crates-io.
- https://release-plz.dev/docs/extra/single-changelog

PR #2603, opened 2026-01-23
fix: consider private `git_only` packages

### LINKED ISSUE #2602: release-plz release-pr fails to create a PR when `git_only=true` and `publish=false` in `Cargo.toml`
## Bug description

If `Cargo.toml` contains `publish=false` (documented [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field)) and `release-plz.toml` contains `git_only=true`, then `release-plz release-pr` will not create a PR due to [this check](https://github.com/release-plz/release-plz/blob/5b2830d1c1529b7b333e2b0cd465dc94af57a2c7/crates/release_plz_core/src/project.rs#L113).

The fix is simple: remove `publish=false` from `Cargo.toml`, but presumably, it's somewhat expected that this field might be set to `false` when using `git_only=true`.

- Would you like to work on a fix? [y/n]

## To Reproduce

1. set `publish=false` in `Cargo.toml
2. set `git_only=true` in `release-pls.toml`
3. run `release-plz release-pr --git-token=$(gh auth token)`

## Expected behavior

A PR is created (assuming the package has changes).

## Screenshots

n/a

## Environment

- OS: MacOS 26.2
- release-plz version: `0.3.151`

plaflamme:
When a package is marked with `publish=false` in its manifest, we should still consider it for release if it is `git_only=true`.

Fixes #2602

### DIFF
--- a/crates/release_plz_core/src/command/update/updater.rs
+++ b/crates/release_plz_core/src/command/update/updater.rs
@@ -224,25 +224,8 @@ impl Updater<'_> {
         // Store diff for each package. This operation is not thread safe, so we do it in one
         // package at a time.
 
-        // Collect packages that are either publishable or git-only, with de-duplication, order is important.
-        let mut packages_to_process: Vec<&Package> = Vec::new();
-        let mut package_names: HashSet<String> = HashSet::new();
-
-        // Add publishable packages
-        for p in self.project.publishable_packages() {
-            if package_names.insert(p.name.to_string()) {
-                packages_to_process.push(p);
-            }
-        }
-
-        // Add git-only packages, not already added
-        for p in self.project.workspace_packages() {
-            if self.req.should_use_git_only(&p.name) && package_names.insert(p.name.to_string()) {
-                packages_to_process.push(p);
-            }
-        }
-
-        let packages_diffs_res: anyhow::Result<Vec<(&Package, Diff)>> = packages_to_process
+        let packages_diffs_res: anyhow::Result<Vec<(&Package, Diff)>> = self
+            .packages_to_process()
             .iter()
             .map(|&p| {
                 let diff = self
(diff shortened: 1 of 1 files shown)


marcoieni:
can you also add one test and document this new behavior in the docs?
I think you can take inspiration from https://github.com/release-plz/release-plz/pull/2001/

crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -223,9 +223,26 @@ impl Updater<'_> {
...
+                packages_to_process.push(p);
+            }
+        }
+
+        // Add git-only packages, not already added
+        for p in self.project.workspace_packages() {
+            if self.req.should_use_git_only(&p.name) && package_names.insert(p.name.to_string()) {
+                packages_to_process.push(p);
+            }
+        }
```
marcoieni: isn't it easier to use an hashmap with `<String, &Package>`? and then take just the keys in the iterator. 🤔

plaflamme: fixed in https://github.com/release-plz/release-plz/pull/2603/commits/5c8b5ab045222573df75f153d102d0af8e351546

marcoieni:
I think that the test failure has nothing to do with your changes, I'll have a look. It's probably related to cargo semver checks

marcoieni:
oh, I see the issue, you changed the order of how packages are collected probably.

From the logs:
image

crates/release_plz_core/src/command/update/updater.rs
```diff
@@ -223,11 +223,25 @@ impl Updater<'_> {
...
-            .publishable_packages()
-            .iter()
-            .map(|&p| {
+
+        // Collect packages that are either publishable or git-only, with de-duplication.
+        let mut packages_to_process: HashMap<String, &Package> = HashMap::new();
+
+        // Add publishable packages
+        for p in self.project.publishable_packages() {
+            packages_to_process.insert(p.name.to_string(), p);
```
marcoieni: we need to maintain the order of the packages, so you were right about using a Vec and an hashset instead of an hashmap 😅

plaflamme: fixed in https://github.com/release-plz/release-plz/pull/2603/commits/9e9ede7f38a4b39b14a72326c519da9bd7d109ad and https://github.com/release-plz/release-plz/pull/2603/commits/4f6e159d0bd2f9c8c9cf0d7ef04c4863f0502158

PR #2623, opened 2026-01-29
Draft: fix: graceful hashkind fallback on CargoIndex initialization

ncatelli:
## General

This PR removes the fallback index field added in #2341 in favor of checking for the correct index hash kind at when doing the initial lookup. This fixes the problem where the fallback behavior was never reached due to an early exit being triggered when the initial index is looked up using the newer stable hash kind.

The most notable place i've seen this triggered is when a registry expects the legacy hash kind, is on cargo 1.85+ and does not currently have a local copy of the crates index, such as in CI.

relates to #2340

### DIFF
--- a/crates/release_plz_core/src/cargo.rs
+++ b/crates/release_plz_core/src/cargo.rs
@@ -31,13 +31,23 @@ fn cargo_cmd() -> Command {
 }
 
 pub fn run_cargo(root: &Utf8Path, args: &[&str]) -> anyhow::Result<CmdOutput> {
+    run_cargo_with_env(root, args, &[])
+}
+
+pub fn run_cargo_with_env(
+    root: &Utf8Path,
+    args: &[&str],
+    envs: &[(String, String)],
+) -> anyhow::Result<CmdOutput> {
     debug!("Run `cargo {}` in {root}", args.join(" "));
 
-    let output = cargo_cmd()
-        .current_dir(root)
-        .args(args)
-        .output()
-        .context("cannot run cargo")?;
+    let mut command = cargo_cmd();
+    command.current_dir(root).args(args);
+    for (key, value) in envs {
+        command.env(key, value);
+    }
+
+    let output = command.output().context("cannot run cargo")?;
 
     let output_stdout = String::from_utf8(output.stdout)?;
     let output_stderr = String::from_utf8(output.stderr)?;
--- a/crates/release_plz_core/src/command/release.rs
+++ b/crates/release_plz_core/src/command/release.rs
@@ -21,7 +21,10 @@ use url::Url;
 use crate::{
     CHANGELOG_FILENAME, DEFAULT_BRANCH_PREFIX, GitForge, PackagePath, Project, Publishable as _,
     ReleaseMetadata, ReleaseMetadataBuilder, Remote,
-    cargo::{CargoIndex, CargoRegistry, CmdOutput, is_published, run_cargo, wait_until_published},
+    cargo::{
+        CargoIndex, CargoRegistry, CmdOutput, is_published, run_cargo_with_env,
+        wait_until_published,
+    },
     cargo_hash_kind::{get_hash_kind, try_get_fallback_hash_kind},
     changelog_parser,
     git::forge::GitClient,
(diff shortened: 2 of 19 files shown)


crates/release_plz_core/src/cargo.rs
```diff
@@ -17,8 +17,6 @@ pub struct CargoRegistry {
     /// [`Option::None`] means default 'crate.io'.
     pub name: Option<String>,
     pub index: CargoIndex,
-    /// A fallback registry index to try if the primary `index` fails.
-    pub fallback_index: Option<CargoIndex>,
```
ncatelli, line 21: One of the most notable changes worth calling out is this change to a public interface ~~, so i should probably technically mark that commit as breaking~~. We no longer need this fallback index if we check the index at initialization.

marcoieni:
Thanks for the PR! This greptile thing is new. I'm experimenting with it. Hopefully it's useful.

I will try to review the pr myself soon

ncatelli:
Sounds good. I think the largest comment from greptile was about a comment which i verified to be pretty reflective of what it was doing. As for functionality, the behavior that I'm trying to fix is pretty unique and I have a local environment _nearly_ setup to test it and should have it validated early next week.

Once completed I'll remove the `Draft` and will detail what I'd done for testing.

I've called out some of the larger changes with comments that would be worth manually reviewing.

crates/release_plz_core/src/cargo.rs
marcoieni: it's fine, we don't care about breaking changes compatibilities in release_plz_core

marcoieni:
> Once completed I'll remove the `Draft` and will detail what I'd done for testing.

Ok, let me know when this is ready to be reviewed 👍

crates/release_plz_core/src/command/release.rs
```diff
@@ -653,22 +650,20 @@ async fn release_package_if_needed(
...
-                is_package_published(input, package, primary_index, fallback_index, &token)
+            let Ok((pkg_is_published, mut index)) =
+                is_package_published(input, package, index, &token)
                     .await
                     .with_context(|| {
                         format!("can't determine if package {} is published", package.name)
-                    })?;
+                    })
+            else {
+                continue;
```
ncatelli, line 668: I may rework this to persist an error if it falls through.... but I may be overthinking this.

ncatelli:
Alright this should be ready for initial review.

I've tested this by using latest stable cargo, on an registry that uses the legacy hash kind by clearing the cargo index and running through a full publish to verify that.

a) both indexes attempt lookups
b) if one fails it falls back to legacy gracefully
c) it can instantiate the index from a clean repo.

I'll update the branch and leave it for further review.

ncatelli:
~~Unrelated to this MR that test suite looks to be failing. It should be reviewable in this state. I can look at fixing that tomorrow in a separate MR.~~

Nevermind, it looks like you started addressing this in https://github.com/release-plz/release-plz/pull/2637.

ncatelli:
This should be ready to pull out of draft and for you to ready for you review @marcoieni

crates/release_plz_core/src/command/release.rs
```diff
@@ -656,22 +653,20 @@ async fn release_package_if_needed(
...
+            };
```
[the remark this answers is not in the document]
ncatelli: Called this out on the comment https://github.com/release-plz/release-plz/pull/2623#discussion_r2761593920.

The reason for this is we walk through each registry index attempting to get a success. It's probably fine but I'd love further feedback on this.

Maybe a better idea would be to keep either a running accumulator for failures and throw if it fails on _all_ indexes.

marcoieni:
What's your take on https://github.com/release-plz/release-plz/issues/2334#issuecomment-3856527891
?

ncatelli:
Both of your suggested PR paths look like they would address the core underlying problem that I'm addressing with this PR. I think we can shelf this one for now if you want to start pursuing one of those 2. All other things I can take up on those PRs.

marcoieni:
thanks for working on this. I opted for https://github.com/release-plz/release-plz/pull/2645 instead of this approach. It's simpler. Hopefully it works well

PR #2634, opened 2026-02-01
don't run cargo package if registry package

marcoieni:


marcoieni:
according to AI:

The change is needed because commit detection was using local packaged files, which include Cargo.lock, so a lockfile-only update looked like a package change and leaked into the changelog. By switching to the registry package’s file list (excluding Cargo.lock) when comparing, Cargo.lock-only updates trigger the dedicated “update [Cargo.lock](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) dependencies” entry instead of showing raw commit messages

marcoieni:
close in favor of #3044

Issue #2651, opened 2026-02-06
git-only update fails for non-verifiable crates


kahojyun:
## Summary
In `git-only` mode, `release-plz update` fails when the previously released crate cannot pass `cargo package` verification (e.g. `build.rs` writes outside `OUT_DIR`).

## Reproduction
1. Create a crate and add a `build.rs` that writes files under crate source dir (not `OUT_DIR`).
2. Configure `release-plz` with:
```toml
[workspace]
git_only = true
```
3. First `update + release` succeeds.
4. Make another commit and run `release-plz update` again.

A repro is available at https://github.com/kahojyun/release-workspace/tree/e680f115723c4559c7352cf403733b03509f8240, although it's based on my previous repro repo and not a minimal repro.

```console
$ release-plz update
2026-02-06T04:38:18.326771Z  INFO using release-plz config file release-plz.toml
2026-02-06T04:38:18.367920Z  INFO Latest release of package definitelymylib: tag `definitelymylib-v0.1.2` (version 0.1.2)
2026-02-06T04:38:20.541639Z ERROR failed to determine next versions

Caused by:
    0: run cargo package
[...]

PR #2655, opened 2026-02-06
fix(git-only): handle root-package workspace path dependencies

kahojyun:
Fixes a `git_only` bug in workspaces where the root crate is a package and depends on another workspace member via `path`.

Use workspace packaging during tag-based version discovery so local dependencies are resolved correctly.

Includes integration coverage for this root-package workspace layout.

---

Generated with AI.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -59,12 +59,12 @@ jobs:
         with:
           persist-credentials: false
       - name: Install Rust toolchain
-        uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1.15.2
-      - uses: taiki-e/setup-cross-toolchain-action@b8d1a322a6009a2b7220f53996695778eef89b41 # v1.38.0
+        uses: actions-rust-lang/setup-rust-toolchain@a0b538fa0b742a6aa35d6e2c169b4bd06d225a98 # v1.15.3
+      - uses: taiki-e/setup-cross-toolchain-action@b49e4f79dac0c252ed7af6792f645c007c218bdd # v1.39.0
         with:
           target: ${{ matrix.target }}
         if: startsWith(matrix.os, 'ubuntu') && !contains(matrix.target, '-musl')
-      - uses: taiki-e/install-action@288875dd3d64326724fa6d9593062d9f8ba0b131 # v2.67.30
+      - uses: taiki-e/install-action@f23382d582832e41d5eb4fff2bddb06bc5adf8d3 # v2.68.21
         with:
           tool: cross
         if: contains(matrix.target, '-musl')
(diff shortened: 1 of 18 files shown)


crates/release_plz/tests/all/git_only.rs
```diff
@@ -990,6 +990,76 @@ This PR was generated with [release-plz](https://github.com/release-plz/release-
...
+async fn git_only_update_handles_root_package_path_dependencies() {
+    use cargo_utils::{CARGO_TOML, LocalManifest};
+
+    let context = TestContext::new().await;
+
+    // Workspace layout for this scenario:
+    // - `.` is package `mybin` (root package, binary)
+    // - `crates/mylib` is package `mylib` (workspace member, library)
+    // - `mybin` depends on `mylib` via `path = "crates/mylib"`
+    // Convert root package to `mybin` and add a workspace member at `crates/mylib`.
```
marcoieni, line 1004: instead of configuring these two crates manually, can you configure them by using the test utils `let context = TestContext::new_workspace(&["pkg1", "pkg2"]).await;`?

marcoieni:
how is this test different from `git_only_workspace_level_applies_to_all`?

crates/release_plz/tests/all/git_only.rs
kahojyun: The current test utils only support virtual workspace structures where all crates are placed in the `crates` folder. Adding support for root-level packages requires significant changes to the test utils, and I'm not sure if this should be completed within this PR. I can try making the changes first.

kahojyun:
> how is this test different from `git_only_workspace_level_applies_to_all`?

The previous tests did not cover the scenario where the project root directory is also a package, so they failed to catch the bug caused by the implicit behavioral difference of `cargo package` depending on whether the current directory is a package. The newly added test is indeed very similar to existing tests in terms of process flow, with the only difference being that the workspace root is also a package. We could consider merging it into existing test cases, but I feel that would make the coverage goal of the test case less clear.

marcoieni:
I see. No need to edit the pr, I'll review it again soon

ormandj:
I tested this and `--workspace` is the correct fix — it causes cargo to create a tmp-registry with all workspace members, so versioned path deps resolve correctly.

One thing I verified: using `-p <name>` instead of `--workspace` does NOT work here. With `-p`, cargo doesn't create the tmp-registry, so it looks up sibling deps on crates.io and fails:

```
$ cargo package -p mybin --allow-dirty
error: failed to prepare local package for uploading
Caused by:
  no matching package named `definitelymyprivatelib999` found
  location searched: crates.io index
```

So `--workspace` is the right approach.

This could also be combined with `--no-verify` from #2652 to handle the build.rs verification issue (#2651). The tradeoff is that `--no-verify` only produces `.crate` archives (no unpacked directories), so the `ensure_packaged_dir` unpacking logic from #2652 would be needed.

One remaining edge case `--workspace` can't handle: path deps without a `version` field fail during manifest rewriting regardless of flags. A source-directory fallback for `run_cargo_package()` failures would catch that.

Issue #1316, opened 2024-02-29
Dev dependency not released in order


sd2k:
## Bug description

I have a possibly unusual situation where my workspace contains a crate which is only ever used as a dev-dependency, but should still be released and published. I think `release-plz` should handle this correctly but for some reason I'm seeing this issue on every release.

My crates' dependency graph (including dev dependencies) looks something like:

- augurs-core
- augurs-ets
  - augurs-core
  - augurs-mstl
  [dev-dependencies]
  - augurs-testing
- augurs-mstl
  - augurs-core
  [dev-dependencies]
  - augurs-testing
- augurs-seasons
  [dev-dependencies]
  - augurs-testing
- augurs-testing

For [this release PR](https://github.com/grafana/augurs/pull/64) I expected `augurs-core` and `augurs-testing` to be released first, then `augurs-mstl` and `augurs-seasons`, and finally `augurs-ets`. The [release pipeline](https://github.com/grafana/augurs/actions/runs/7976776252/job/21778113330), however, attempted to release `augurs-mstl` before `augurs-testing`, which obviously failed because there was no matching version on crates.io.

Once I released augurs-testing manually, 

I wonder if this is related to it being a dev dependency, or because it's specified as a workspace dependency (it's in `[workspace.dependencies]` as `augurs-testing = { version = "0.1.2", path = "crates/augurs-testing" }`, and other crates depend on it as `augurs-testing.workspace = true`), or some interaction between the two?

- Would you like to work on a fix? y

## To Reproduce

Hmm, not certain but I think a similar setup to above would do it? Possibly just a 2-crate workspace with one crate _only_ being required as a dev dependency?

## Expected behavior

Dependencies should be released in the correct order.

## Screenshots

N/A

## Environment

- OS: GitHub runner, `ubuntu-latest`
- release-plz version: latest, I think - it's using the GitHub action.

## Additional context

- [Release PR](https://github.com/grafana/augurs/pull/64)
- [Failing CI run](https://github.com/grafana/augurs/actions/runs/7976776252/job/21778113330)
- [Successful CI run](https://github.com/grafana/augurs/actions/runs/7976776252/job/21778591086)
- [workspace Cargo.toml](https://github.com/grafana/augurs/blob/main/Cargo.toml)

sd2k:
[...]

PR #2698, opened 2026-02-28
fix: include dev-dependencies in topological release order

### LINKED ISSUE #2697: Dev-dependencies not included in topological release order, causing cargo publish failures
## Problem

`release_order` excludes dev-dependencies from topological sorting when determining workspace publish order. However, `cargo publish` (via `cargo package`) validates **all** dependencies — including dev-dependencies — against the registry. This means workspace dev-dependencies must be published before their dependents, but release-plz doesn't enforce this ordering.

## Reproduction

In the [fgumi](https://github.com/fulcrumgenomics/fgumi) workspace, `fgumi-metrics` has a **dev-dependency** on `fgumi-sam`:

```toml
# crates/fgumi-metrics/Cargo.toml
[dev-dependencies]
fgumi-sam = { version = "0.1.1", path = "../fgumi-sam" }
```

When release-plz publishes the workspace, it attempts to publish `fgumi-metrics` before `fgumi-sam` because dev-deps are excluded from the topological sort. `cargo package` then fails:

```
failed to select a version for the requirement `fgumi-sam = "^0.1.1"`
candidate versions found which didn't match: 0.1.0
location searched: crates.io index
required by package `fgumi-metrics v0.1.1 (/home/runner/work/fgumi/fgumi/crates/fgumi-metrics)`
```

CI failure: https://github.com/fulcrumgenomics/fgumi/actions/runs/22526673914

## Root Cause

In `crates/release_plz_core/src/release_order.rs`, the `should_dep_be_released_before` function explicitly ignores dev-dependencies:

```rust
fn should_dep_be_released_before(dep: &Dependency, pkg: &Package) -> bool {
    matches!(dep.kind, DependencyKind::Normal | DependencyKind::Build)
      || is_dep_in_features(pkg, &dep.name)
}
```

The comment says "Ignore development dependencies. They don't need to be published before the current package" — but this is incorrect. `cargo package` resolves all dependency kinds against the registry index, not just Normal and Build. When a workspace dev-dep has been version-bumped but not yet published, `cargo package` fails to find the required version.

## Expected Behavior

Dev-dependencies should be included in the topological sort, so workspace crates that are dev-depended upon are published before their dependents.

## Prior Art

This issue has been reported and attempted before:

- #1316 — Dev dependency not released in order
- #1602 — PR to include dev-deps in ordering (closed)
- #1899 — PR for same fix (closed)
- #2593 — Larger PR attempt (closed, too large)

Note that `cargo publish --workspace` (stable since Rust 1.84) handles this correctly by resolving workspace path deps locally.

## Proposed Fix

Change `should_dep_be_released_before` to return `true` for `DependencyKind::Development`, same as Normal and Build dependencies. This is a one-line change.

nh13:
## Summary

`cargo publish` (via `cargo package`) validates **all** dependencies against the registry index, including dev-dependencies that have a version requirement. When workspace dev-deps are version-bumped but not yet published, `cargo package` fails:

```
failed to select a version for the requirement `fgumi-sam = "^0.1.1"`
candidate versions found which didn't match: 0.1.0
location searched: crates.io index
required by package `fgumi-metrics v0.1.1`
```

This PR changes `should_dep_be_released_before` to include dev-dependencies that have a version requirement specified, **matching the exact behavior of `cargo publish --workspace`**.

## Matching cargo's behavior

Cargo's [`local_deps()`](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package/mod.rs) function:

```rust
// If local dev-dependencies don't have a version specified, they get stripped
// on publish so we should ignore them.
if dep.kind() == DepKind::Development && !dep.specified_req() {
    continue;
};
```

| Dev-dep type | `cargo publish --workspace` | This PR |
|---|---|---|
| With version (e.g., `foo = { path = "..", version = "1.0" }`) | **Included** in ordering | **Included** |
| Without version (e.g., `foo = { path = ".." }`) | **Excluded** (stripped on publish) | **Excluded** |

## Reproduction

- Workspace: [fgumi](https://github.com/fulcrumgenomics/fgumi) — `fgumi-metrics` has a versioned dev-dependency on `fgumi-sam`
- CI failure: https://github.com/fulcrumgenomics/fgumi/actions/runs/22526673914
- Verified locally with `cargo publish --dry-run`: publishing the dependent alone fails; `--workspace` succeeds by ordering the dev-dep first

## What changed

- `should_dep_be_released_before` now returns `true` for `DependencyKind::Development` when the version requirement is specified (`!dep.req.comparators.is_empty()`)
- Added `unversioned_dev()` builder to `FakeDependency` for testing path-only dev-deps
- Updated cycle tests: versioned dev-dep cycles are detected; unversioned dev-dep cycles remain allowed
- Added `dev_dep_is_released_before_dependent` test for the core fix
- `workspace_release_order_is_correct` unchanged (release-plz's own dev-deps are unversioned path deps)

## Prior art

This issue was reported in #1316 and attempted in #1602, #1899, and #2593, all closed. This PR is minimal, matches cargo's exact behavior, and includes a concrete reproduction case.

Fixes #2697

### DIFF
--- a/crates/fake_package/src/dependency.rs
+++ b/crates/fake_package/src/dependency.rs
@@ -4,6 +4,7 @@ use cargo_metadata::{Dependency, DependencyKind};
 pub struct FakeDependency {
     name: String,
     kind: DependencyKind,
+    req: String,
 }
 
 impl FakeDependency {
@@ -11,6 +12,7 @@ impl FakeDependency {
         Self {
             name: name.into(),
             kind: DependencyKind::Normal,
+            req: "0.1.0".to_string(),
         }
     }
 
@@ -20,13 +22,23 @@ impl FakeDependency {
             ..self
         }
     }
+
+    /// Create a dev-dependency without a version requirement (path-only).
+    /// These are stripped from the published manifest by cargo.
+    pub fn unversioned_dev(self) -> Self {
+        Self {
+            kind: DependencyKind::Development,
+            req: "*".to_string(),
+            ..self
+        }
+    }
 }
 
 impl From<FakeDependency> for Dependency {
     fn from(dep: FakeDependency) -> Self {
         serde_json::from_value(serde_json::json!({
             "name": dep.name,
-            "req": "0.1.0",
+            "req": dep.req,
             "kind": dep.kind,
             "optional": false,
             "uses_default_features": true,
--- a/crates/release_plz_core/src/release_order.rs
+++ b/crates/release_plz_core/src/release_order.rs
@@ -36,7 +36,7 @@ fn release_order_inner<'a>(
             d.name == *p.name
               // Exclude the current package.
               && p.name != pkg.name
-              && should_dep_be_released_before(d, pkg)
+              && should_dep_be_released_before(d)
         }) {
             anyhow::ensure!(
                 !is_package_in(dep, passed),
@@ -60,27 +60,21 @@ fn is_package_in(pkg: &Package, packages: &[&Package]) -> bool {
     packages.iter().any(|p| p.name == pkg.name)
 }
 
-/// Check if the dependency is enabled in features.
-fn is_dep_in_features(pkg: &Package, dep: &str) -> bool {
-    pkg.features
-        // Discard features name.
-        .values()
-        // Any feature contains the dependency in the format `dep/feature`.
-        .any(|enabled_features| {
-            enabled_features
-                .iter()
-                .filter_map(|feature| feature.split_once('/').map(|split| split.0))
-                .any(|enabled_dependency| enabled_dependency == dep)
-        })
-}
-
 /// Check if the dependency should be released before the current package.
-fn should_dep_be_released_before(dep: &Dependency, pkg: &Package) -> bool {
-    // Ignore development dependencies. They don't need to be published before the current package...
-    matches!(dep.kind, DependencyKind::Normal | DependencyKind::Build)
-      // ...unless they are in features. In fact, `cargo-publish` compiles crates that are in features
-      // and dev-dependencies, even if they are not present in normal dependencies.
-      || is_dep_in_features(pkg, &dep.name)
+/// Normal and Build dependencies always affect release order.
+/// Dev-dependencies affect release order only if they have a version
+/// requirement specified — matching `cargo publish --workspace` behavior.
+/// Dev-deps without a version (`req == "*"`) are stripped from the published
+/// manifest, so they don't need to be published first.
+fn should_dep_be_released_before(dep: &Dependency) -> bool {
+    match dep.kind {
+        DependencyKind::Normal | DependencyKind::Build => true,
+        // Dev-deps with a version requirement remain in the published manifest
+        // and must be resolvable on the registry. Dev-deps without a version
+        // (req == "*") are stripped by cargo on publish, so they can be ignored.
+        DependencyKind::Development => !dep.req.comparators.is_empty(),
+        _ => false,
+    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


marcoieni:
Do `cargo publish --workspace` or `cargo release` also takes into account dev-dependencies?
As you noted, I closed all PRs that were doing this because to my understand this behavior is wrong. Do you have proof that this new behavior you are introducing is correct?

nh13:
Yes — both `cargo publish` and `cargo publish --workspace` validate dev-dependencies against the registry.

Here's a minimal reproduction. Given a workspace with two crates where `crate-b` has a **dev-dependency** on `crate-a`:

```toml
# crate-b/Cargo.toml
[dev-dependencies]
crate-a = { version = "0.1.0", path = "../crate-a" }
```

**Publishing `crate-b` alone fails** because cargo resolves the dev-dep against the registry:

```
$ cargo publish --dry-run --allow-dirty -p crate-b

   Packaging crate-b v0.1.0
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  no matching package named `crate-a` found
  location searched: crates.io index
  required by package `crate-b v0.1.0`
```

**`cargo publish --workspace` succeeds** — it packages `crate-a` first, then `crate-b`:

```
$ cargo publish --dry-run --allow-dirty --workspace

   Packaging crate-a v0.1.0
    Packaged 4 files, 1.2KiB (866B compressed)
   Packaging crate-b v0.1.0
    Packaged 4 files, 1.6KiB (1022B compressed)
   Verifying crate-a v0.1.0
   ...
   Verifying crate-b v0.1.0
   ...
   Uploading crate-a v0.1.0
warning: aborting upload due to dry run
   Uploading crate-b v0.1.0
warning: aborting upload due to dry run
```

This shows that `cargo publish --workspace` **does** include dev-dependencies in its topological ordering — it knows to package/upload `crate-a` before `crate-b`. Since release-plz publishes crates sequentially (not via `cargo publish --workspace`), it needs to replicate this ordering.

You can reproduce this yourself:

```bash
mkdir /tmp/test-ws && cd /tmp/test-ws
cat > Cargo.toml <<'EOF'
[workspace]
resolver = "2"
members = ["crate-a", "crate-b"]
EOF
cargo init --lib crate-a
cargo init --lib crate-b
cat > crate-a/Cargo.toml <<'EOF'
[package]
name = "zzz-test-dep"
version = "0.1.0"
edition = "2021"
description = "test"
license = "MIT"
EOF
cat > crate-b/Cargo.toml <<'EOF'
[package]
name = "zzz-test-main"
version = "0.1.0"
edition = "2021"
description = "test"
license = "MIT"
[dev-dependencies]
zzz-test-dep = { version = "0.1.0", path = "../crate-a" }
EOF

# This fails:
cargo publish --dry-run --allow-dirty -p zzz-test-main

# This succeeds (dev-dep ordered correctly):
cargo publish --dry-run --allow-dirty --workspace
```

nh13:
### Follow-up: exact cargo and cargo-release behavior from source

I dug into the source code of both `cargo` and `cargo-release` to answer your question precisely.

**`cargo publish --workspace`** ([`src/cargo/ops/cargo_package/mod.rs`](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package/mod.rs), `local_deps()` function):

Cargo has a **partial** inclusion of dev-dependencies in publish ordering:

```rust
// If local dev-dependencies don't have a version specified, they get stripped
// on publish so we should ignore them.
if dep.kind() == DepKind::Development && !dep.specified_req() {
    continue;
};
```

- Dev-deps **with** a version (e.g., `foo = { path = "../foo", version = "1.0" }`) → **included** in ordering
- Dev-deps **without** a version (e.g., `foo = { path = "../foo" }`) → **excluded** (stripped from published manifest)

The rationale: versionless dev-deps get stripped from the published `Cargo.toml`, so they create no registry dependency. But versioned dev-deps remain in the published manifest and must be resolvable.

**`cargo-release`** ([`src/ops/cargo.rs`](https://github.com/crate-ci/cargo-release/blob/master/src/ops/cargo.rs), `sort_workspace()`):

cargo-release **excludes all dev-dependencies** from ordering regardless of version:

```rust
let dev_only = dep.dep_kinds.iter()
    .all(|info| info.kind == cargo_metadata::DependencyKind::Development);
if dev_only { None } else { Some(&dep.pkg) }
```

| Tool | Dev-dep with version | Dev-dep without version |
|------|---------------------|------------------------|
| `cargo publish` | **Included** | Excluded |
| `cargo-release` | Excluded | Excluded |

I've updated this PR to match cargo's exact behavior: only include dev-deps that have a version requirement specified. This is the most principled approach since it mirrors what `cargo publish --workspace` does.

I also verified this locally by creating a minimal two-crate workspace where `crate-b` has a versioned dev-dep on `crate-a`, and confirmed with `cargo publish --dry-run` that:
- Publishing `crate-b` alone fails (dev-dep not found on registry)
- `cargo publish --workspace` succeeds (orders `crate-a` before `crate-b`)

The reproduction script from my previous comment demonstrates this.

dkales:
Also hitting this in our workspace where we moved one of the internal crates from a normal to a dev-dependency, would love to see this or some other solution merged.

nh13:
@marcoieni any thoughts on this? Rebased on main and CI's green. I went through the cargo behavior in the comments above. It only orders versioned dev-deps, same as `cargo publish --workspace`. Let me know if there's anything still holding it back.

gustavovalverde:
+1 to this

PR #2757, opened 2026-03-22
feat: add `semver_check_features` config option

### LINKED ISSUE #2291: Customize feature selection when calling semver checker
## Motivations

My crate requires custom features to be properly validated, which I cannot do either via release-plz configuration, nor via the `[package.metadata.cargo-semver-checks]` section in Cargo.toml.

Please provide a way to pass through the semver checker config values.

- Would you like to implement this feature? [y/n] maybe

## Solution

add semver checker feature-related params like `--default-features`, `--only-explicit-features`, and `--features` configurable via release-plz config

## Alternatives

I filed a request for semver checker to allow Cargo.toml / metadata based configuration, but apparently this is far less trivial -- see reply at https://github.com/obi1kenobi/cargo-semver-checks/issues/1355

DaleSeo:
Closes https://github.com/release-plz/release-plz/issues/2291

Currently release-plz invokes `cargo-semver-checks` without any feature flags, so it uses its default heuristic which enables all features except ones prefixed with `_`, `unstable-`, etc. This causes false-positive breaking-change detections for crates whose non-default features intentionally change the API surface.

This adds a `semver_check_features` config field at both workspace and package level. When set, it passes `--only-explicit-features` and `--features <list>` to `cargo-semver-checks`, giving users independent control over which features are checked. The implementation follows the same pattern as `publish_features`. When the field is not set, behavior is unchanged.

Config example:

```toml
[[package]]
name = "my-crate"
semver_check_features = ["default", "server", "client"]
```

### DIFF
--- a/.schema/latest.json
+++ b/.schema/latest.json
@@ -67,7 +67,8 @@
         "release_always": null,
         "release_commits": null,
         "repo_url": null,
-        "semver_check": null
+        "semver_check": null,
+        "semver_check_features": null
       }
     }
   },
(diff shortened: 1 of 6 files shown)


DaleSeo:
Hi @marcoieni 👋, could you please take a look at this PR?

marcoieni:
Hi! I work on this project in my free time. I will have a look when I feel like

marcoieni:
Wdyt about https://github.com/release-plz/release-plz/issues/2291#issuecomment-3034712950 ? I.e. reusing `publish_features` for cargo-semver-checks

PR #2782, opened 2026-04-09
docs: update pinned action version and suggest automated updates

### LINKED ISSUE #2488: Keep action up to date in docs
Update the action in https://release-plz.dev/docs/github/security#-solution-pin-the-action-version
We use renovate

Also, suggest to the reader that they should use renovate or dependabot to keep the action up to date.

mvanhorn:
## Summary

Updates the pinned action example in the security docs from v0.5.50 to v0.5.128 (latest). Adds a tip suggesting Renovate or Dependabot for keeping the pin current.

## Changes

- `website/docs/github/security.md`: Updated commit hash and version tag in the pinned action example. Added a Docusaurus tip admonition about automated dependency update tools.

Fixes #2488

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/website/docs/github/security.md
+++ b/website/docs/github/security.md
@@ -52,12 +52,20 @@ jobs:
       - ...
       - name: Run release-plz
 # highlight-next-line
-        uses: release-plz/action@63ab0c2746bedc448370bad4b0b3d536458398b0 # v0.5.50
+        uses: release-plz/action@1528104d2ca23787631a1c1f022abb64b34c1e11 # v0.5.128
 
 ```
 
 This is the same approach used in the crates.io
 [repository](https://github.com/rust-lang/crates.io/blob/7e52e11c5ddeb33db70f0000bbcdfb01e9b43b0d/.github/workflows/ci.yml#L30C32-L31C1).
+
+:::tip
+
+Use [Renovate](https://docs.renovatebot.com/) or [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates)
+to keep the pinned action version up to date automatically.
+Both tools can detect pinned GitHub Action commits and open PRs when new versions are released.
+
+:::
 
 ## Pass environment variables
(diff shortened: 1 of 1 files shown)


website/docs/github/security.md
```diff
@@ -52,13 +52,21 @@ jobs:
...
 ```
 
 This is the same approach used in the crates.io
 [repository](https://github.com/rust-lang/crates.io/blob/7e52e11c5ddeb33db70f0000bbcdfb01e9b43b0d/.github/workflows/ci.yml#L30C32-L31C1).
 
+:::tip
+
+Use [Renovate](https://docs.renovatebot.com/) or [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates)
+to keep the pinned action version up to date automatically.
+Both tools can detect pinned GitHub Action commits and open PRs when new versions are released.
```
marcoieni, line 66: are you sure Dependabot can do this?
Can you point to an example?

mvanhorn:
Fair to push back, and yes, Dependabot does handle SHA-pinned actions. It updates the commit SHA and rewrites the trailing version comment.

A merged example: [ossf/scorecard#5172](https://github.com/ossf/scorecard/pull/5172), opened by Dependabot in its `github-actions` group.

```diff
- uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+ uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
```

Scorecard is a useful example because it requires SHA pinning as policy, so every action there is pinned and Dependabot keeps them current.

One caveat worth adding to the tip: Dependabot relies on that `# vX.Y.Z` comment to know the current version. If the SHA is pinned with no version comment it cannot tell what to bump from. I will add that note unless you would rather cut the Dependabot half entirely.

Issue #2595, opened 2026-01-20
Bug: release-plz update fails for workspaces with internal dependencies in git_only mode


kahojyun:
## Bug description

In `git_only` mode, release-plz fails to handle workspace internal dependencies when running `cargo package`. When a binary crate depends on a library crate within the same workspace, the packaging process fails because cargo cannot find the dependency metadata (it searches [crates.io](http://crates.io/) instead of using the local workspace).

## To Reproduce

1. Create a workspace containing two crates: `definitelymylib` (library) and `mybin` (binary)
2. Configure `mybin` to depend on `definitelymylib` with both at initial version `0.1.0`
3. Run `release-plz update` and `release-plz release` for the first time (this works correctly - changelog is generated and tags are created)
4. Run `release-plz update` for the second time
5. Observe the error: `cargo package` fails with "no matching package named `definitelymylib` found, location searched: crates.io index"

## Expected behavior

The `release-plz update` command should successfully package workspace crates even when they have internal dependencies, without requiring those dependencies to be published to [crates.io](http://crates.io/).

## Logs

```bash
release-plz update
2026-01-20T03:40:13.061413Z  INFO using release-plz config file release-plz.toml
2026-01-20T03:40:13.122181Z  INFO Latest release of package definitelymylib: tag `definitelymylib-v0.1.0` (version 0.1.0)
2026-01-20T03:40:16.132506Z  INFO Latest release of package mybin: tag `mybin-v0.1.0` (version 0.1.0)
2026-01-20T03:40:18.655758Z  INFO determining next version for definitelymylib 0.1.0
2026-01-20T03:40:19.012659Z  INFO determining next version for mybin 0.1.0
2026-01-20T03:40:20.245849Z ERROR failed to determine next versions

Caused by:
    0: failed to retrieve difference of package mybin
    1: failed to check package equality for `mybin` at commit a62bc7ee88e584ec471c434bedee95a82dfd931c
    2: cannot compare packages
    3: cannot determine packaged files of registry package "/private/var/folders/q2/152_qmh54jz303ydh5mmk6zh0000gn/T/release-plz-mybin-2F5up2/worktree/target/package/mybin-0.1.0"
    4: error while running `cargo package`: error: no matching package named `definitelymylib` found
       location searched: crates.io index
[...]

Issue #2494, opened 2025-11-18
Bug: Changelog missing commits after git tag - only shows most recent commit


tusharmath:
When `release-plz` generates a changelog, it's missing all commits that occurred after the last git tag, even though those commits modified package source files. Only the most recent commit appears in the changelog.

## Environment

- **release-plz version**: v0.5 (via GitHub Actions)
- **Repository**: https://github.com/tailcallhq/gh-workflow
- **Affected PR**: #176, #178
- **Workspace**: Multi-package workspace with 3 crates

## Configuration

`.release-plz.toml`:
```toml
[workspace]
dependencies_update = true
release_always = false

[[package]]
name = "gh-workflow-macros"
version_group = "core"

[[package]]
name = "gh-workflow"
version_group = "core"
```

## Reproduction Steps

1. **Initial State**: 
   - Package `gh-workflow` published at v0.8.0 on crates.io
   - Git tag `gh-workflow-v0.8.0` exists at commit `8ec2428`

2. **Made 10+ commits** between 2025-09-08 and 2025-11-18 that modified `crates/gh-workflow/src/`:
   ```bash
   cd43270 feat(job): set default runs_on to ubuntu-latest
   ea47813 refactor(container): change Port enum serialization to untagged
   32814ec refactor(job): reorder fields to match GitHub Actions YAML structure
   a69df30 feat(job): add builder methods for outputs, services, and secrets
   23be201 Add Step<Run>::shell (#174)
   049a1e8 feat: add `WorkflowCall::add_input` (#173)
   6b8ca86 chore(lint): add workspace clippy use_self lint
   # ... plus autofix commits
   ```

3. **Verified commits modified package**:
   ```bash
   $ git log --oneline gh-workflow-v0.8.0..592d4dd -- crates/gh-workflow/
   # Shows 10 commits that changed files in crates/gh-workflow/src/
   
   $ git diff gh-workflow-v0.8.0 592d4dd -- crates/gh-workflow/src/ | wc -l
   648 lines of diff
   ```

4. **Release-plz ran** and created PR #176

## Expected Behavior

The changelog should include all 10 commits:
```markdown
## [0.8.1] - 2025-11-18

### Added
- feat: add `WorkflowCall::add_input`
- feat(job): add builder methods for outputs, services, and secrets
- feat(job): set default runs_on to ubuntu-latest
- Add Step<Run>::shell

### Changed
- refactor(job): reorder fields to match GitHub Actions YAML structure
[...]

Issue #2147, opened 2025-03-26
`git_tag_name` default should not consider `publish = false` workspace


JounQin:
## Bug description

- Would you like to work on a fix? [y/n]

## To Reproduce

https://github.com/unrs/unrs-resolver/blob/d0ee2a8717cb11bca2b488be1695ff8a684f99ee/napi/Cargo.toml#L5

## Expected behavior

`"v{{ version }}"` should be used by default in this case

## Environment

- OS: Ubuntu 24.04.2
- release-plz version: MarcoIeni/release-plz-action@4cd77ee4d22f0cdb1a461e6eb3591cddc5e1f665 # v0.5

marcoieni:
PR welcome 👍
Also it would be nice to add an integration test for this. Let me know if you want to work on it 👍

JounQin:
I'll find time to contribute.

PR #2800, opened 2026-04-17
feat(update): opt out of cascade version bumps for private packages

### LINKED ISSUE #2799: Cascade version bumps pollute release PR and changelog for `publish = false` crates in monorepos
## Problem

In a Cargo workspace where **all** (or most) crates have `publish = false` (e.g. a service monorepo where only one crate is the shipped binary and the rest are internal library crates), release-plz's cascade-bump behaviour produces noisy, confusing release PRs:

1. One crate with a real commit (e.g. `feat(query): add X` touching only `crates/query/`) triggers a version bump for `query`.
2. Every other crate that (transitively) depends on `query` via `workspace = true` path-deps gets a **cascade bump**, even though no source files in those crates changed.
3. For cascaded crates, release-plz synthesises a `chore: updated the following local packages: …` commit. That commit is usually filtered out by `commit_parsers` (most projects skip `chore`), leaving **no new changelog content**. The PR body then **re-displays the previous (old) changelog entry** for each cascaded crate — so the release PR for v0.3.20 ends up showing entries dated v0.3.9, v0.3.11, v0.3.12 for crates that have had no changes since.
4. Same noise ends up in `crates/*/CHANGELOG.md` as empty-ish sections with the old body.

This is **semantically correct** for published crates (downstream consumers need a new version to pull in the new dependency), but for `publish = false` crates there are no downstream consumers — the cascade is pure noise.

## Reproduction

A monorepo similar to the pattern described in #2147 / #2602 / #2595, e.g.:

```toml
# workspace Cargo.toml
[workspace.package]
publish = false

[workspace.dependencies]
lib-a = { path = "crates/lib-a" }
lib-b = { path = "crates/lib-b" }  # depends on lib-a
```

All crates are `publish = false`. Release-plz with `git_only = true`.

1. Commit `feat(lib-a): add foo` that only touches `crates/lib-a/`.
2. Release-plz creates a release PR. Expected: only `lib-a` is bumped, only `lib-a` appears in the PR body.
3. Actual: both `lib-a` **and** `lib-b` get a patch bump. `lib-b`'s section in the PR body shows its last real changelog entry (from months ago) because no new conventional commits exist for it. `crates/lib-b/CHANGELOG.md` gets a new heading with an empty/filtered body.

## Expected behaviour

For crates with `publish = false`, release-plz should **not** cascade-bump them when a dependency changes. They should only appear in the release PR when they have their own direct conventional commits since baseline.

## Proposed solution

Add a new workspace-level config option, defaulting to the semantically safer value:

```toml
[workspace]
# When true (default for publish=false crates), skip cascade version bumps
# for packages that are `publish = false`. They will only appear in the
# release PR if they have direct commits since the baseline tag.
# Opt-out per-package with `cascade_bumps = true`.
```

Minimal behaviour change: in `crates/release_plz_core/src/command/update/updater.rs` → `dependent_packages_update`, add a filter that skips packages where `!p.is_publishable()` unless explicitly overridden.

### Why this is safe

- `publish = false` means the Cargo manifest author declared "no registry consumes this crate". There is by definition no one whose builds would break from a stale version number.
- For `Cargo.lock` consistency: the lock file is regenerated by `cargo` on the next build; the absence of a manifest version bump does not break lockfile semantics.
- For users who want the current behaviour: opt-in per-package override.

### Related issues

- #2147 — also about `publish = false` workspace semantics being disrespected by defaults.
- #2494 — different bug (missing commits) but same area (changelog generation in monorepos).
- #1029 — release PR groups (related but more ambitious).

## Would you like to work on a fix?

Yes — I'll open a PR shortly with the implementation, tests, and docs updates.

polaz:
Closes #2799.

## Summary

Adds a new per-package config field `dependent_update: Option<bool>` that
controls whether a package is cascade-bumped when one of its workspace
dependencies changes.

- `true`: always cascade-bump (the legacy behaviour for all packages).
- `false`: never cascade-bump.
- Unset (default): auto — cascade-bump iff the package is publishable
  (i.e. `publish` is not `false`/`[]` in `Cargo.toml`). `publish = false`
  crates have no downstream consumers, so cascade bumps produce only
  empty changelog sections and uninformative tags.

## Why

See #2799 for the motivating scenario. Short version: in monorepos where
most/all crates are `publish = false` (e.g. a service monorepo with one
shipped binary + several internal library crates), the current cascade
behaviour produces noisy release PRs where unrelated crates get their
previous changelog entry re-rendered under a new version header,
because the synthetic `chore: updated the following local packages`
commit is typically filtered out by `commit_parsers` and the
package has no real commits since baseline.

## What changed

- `release_plz_core::UpdateConfig` gains `dependent_update: Option<bool>`.
- `PackageUpdateConfig` gains a helper `should_cascade_bump(is_publishable_fallback)`.
- `Updater::dependent_packages_update` checks the flag before queuing a
  cascade bump; when `dependent_update` is `None`, it falls back to
  the package's `Package::is_publishable()` (from Cargo `publish`).
- `release_plz::config::PackageConfig` gains the same field (with
  `# Dependent Update` docs), threaded through `merge()` and the
  `From<PackageConfig> for UpdateConfig` conversion.
- JSON schema regenerated via `release-plz generate-schema`.
- Docs updated in `website/docs/config.md` for both workspace and
  package sections.

## Testing

Added two unit tests in `update_config::tests`:

- `should_cascade_bump_defaults_to_publishable` — verifies the `None`
  default falls back to the `is_publishable` argument.
- `should_cascade_bump_explicit_override_wins` — verifies that
  explicit `Some(true)` / `Some(false)` override the auto behaviour.

All existing 73 unit tests in `release_plz_core` still pass.
`cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings`
are clean.

## Backwards compatibility

The default behaviour changes only for `publish = false` crates. For
publishable crates (the majority of release-plz users), nothing changes:
`is_publishable()` returns `true`, `dependent_update` defaults to `None`,
auto-resolves to `true`, and cascade bumps happen exactly as before.

Users who rely on cascade bumps for `publish = false` crates can
restore the old behaviour at the workspace level:

```toml
[workspace]
dependent_update = true
```

or per-package:

```toml
[[package]]
name = "my-private-but-cascade-bumping-crate"
dependent_update = true
```

## Related

- #2147 — also about `publish = false` defaults being disrespected.
- #2494 — different bug, same area (changelog in monorepos).
- #2595 — `git_only` + path deps (already fixed upstream and in this tree).

### DIFF
--- a/.schema/latest.json
+++ b/.schema/latest.json
@@ -41,6 +41,7 @@
         "custom_major_increment_regex": null,
         "custom_minor_increment_regex": null,
         "dependencies_update": null,
+        "dependent_update": null,
         "features_always_increment_minor": null,
         "git_only": null,
         "git_release_body": null,
@@ -302,6 +303,14 @@
           "description": "Custom regex to match commit types that should trigger a minor version increment.\nUseful when using non-conventional commit prefixes.",
           "type": [
             "string",
+            "null"
+          ]
+        },
+        "dependent_update": {
+          "title": "Dependent Update",
+          "description": "Whether to bump this package when one of its workspace dependencies changes.\n\n- `true`: always cascade-bump (the legacy behaviour for all packages).\n- `false`: never cascade-bump (opt out explicitly).\n- Unset (default): auto — cascade-bump iff the package is publishable\n  (i.e. `publish` is not `false`/`[]` in Cargo.toml). `publish = false`\n  crates have no downstream consumers, so cascade bumps produce only\n  noise in the release PR and changelog.\n\nSee <https://github.com/release-plz/release-plz/issues/2799>.",
+          "type": [
+            "boolean",
             "null"
           ]
         },
@@ -575,6 +584,14 @@
         "dependencies_update": {
           "title": "Dependencies Update",
           "description": "- If `true`, update all the dependencies in the Cargo.lock file by running `cargo update`.\n- If `false` or [`Option::None`], only update the workspace packages by running `cargo update --workspace`.",
+          "type": [
+            "boolean",
+            "null"
+          ]
+        },
+        "dependent_update": {
+          "title": "Dependent Update",
+          "description": "Whether to bump this package when one of its workspace dependencies changes.\n\n- `true`: always cascade-bump (the legacy behaviour for all packages).\n- `false`: never cascade-bump (opt out explicitly).\n- Unset (default): auto — cascade-bump iff the package is publishable\n  (i.e. `publish` is not `false`/`[]` in Cargo.toml). `publish = false`\n  crates have no downstream consumers, so cascade bumps produce only\n  noise in the release PR and changelog.\n\nSee <https://github.com/release-plz/release-plz/issues/2799>.",
           "type": [
             "boolean",
             "null"
--- a/crates/release_plz/src/config.rs
+++ b/crates/release_plz/src/config.rs
@@ -464,6 +464,18 @@ pub struct PackageConfig {
     /// Custom regex to match commit types that should trigger a major version increment.
     /// Useful when using non-conventional commit prefixes.
     pub custom_major_increment_regex: Option<String>,
+    /// # Dependent Update
+    /// Whether to bump this package when one of its workspace dependencies changes.
+    ///
+    /// - `true`: always cascade-bump (the legacy behaviour for all packages).
+    /// - `false`: never cascade-bump (opt out explicitly).
+    /// - Unset (default): auto — cascade-bump iff the package is publishable
+    ///   (i.e. `publish` is not `false`/`[]` in Cargo.toml). `publish = false`
+    ///   crates have no downstream consumers, so cascade bumps produce only
+    ///   noise in the release PR and changelog.
+    ///
+    /// See <https://github.com/release-plz/release-plz/issues/2799>.
+    pub dependent_update: Option<bool>,
 }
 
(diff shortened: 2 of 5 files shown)


polaz:
Extended the PR with a second commit (5d8d2bf) covering the other code path the same issue manifests through: **workspace-version inheritance**.

Problem with just the first commit:

In monorepos that use `[workspace.package] version = "X"` + `version.workspace = true` in every crate (a very common pattern for tightly-coupled service monorepos), when the workspace version bumps from X → Y, **every** inheriting crate has `next_version != current_version` in the main `packages_to_update` loop — regardless of whether that crate had any commits since baseline. The first commit only filtered the cascade-through-dep-graph path (`dependent_packages_update`); the workspace-inheritance path still queued untouched crates into the update list.

The symptom is very visible in release PRs: untouched crates appear in the body with their previous changelog entry re-rendered under the new version header (see screenshots in #2799 — entries dated months ago show up in a PR for today's release).

The second commit applies the same `should_cascade_bump()` check in the main loop, right before the version-changed / not-published decision. Same opt-out behaviour: `publish = false` ⇒ auto-skip; overridable via `dependent_update = true/false` at workspace or per-package level.

Both commits share the `should_cascade_bump` helper and the `dependent_update` config field, and both tests added in the first commit cover the same helper — no new tests needed for the second commit since it calls the same resolver.

PR #2812, opened 2026-04-25
fix(set-version): update Cargo.lock after version changes

zeitlinger:
## Summary
- refresh the workspace Cargo.lock after `set-version` updates package versions
- only do this when a workspace-root Cargo.lock already exists
- cover both workspace and single-package flows in the integration tests

## Testing
- cargo test -p release-plz --features docker-tests set_version -- --nocapture


tests/fixtures/set-version-in-package/Cargo.lock
```diff
@@ -0,0 +1,7 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
```
marcoieni: we are at version 4. How was this file generated?

zeitlinger: I regenerated this with the current Cargo after making the package fixture explicitly standalone via an empty `[workspace]` table. Without that marker, running `cargo generate-lockfile` from inside the fixture is blocked by the parent repo workspace; with it, Cargo generates lockfile format version 4.

marcoieni:
tested and it works great, thanks!

PR #2861, opened 2026-05-31
fix: surface cargo-semver-checks tool failures instead of treating them as compatible

### LINKED ISSUE #2294: Error from cargo-semver-checks gets treated as "API compatible change"
## Bug description

I appear to have had an old version of cargo-semver-checks installed, so it says this when executed stand-alone:

```
error: unsupported rustdoc format v43 for file: ...
```

However, release-plz declares this crate as having `(✓ API compatible changes)`, without any hint of problems. This caused me to publish some packages incorrect version numbers, incrementing minor when should have incremented major.

- Would you like to work on a fix? N

## To Reproduce

1. Use an older version of cargo-semver-checks that does not support the rustdoc version of the currently installed Rust toolchain.
2. `release-plz update`

## Expected behavior

Error from cargo-semver-checks is reported as release-plz error.

## Environment

- OS: Windows 11
- release-plz version: v0.3.137

estelledc:
Closes #2294.

## Problem

`cargo-semver-checks` exits with code 1 for both **detected breaking changes** and **tool failures** ([upstream confirms](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/main.rs) it has only two exit codes — 0 success, 1 anything-else). release-plz disambiguates by looking for the string `"semver requires new major version"` in stderr.

The previous code's fallback was wrong: any non-zero exit *without* that marker was reported as `SemverCheck::Compatible`. So when a user had an outdated `cargo-semver-checks` binary and the run failed with `error: unsupported rustdoc format v43`, release-plz silently printed `(✓ API compatible changes)` and the user published a minor bump for an actually-breaking change. That's the symptom in #2294.

## Fix

Treat any non-success exit without the breaking-change marker as a real error and bail with the underlying stderr. Compatibility data we cannot verify must not be presented as if it were verified.

The classification logic is pulled into a small pure function (`classify_semver_check_output`) so it can be unit-tested without invoking the cargo-semver-checks binary. Tests cover the three branches: success → Compatible, exit 1 with marker → Incompatible (with stdout passed through), exit 1 without marker → bail (the regression).

## Behavior change to be aware of

This is a behavioral change for users whose `cargo-semver-checks` install is broken: before this PR release-plz would limp along reporting Compatible; after this PR it errors out. I think that's the correct trade-off given the bug report — silently misclassifying breaking changes is worse than failing loudly — but happy to gate this behind a flag (or a warning-only path) if you'd prefer a softer rollout.

## Test

```
$ cargo test -p release_plz_core --lib semver_check::tests
running 3 tests
test semver_check::tests::success_is_compatible ... ok
test semver_check::tests::breaking_change_is_incompatible ... ok
test semver_check::tests::tool_failure_is_reported_not_swallowed ... ok
test result: ok. 3 passed; 0 failed
```

`cargo fmt` and `cargo clippy --no-deps -p release_plz_core -- -D warnings` clean.

### DIFF
--- a/crates/release_plz_core/src/semver_check.rs
+++ b/crates/release_plz_core/src/semver_check.rs
@@ -79,19 +79,118 @@ pub fn run_semver_check(
         fs_err::remove_dir_all(registry_target_dir)?;
     }
 
-    if output.status.success() {
-        Ok(SemverCheck::Compatible)
-    } else {
-        let stderr = String::from_utf8(output.stderr)?;
-        if stderr.contains("semver requires new major version") {
-            let stdout = strip_ansi_escapes::strip(output.stdout);
-            let stdout = String::from_utf8(stdout)?.trim().to_string();
-            if stdout.is_empty() {
-                anyhow::bail!("unknown source of semver incompatibility");
-            }
-            Ok(SemverCheck::Incompatible(stdout))
-        } else {
-            Ok(SemverCheck::Compatible)
+    let stderr = String::from_utf8(output.stderr)?;
+    let stdout = output.stdout;
+    classify_semver_check_output(output.status.success(), &stderr, &stdout)
+}
+
+/// Decide what `cargo-semver-checks` told us based on its exit status and output.
+///
+/// cargo-semver-checks uses exit code 1 for both detected breaking changes and
+/// tool failures (e.g. unsupported rustdoc format, IO errors), so we have to
+/// look at stderr to tell them apart. Without that distinction a tool failure
+/// silently gets reported as `(✓ API compatible changes)` and the user
+/// publishes a wrong version.
+/// See <https://github.com/release-plz/release-plz/issues/2294>
+fn classify_semver_check_output(
+    success: bool,
+    stderr: &str,
+    stdout: &[u8],
+) -> anyhow::Result<SemverCheck> {
+    if success {
+        return Ok(SemverCheck::Compatible);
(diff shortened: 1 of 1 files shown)


estelledc:
Force-pushed an update addressing the CI failures on the previous revision:

- **Clippy**: bare URL in a doc comment now wrapped in `<>`.
- **Test Suite (docker)**: integration tests exposed a real semantic gap in my fix — `cargo-semver-checks` exits non-zero with `error: no library targets found in package` when the registry baseline package legitimately has no lib target (the integration suite exercises this when a package was a binary in the previous release). Under the previous fix that gets misclassified as a tool failure and `release-plz` bails. Added an explicit branch that treats `"no library targets found"` as Compatible, matching the historical behavior for that path, plus a new unit test (`no_library_targets_is_treated_as_compatible`) covering it.

The classification now is:

- exit 0 → Compatible
- exit 1 + `"semver requires new major version"` → Incompatible
- exit 1 + `"no library targets found"` → Compatible (tool not applicable)
- exit 1 + anything else → bail with the underlying stderr (the bug fix)

Local: `cargo clippy --all-targets --all-features --workspace -- -D warnings`, `cargo fmt --check`, and the 4 unit tests all clean.

estelledc:

francisdb:
I'd like to see this one revived and applied, this is a serious issue where incompatible changes are not marked as such.

Issue #1554, opened 2024-07-12
`set-version`: add support for workspace version


marcoieni:
## Motivations

It would be nice to change the workspace [version](https://doc.rust-lang.org/cargo/reference/workspaces.html?highlight=workspace#the-package-table) via release-plz.

However, at the moment `release-plz set-version 1.2.3` only works in projects with a single crate (not in a workspace).

## Solution

`release-plz set-version <version>` is also allowed in workspaces that have a workspace version.
There's already a test written in https://github.com/MarcoIeni/release-plz/pull/1552 (I don't plan to work on that PR, so if you want to see this happen, you can write it here and start working on it).

You should update all the changelogs of all the packages that inherit the workspace version 👍

PR #2864, opened 2026-05-31
fix(set-version): honor workspace-level changelog_path

### LINKED ISSUE #2441: set-version fails for shared workspace changelog
## Bug description

In ratatui, we have a single workspace changelog. Attempting to call set-version on a project fails as it doesn't respect the setting:

```toml
[workspace]
changelog_path = "./CHANGELOG.md"
```

```
❯ release-plz set-version ratatui@0.30.0-beta.0
2025-10-01T20:10:07.788131Z  INFO using release-plz config file release-plz.toml
2025-10-01T20:10:07.845303Z ERROR failed to update changelog at /Users/joshka/local/ratatui/ratatui/CHANGELOG.md

Caused by:
    failed to open file `/Users/joshka/local/ratatui/ratatui/CHANGELOG.md`: No such file or directory (os error 2)
Error: failed to update changelog at /Users/joshka/local/ratatui/ratatui/CHANGELOG.md

Caused by:
    failed to open file `/Users/joshka/local/ratatui/ratatui/CHANGELOG.md`: No such file or directory (os error 2)
```

- Would you like to work on a fix? [y/n] n

## To Reproduce

```sh
gh clone ratatui/ratatui
gh pr checkout 2120
release-plz set-version ratatui@0.30.0-beta.0
```

## Expected behavior

Version is updated in Cargo.toml and changelog

## Environment

- OS: macOS
- release-plz version: release-plz 0.3.147

estelledc:
Closes #2441.

## Problem

In monorepos that share a single workspace changelog (the ratatui setup that the OP reported, but also any project using `[workspace] changelog_path = "./CHANGELOG.md"`), `release-plz set-version` ignores the workspace setting and falls back to the per-crate default. The command then errors out because that per-crate file doesn't exist:

```
$ release-plz set-version ratatui@0.30.0-beta.0
ERROR failed to update changelog at /Users/joshka/local/ratatui/ratatui/CHANGELOG.md

Caused by:
    failed to open file ...: No such file or directory
```

`Config::fill_set_version_config` only iterates `self.packages()` — i.e. `[[package]]` sections — and never reads `self.workspace.packages_defaults.changelog_path`. So with workspace-only config, every `VersionChange` ends up with `changelog_path: None` and the code falls through to `pkg_path.join("CHANGELOG.md")`. (`update` and `release` don't have this problem because `UpdateRequest::changelog_path` does merge in the workspace default — `update_request.rs:78-84`.)

## Fix

Two surgical changes:

**`fill_set_version_config`** now applies the workspace-level default first, then per-package overrides. Per-package config still wins for any package that overrides it.

**New `SetVersionRequest::set_default_changelog_path`** walks both the `Single` and `Workspace` spec variants and sets the path on every entry that doesn't already have one. Symmetric with the existing `set_changelog_path` but only fills empties — `is_none()` guards keep per-package overrides intact.

**Path resolution in `set_version_in_package`** now joins a configured `changelog_path` to the workspace manifest's parent directory, mirroring `UpdateRequest::changelog_path`. Without this, `./CHANGELOG.md` was being resolved against the process cwd — unreliable and inconsistent with how every other release-plz command interprets the field.

## Tests

Three new unit tests in `command::set_version::tests`:

- `workspace_default_propagates_to_all_packages` — workspace default lands on every package in a workspace spec
- `per_package_override_wins_over_workspace_default` — order independence: per-package wins regardless of when set_default_changelog_path is called
- `workspace_default_applies_to_single_spec` — the single-package case

```
$ cargo test -p release_plz_core --lib command::set_version::tests
running 3 tests
test command::set_version::tests::workspace_default_applies_to_single_spec ... ok
test command::set_version::tests::workspace_default_propagates_to_all_packages ... ok
test command::set_version::tests::per_package_override_wins_over_workspace_default ... ok
test result: ok. 3 passed
```

`cargo clippy --all-targets --all-features --workspace -- -D warnings` and `cargo fmt --check` clean.

## Note

This does not fully address #1554 (workspace-level *version* support), only the changelog-path piece of #2441 that's specific to set-version.

### DIFF
--- a/crates/release_plz/src/config.rs
+++ b/crates/release_plz/src/config.rs
@@ -86,6 +86,14 @@ impl Config {
         &self,
         set_version_request: &mut SetVersionRequest,
     ) -> anyhow::Result<()> {
+        // Apply the workspace-level default first so every package picks it
+        // up. Per-package `[[package]] changelog_path` is applied below and
+        // overrides this default for that specific package.
+        // See <https://github.com/release-plz/release-plz/issues/2441>
+        if let Some(changelog_path) = self.workspace.packages_defaults.changelog_path.clone() {
+            let changelog_path = to_utf8_pathbuf(changelog_path)?;
+            set_version_request.set_default_changelog_path(changelog_path);
+        }
         for (package, config) in self.packages() {
             if let Some(changelog_path) = config.common.changelog_path.clone() {
                 let changelog_path = to_utf8_pathbuf(changelog_path)?;
--- a/crates/release_plz_core/src/command/set_version.rs
+++ b/crates/release_plz_core/src/command/set_version.rs
@@ -32,6 +32,31 @@ impl SetVersionRequest {
             }
         }
     }
+
+    /// Apply a workspace-level default changelog path to every package that
+    /// does not already have its own per-package override.
+    ///
+    /// Without this, `[workspace] changelog_path = "./CHANGELOG.md"` would be
+    /// silently ignored by `set-version` and the command would look for a
+    /// per-crate `CHANGELOG.md` that doesn't exist in monorepos that share a
+    /// single workspace changelog.
+    /// See <https://github.com/release-plz/release-plz/issues/2441>
+    pub fn set_default_changelog_path(&mut self, changelog_path: Utf8PathBuf) {
+        match &mut self.version_changes {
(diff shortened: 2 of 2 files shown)


estelledc:

PR #2865, opened 2026-05-31
docs(faq): explain why empty commits don't trigger a version bump

### LINKED ISSUE #2673: Empty commit to force major version bump not picked up
## Bug description

I tried the following:

```
git commit --allow-empty -m 'feat!: force major version bump after adding mesh generation'
```

This was however not picked up and I needed to do a non-empty commit to make this work

https://github.com/francisdb/vpin/blob/main/.github/workflows/release-plz.yml

Empty:
https://github.com/francisdb/vpin/commit/ace0c0cc67566173abe05efcf41a6c33fa96982e

With small readme change:
https://github.com/francisdb/vpin/commit/974855bb75e11c2bd3e13fd5dd1ad04679ae7ff6

- Would you like to work on a fix? [n]

estelledc:
Closes #2673.

The reporter ran `git commit --allow-empty -m 'feat!: force major version bump'` expecting release-plz to pick it up, and was surprised when only a real content change (a README edit) actually triggered the bump. @marcoieni explained why in the issue thread:

> release-plz compares the two packages and detects that they are the same so it doesn't bump the version.

The reporter then asked:

> Maybe adding a note to https://release-plz.dev/docs/faq#release-plz-bumped-the-version-in-a-way-i-didnt-expect could be enough to close this?

@marcoieni said yes, PR welcome — that's this PR.

I added a subsection under the existing "Release-plz bumped the version in a way I didn't expect" entry. It states the cause concisely (release-plz diffs against the registry, not commit messages alone) and lists the two reliable workarounds:

1. Make any real content change so the package differs from the registry copy.
2. Use `release-plz set-version` to bump explicitly, then commit normally.

Pure docs change, no code touched.

### DIFF
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -83,3 +83,21 @@ Release-plz uses the [next_version](https://crates.io/crates/next_version)
 crate to determine the next version.
 Please read the [documentation](https://docs.rs/next_version/latest/next_version/),
 and open an issue if it's not clear enough.
+
+### Empty commits don't trigger a bump
+
+Release-plz detects releaseable changes by diffing the local package against
+the registry version, not by parsing every commit message in isolation. An
+empty commit (e.g. `git commit --allow-empty -m 'feat!: force major bump'`)
+leaves the package contents identical to the registry version, so release-plz
+sees no changes to release and does not bump the version.
+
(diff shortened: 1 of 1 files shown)


estelledc:

PR #2867, opened 2026-05-31
docs(set-version): document the release-PR workflow + run-on-main footgun

### LINKED ISSUE #2555: Not documented how `set-version` workflow works
I have `release-plz` configured via GitHub Actions to automatically open a PR.

I recently forgot to add the `feat:` prefix in a pull request and got a wrong version bump.

I saw documentation page https://release-plz.dev/docs/usage/set-version and it seems this is the tool that I need Except the intended workflow how to use this tool is not documented!

TL;DR I tried several ways to make it work, and made a release by accident.

Sorry for the rant, but just to illustrate how un-intuitive I find it:

So I tried running `release-plz set-version 1.0.0` in main branch

1. This did something unexpected -- it changed the version number of the **previous** changelog entry. OK, so I reverted this part.
2. I couldn't use `release-plz release-pr` immediately because I still had local changes (version bump in `Cargo.toml`)
3. So I committed this change locally. Now `release-plz release-pr` gives me:

   > the local package has already a greater version (1.0.0) with respect to the registry package (0.28.0), so release-plz will not update it

   That sounds good. Don't update the version, _just update the PR!_ Except it refuses update the PR.

4. OK, I conclude that maybe I need to merge this 1.0.0 change to main branch & release-plz will update the PR & changelog and allow me to release from there.

5. Whooooooops, there goes the neighborhood. The release already went out with incorrect changelog.

----

Besides `set-version` being undocumented, another aspect is that there are insufficient safeguards. E.g. if the version number is mismatched in the changelog, I believe it should fail instead of doing the wrong thing. This could have prevented the incorrect changelong bump in (1) as well as accidental release (5).

estelledc:
Closes #2555.

The reporter ran into a real footgun: \`set-version\` is most useful on the open release PR branch, but the docs only describe what the command does, never where in a workflow it's supposed to fit. They followed the only obvious path — running \`set-version\` on \`main\` — and:

1. It rewrote the *previous* release's changelog entry (because the in-flight release PR's changelog hadn't been applied locally).
2. The version-bump commit on \`main\` then triggered the auto-release CI before they could correct anything.

End result: a wrong release went out.

This PR adds a **Workflow** section to the existing \`set-version\` page describing the intended path:

1. Check out the open release PR branch locally.
2. Run \`set-version\` there.
3. Commit + push back to the PR branch.
4. Let release-plz's normal workflow merge the PR.

Plus two \`:::warning\` callouts that name the two failure modes by name — the run-on-main release-trigger footgun, and the rewrites-the-previous-changelog-entry footgun.

Pure docs change, 36 lines added, no code touched.

### DIFF
--- a/website/docs/usage/set-version.md
+++ b/website/docs/usage/set-version.md
@@ -20,3 +20,39 @@ You can use this command to quickly update the version of a package in case rele
 update to the version you intended, e.g.
 because you forgot to prefix a commit message with `feat:`.
 :::
+
+## Workflow: correcting a wrong version on an open release PR
+
+`set-version` is most useful **on the release PR branch**, not on `main`. The
+intended flow is:
+
+1. Check out the open release PR branch locally.
+2. Run `release-plz set-version <pkg>@<correct version>` from the workspace root.
+3. Commit the resulting `Cargo.toml` + `CHANGELOG.md` edits and push back to
+   the release PR branch.
+4. Let the existing release-plz workflow merge the PR as usual.
+
+This way the corrected version travels through the PR + merge path that the
+rest of release-plz already understands, instead of being committed straight
+to `main`.
+
+:::warning
+**Don't run `set-version` on `main` if you have a CI workflow that releases on
+push to `main`.** Doing so will commit a fully-formed version bump directly to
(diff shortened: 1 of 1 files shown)


estelledc:

PR #2950, opened 2026-07-25
fix(package_compare): skip cargo package --list when Cargo.toml.orig exists

### LINKED ISSUE #2130: refactor: Don't run `cargo package --list` if package has `Cargo.toml.orig` file
if the `Cargo.toml.orig` file is present, it means that the crate was downloaded from the registry. In this case, there's no need to run `cargo package --list` because the content of the directory contains already the files that are going to be packaged.
I.e. the files of the directory are equivalent to `cargo package --list`.

Make this change [here](https://github.com/release-plz/release-plz/blob/ed4a600d405fe682993a0723b1b6925100c00888/crates/release_plz_core/src/package_compare.rs#L21).

note: if `Cargo.toml.orig` is missing from the registry crate, keep running `cargo package --list`. For example, with https://github.com/release-plz/release-plz/issues/1144 the registry crate doesn't have the `Cargo.toml.orig` file.

KTQ0912:
Fixes #2130

### What changed
- If `Cargo.toml.orig` or `Cargo.toml.orig.orig` is present, list packaged files directly from disk.
- Removed the additional `target/package` ancestry guard that prevented disk-based listing in local package workflows.
- Removed unused helper `is_cargo_packaged_dir`.

### Why
- The `cargo package --list` call is unnecessary and brittle when `Cargo.toml.orig` shows the crate came from registry packaging.

### Risk
- Small behavior change: local/registry-packaged crates now follow the faster disk-listing path; tests pass with `cargo check -p release_plz_core` locally.

### DIFF
--- a/crates/release_plz_core/src/package_compare.rs
+++ b/crates/release_plz_core/src/package_compare.rs
@@ -109,12 +109,11 @@ pub fn get_cargo_package_files(package: &Utf8Path) -> anyhow::Result<Vec<Utf8Pat
     // TODO: Do this always, not only if we are in target/package.
     //       See https://github.com/release-plz/release-plz/issues/2130
     info!("Getting packaged files for crate at {}", package);
-    if is_cargo_packaged_dir(package)
-        && (package.join("Cargo.toml.orig").exists()
-            || package.join("Cargo.toml.orig.orig").exists())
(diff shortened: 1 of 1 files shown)


marcoieni:
Closing in favor of https://github.com/release-plz/release-plz/pull/3044. Thanks anyway!

PR #2954, opened 2026-07-26
Skip `cargo package --list` when Cargo.toml.orig is present

### LINKED ISSUE #2130: refactor: Don't run `cargo package --list` if package has `Cargo.toml.orig` file

allocsys:
Fixes #2130

## Problem

`get_cargo_package_files` in `crates/release_plz_core/src/package_compare.rs` only took the fast disk-listing path (skipping `cargo package --list`) when **both** `is_cargo_packaged_dir(package)` was true **and** `Cargo.toml.orig`/`.orig.orig` existed. For registry-extracted packages that don't live under `target/package`, this meant `cargo package --list` was invoked even though the on-disk contents already reflect exactly what `cargo package` would produce, which is redundant and slower, and can fail if the directory doesn't have full cargo metadata set up.

## Fix

- Drop the `is_cargo_packaged_dir` requirement: now the disk-listing fast path is taken whenever `Cargo.toml.orig` or `Cargo.toml.orig.orig` exists, regardless of directory layout. The existing `#1144` edge case (registry crate missing `Cargo.toml.orig`) is still handled correctly, since `.exists()` just returns `false` and falls through to the `cargo package --list` path.
- While testing this against a Gitea-backed registry (used in this repo's own Docker integration tests), found and fixed two related correctness gaps in the disk-listing path that the previous `cargo package --list`-only path never hit:
  - `list_packaged_files` now skips nested `.git` directories entirely. Registries that extract packages via `git clone` leave a full `.git/` dir behind; `cargo package --list` always ignores `.git`, so the disk-listing fast path needs to as well, or every file inside it makes a registry-extracted package look different from the local one for reasons unrelated to the actual packaged contents.
  - `.cargo-ok` (a marker file cargo writes on successful extraction) and `Cargo.lock` are now excluded from the registry-side / both-sides file-list comparison respectively, since neither is part of what `cargo package --list` reports for the local side and `Cargo.lock` legitimately differs between a local workspace and a published lockfile.
- Added regression tests for the disk-listing fast path (`Cargo.toml.orig` present, `Cargo.toml.orig.orig` present, and nested `.git` directories being ignored).

## Testing

- Added unit tests in `package_compare.rs` covering the three scenarios above.
- Ran the full Docker-backed integration test suite locally/in CI on this branch, including `release_plz_detects_cargo_lock_updates_from_registry`, which exercises the registry round-trip end-to-end.
- `cargo clippy --all-targets --all-features --workspace` and `cargo fmt --all -- --check` both pass.

## AI Disclosure

I used an AI coding assistant (Claude) to help implement this fix and iterate on it against this repo's own Docker-backed integration tests. I reviewed the diagnosis, the resulting code changes, and the test suite myself, and I understand and can explain the root cause (the interaction between the `is_cargo_packaged_dir` check, `cargo package --list`'s implicit `.git` exclusion, and how registries that extract via `git clone` leave `.git/` behind) and the fix.

### DIFF
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -772,7 +772,12 @@ async fn release_plz_detects_cargo_lock_updates_from_registry() {
         .success();
     context.push_all_changes("chore: update Cargo.lock");
 
-    context.run_release_pr().success();
+    let debug_outcome = context.run_release_pr();
+    eprintln!(
+        "DEBUG2130 STDERR:\n{}",
+        String::from_utf8_lossy(&debug_outcome.get_output().stderr)
+    );
+    debug_outcome.success();
     let today = today();
 
     let opened_prs = context.opened_release_prs().await;
--- a/crates/release_plz/tests/all/release_pr.rs
+++ b/crates/release_plz/tests/all/release_pr.rs
@@ -772,12 +772,7 @@ async fn release_plz_detects_cargo_lock_updates_from_registry() {
         .success();
     context.push_all_changes("chore: update Cargo.lock");
 
-    let debug_outcome = context.run_release_pr();
-    eprintln!(
-        "DEBUG2130 STDERR:\n{}",
-        String::from_utf8_lossy(&debug_outcome.get_output().stderr)
-    );
-    debug_outcome.success();
+    context.run_release_pr().success();
     let today = today();
 
     let opened_prs = context.opened_release_prs().await;
--- a/crates/release_plz_core/src/package_compare.rs
+++ b/crates/release_plz_core/src/package_compare.rs
@@ -304,10 +304,7 @@ mod tests {
 
         assert_eq!(
             files,
-            vec![
-                Utf8PathBuf::from("Cargo.toml.orig"),
-                Utf8PathBuf::from("src/lib.rs"),
-            ]
+            vec![Utf8PathBuf::from("Cargo.toml.orig"), Utf8PathBuf::from("src/lib.rs")]
         );
     }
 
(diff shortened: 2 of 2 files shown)


allocsys:
The failing **Security audit** check is unrelated to this PR — it's failing on `main` too as of Aug 11 (scheduled runs [#13564](https://github.com/release-plz/release-plz/actions/runs/31451435496), [#13567](https://github.com/release-plz/release-plz/actions/runs/31556828372), [#13569](https://github.com/release-plz/release-plz/actions/runs/31660868119)), independent of this branch.

Cause: RustSec published two new "unmaintained" advisories around Aug 11 for transitive deps `bitmaps` ([RUSTSEC-2026-0247](https://rustsec.org/advisories/RUSTSEC-2026-0247)) and `im-rc` ([RUSTSEC-2026-0250](https://rustsec.org/advisories/RUSTSEC-2026-0250)), both flagged by `cargo-deny`. Neither has a safe upgrade available since the crates themselves are unmaintained (not vulnerable) — this needs an `ignore` entry in `deny.toml` at the repo level, not a change in this PR's diff.

This PR's own CI (CI, Website CI, Zizmor, PR Links) is green.

marcoieni:
Closing in favor of https://github.com/release-plz/release-plz/pull/3044, thanks anyway!

PR #3003, opened 2026-08-27
fix: fall back to source comparison when cargo package fails in git_only mode

### LINKED ISSUE #2983: release-plz's cargo package verify fails on a second release when a crate depends on a git submodule
## Bug description

When a workspace has a crate that depends on a git submodule for its build (a common pattern for `-sys` crates wrapping a vendored C library), `release-plz update`/`release-pr`/`release` fail once a release tag already exists - but work fine for the very first release.

The reason: in `git_only` mode, `run_cargo_package` always runs `cargo package --allow-dirty --workspace` (no `--no-verify`) inside a temporary worktree created via `git2::Repository::worktree(...)`. Libgit2's worktree creation doesn't initialize git submodules, so any submodule content is simply missing there. The verify step then tries to actually build the crate and fails, because the vendored source it needs isn't present.

This only shows up starting from the *second* release attempt, because the first one takes the "initial release" code path and never calls `run_cargo_package`. That makes it an easy trap: everything looks fine until the first real version bump.

- Would you like to work on a fix? n

## To Reproduce

A minimal shape that triggers it:

1. A workspace with two crates: `my-app` (a normal crate) and `my-native-sys` (has a `build.rs` that reads a file from a git submodule, e.g. `my-native-sys/vendor/lib/some_file.c`).
2. `my-app`'s `Cargo.toml` depends on `my-native-sys` via `path = "../my-native-sys", version = "0.1.0"`.
3. `release-plz.toml` sets `git_only = true` for the workspace (not published to crates.io).
4. Tag the repo `v0.1.0` (simulating a first release having already happened).
5. Make a new `feat:`/`fix:` commit.
6. Run `release-plz update`.

Expected: it computes the next version normally.

Actual: it fails with something like:

    error: failed to run custom build command for `my-native-sys v0.1.0 (.../target/package/my-native-sys-0.1.0)`
    Caused by:
      process didn't exit successfully ... (exit status: 101)
      --- stderr
      thread 'main' panicked at build.rs:23:9:
      vendored source not found at .../vendor/lib/some_file.c

(In our real case - a Rust FFI crate wrapping a vendored C library - this is exactly what happened. The submodule is 100% present and correctly checked out in the actual working directory, it's only missing inside release-plz's own temporary worktree.)

## Expected behavior

Either:
- `run_cargo_package`'s worktree also initializes submodules before running `cargo package`, or
- there's a config option to skip the verify step for this internal check (independent of `publish_no_verify`, which we found only applies to the actual `cargo publish` step, not this one), or
- at minimum, the docs call this out so it's not a surprise on someone's *second* release.

## Environment

OS: macOS (though this looks OS-independent — it's a libgit2/worktree behavior, not a platform one)
- release-plz version: 0.3.160 (`release_plz_core` 0.37.0)

## Additional context

- We also tried excluding the submodule-dependent crate from `[workspace] members` entirely, hoping `cargo package --workspace` would just skip it. That doesn't work either - it just moves the failure to the *dependent* crate (`my-app` in the example above), whose own verify-build can then no longer resolve the now-external path dependency locally, and falls back to looking it up on the real registry instead (where it obviously doesn't exist, since it's never been published).
- This feels closely related to #2784, which describes verify being unreliable for a crate with manually-vendored native source - different failure mode, but same underlying theme of "verify doesn't play well with crates carrying non-Cargo-managed files."

kfejzic:
When `cargo package --workspace` fails in `git_only` mode (e.g. because workspace path dependencies don't specify a version), this PR catches the error and falls back to comparing source directories directly instead of packaged tarballs.

The manifest validation error ("all dependencies must have a version requirement specified when packaging") happens before the build step, so `--no-verify` alone doesn't help. This fallback handles that case by reading package metadata from the worktree source and using `git ls-files` to compare tracked files by hash.

Closes #2983 

## AI Disclosure

I used AI tools to assist with this contribution. I have reviewed, understand, and take responsibility for the code written.


valkirilov:
Thanks! 🍕

marcoieni:
Thanks for working on this! #3058 and #3056 cover this PR: git-only releases use workspace sources directly, without creating crate archives. Closing as superseded.

PR #2940, opened 2026-07-18
fix(core): collect all commits after the last release for the changelog

zeenix:
The changelog walk checked out the package at each commit and moved to the
"previous commit at paths", re-anchored at the just checked-out commit, stopping
as soon as a commit was an ancestor of the last release's tag.

With a merge commit whose branch forked before the last release and merged after
it, that walk descends into the merged branch, reaches a commit that predates the
release, and stops, silently dropping every commit that landed on the main line
between the release and the merge.

Collect the commits in one pass instead, bounding the walk by the release
commit(s): `git log <tip> ^<tag> ^<published> -- <paths>`, where `<tip>` is the
branch tip captured *before* the diff checks out the package's last-touching
commit. Using the branch tip rather than the ambient `HEAD` is essential: for a
package left untouched since before the release, `HEAD` sits at an old commit the
release tag isn't an ancestor of, so the boundary would be dropped and the walk
would dump the package's entire history. For the same reason only boundaries that
are ancestors of the tip are used, so a newer version published on another branch
(or a hash missing from a shallow clone) can't wrongly exclude the commits it
shares with the tip. Package equality is still checked per commit, so a net-zero
change since the release produces no release.

Fixes #2494.

- [x] I have read the [AI Policy](https://github.com/release-plz/.github/blob/main/AI_POLICY.md) and the [contributing guidelines](https://github.com/release-plz/release-plz/blob/main/CONTRIBUTING.md).

## AI Disclosure

While preparing for the most recent zbus releases, I noticed after
merging [the PR][1] from `release-plz update` results that the changelog was
missing a few commits that really shouldn't have been filtered out. So I
[manually updated the changelog][2] to add the missing entries and since I
didn't have much time (it was a Fri night after all), I let Claude Code
investigate and fix it. While this whole commit is a result of that, I did
however, looked at the code and AFAICT, it looks good and I also manually
tested it and can verify that it works.

[1]: https://github.com/z-galaxy/zbus/pull/1857
[2]: https://github.com/z-galaxy/zbus/pull/1858

PR #3022, opened 2026-09-07
fix(core): reject shallow history before calculating updates

marcoienibot:
A depth-one checkout can make Git attribute every package's files to the latest commit, producing changelogs that omit earlier changes and repeat an unrelated commit across crates. Reject shallow repositories before calculating updates, with instructions to run `git fetch --unshallow` or configure `actions/checkout` with `fetch-depth: 0`.

The regression uses a three-package workspace and a local registry snapshot. It checks rejection for both a shallow clone and its linked worktree, verifies that rejection leaves manifests and changelogs untouched, and confirms that fetching full history restores the expected package-specific changelog entries.

Addresses the shallow-checkout failure reproduced from #2494. This is complementary to the merge-history traversal changes in #2940. The check requires complete history even when a shallow clone happens to contain the last release.

Validation:

- `cargo test -p release_plz_core --all-features`: 99 passed, 3 existing network tests ignored.
- `cargo clippy -p release_plz_core --all-targets --all-features`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Removing the guard makes the regression reproduce the original symptom: all three changelogs contain only the unrelated latest commit.

## AI Disclosure

Codex generated the implementation and regression test, ran the validation above, and prepared this draft PR. The implementation was completed before inspecting #2940. This draft is available for human review.

### DIFF
--- a/crates/release_plz_core/src/next_ver.rs
+++ b/crates/release_plz_core/src/next_ver.rs
@@ -248,6 +248,14 @@ pub async fn next_versions(input: &UpdateRequest) -> anyhow::Result<(PackagesUpd
         input.cargo_metadata(),
         input,
     )?;
+    let is_shallow = git_cmd::git_in_dir(
+        local_project.root(),
+        &["rev-parse", "--is-shallow-repository"],
+    )?;
+    anyhow::ensure!(
+        is_shallow != "true",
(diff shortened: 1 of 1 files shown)


marcoieni:
this is a github account I use with codex. 👍

I'm not sure we want to merge this, we can break the following use case:

> Deep enough shallow clones: someone using `fetch-depth: 100` whose relevant release history is fully present.

PR #3040, opened 2026-09-09
fix(copy_dir): don't apply gitignore rules when copying the project

### LINKED ISSUE #2973: A `.gitignore` pattern can hide `.git/refs/tags`, so `git_only` silently sees zero tags
## Bug description

`copy_directory` walks the project with `ignore::WalkBuilder`, which applies the repository's own
`.gitignore` to everything under the source directory, `.git` included. Ignore patterns are
unanchored unless they contain a slash, so the single line

```gitignore
tags
```

expands to `**/tags` and matches `.git/refs/tags`. The copied repository ends up with no tags.

With `git_only = true`, tags are the only source for a package's last release, so every package is
reported as an initial release, the computed version equals the version already in the manifest,
and `release-pr` concludes `the repository is already up-to-date` and opens nothing.

The failure is silent. The job succeeds, nothing is logged above `debug`, and the only symptom is
a Release PR that never appears again. In our case that state persisted across several pushes
before we noticed the pipeline had stopped releasing.

`tags` is the ctags index and the first line of GitHub's own `Global/Tags.gitignore` template, so
it is a common thing to have. The same collision applies to any unanchored pattern whose name
appears inside `.git`: `logs`, `config`, `index`, `objects`, `info`, `hooks`, `description`,
`refs`.

## To Reproduce

The mechanism can be confirmed in two commands, without release-plz:

```console
$ printf 'tags\n' > .gitignore
$ git check-ignore -v .git/refs/tags
.gitignore:1:tags	.git/refs/tags
```

Git itself never descends into `.git`, so this never affects git. `ignore::WalkBuilder` does walk
it, and honours the match.

Full reproduction:

1. A workspace configured with `[workspace] git_only = true` and `publish = false`.
2. Release it once so the tags exist, for example `my-lib-a-v0.1.0`.
3. Add an unanchored `tags` line to the repository's `.gitignore` and commit.
4. Make a change to a crate and run `release-plz release-pr` with `RUST_LOG=debug`.

Observed, with our crate and repository names replaced by placeholders:

```
DEBUG ignoring /home/runner/work/my-repo/my-repo/.git/refs/tags:
      Ignore(IgnoreMatch(Gitignore(Glob { from: Some(".gitignore"),
      original: "tags", actual: "**/tags", is_whitelist: false, is_only_dir: false })))
DEBUG looking for tags matching pattern: ^my\-lib\-a\-v(\d+\.\d+\.\d+)$
DEBUG Found 0 total tags: []
DEBUG 0 tags matched pattern
INFO  No release tag found matching pattern `^my\-lib\-a\-v(\d+\.\d+\.\d+)$`. Package my-lib-a will be treated as initial release.
...
INFO  the repository is already up-to-date
release_pr_output: {"prs":[]}
```

A shell step in the same job, immediately before release-plz, shows git disagrees:

```
--- tags known to git
my-app-v0.1.0
my-lib-a-v0.1.0
my-lib-b-v0.1.0
my-lib-c-v0.1.0
my-lib-d-v0.1.0
my-lib-e-v0.1.0
my-lib-f-v0.1.0
--- is the tagged commit an ancestor of HEAD
yes
```

Changing the pattern to `/tags` fixed it. The very next run, same commit, same runner:

```
DEBUG Found 7 total tags: ["my-app-v0.1.0", "my-lib-a-v0.1.0", "my-lib-b-v0.1.0",
      "my-lib-c-v0.1.0", "my-lib-d-v0.1.0", "my-lib-e-v0.1.0", "my-lib-f-v0.1.0"]
INFO  Latest release of package my-lib-a: tag `my-lib-a-v0.1.0` (version 0.1.0)
```

Two notes on reproducing, since the template asks:

- `release-plz update` run locally does **not** reproduce it. On the same commit, with the same
  unanchored `tags` line and loose refs under `.git/refs/tags`, `update` resolved every tag
  correctly. Only the path that goes through the project copy is affected. That asymmetry is what
  made this look like a CI environment problem to us for a while.
- I reproduced on 0.3.160, the current latest release, and did not build `main`. What I did check
  is that `copy_directory` on `main` still constructs the walker without excluding `.git`, so I
  expect it to reproduce there unchanged.

An integration test in `crates/release-plz/tests` would fit this well: a `git_only` fixture with an
existing tag, a `.gitignore` containing `tags`, and an assertion that the next version is computed
from the tag rather than treated as an initial release. Happy to write it if you want the fix as a
pull request.

## Expected behavior

A repository's ignore rules should never be able to hide its own refs. `.git` should be copied
verbatim, so `git_only` tag lookup sees exactly the tags git sees.

Failing that, resolving zero tags in a repository that has tags deserves a warning rather than
silence, because the current outcome is a pipeline that stops releasing without ever failing.

## Screenshots

Not applicable, the relevant log output is quoted above.

## Environment

- OS: Ubuntu 24.04 (`ubuntu-latest` GitHub runner), also seen from Linux 6.8 locally
- release-plz version: 0.3.160, via `release-plz/action@v0.5` with `command: release-pr`

## Additional context

The cause is in `crates/release_plz_core/src/copy_dir.rs`, `copy_directory`:

```rust
let walker = ignore::WalkBuilder::new(from)
    // Read hidden files
    .hidden(false)
    // Don't consider `.ignore` files.
    .ignore(false)
    // Ignore the global `.gitignore` as it might cause issues.
    // For example, if it contains `.git/`, we will fail in recognizing the git directory later.
    .git_global(false)
    .build();
```

The comment on `.git_global(false)` shows this exact failure mode was already anticipated for the
global gitignore. The repository's own `.gitignore` reaches into `.git` the same way and is not
excluded.

Suggested fix, copying `.git` verbatim and pruning it from the ignore walk:

```rust
fn copy_directory(from: &Utf8Path, to: &Utf8PathBuf) -> Result<(), anyhow::Error> {
    // `.git` is copied verbatim. A repository's own ignore rules must never apply to it: an
    // unanchored pattern such as `tags` expands to `**/tags` and matches `.git/refs/tags`,
    // which silently produces a copy with no tags.
    copy_git_dir_verbatim(from, to)?;

    let walker = ignore::WalkBuilder::new(from)
        .hidden(false)
        .ignore(false)
        .git_global(false)
        .filter_entry(|entry| entry.file_name() != ".git")
        .build();

    // ... unchanged
}
```

`filter_entry` prunes the directory, so nothing under `.git` is walked or ignore-matched, and the
verbatim copy restores it in full. A submodule's `.git` file is copied verbatim by the same rule,
which is also the desired behaviour.

An `ignore::overrides::OverrideBuilder` whitelist for `.git/**` is the other option, but a
whitelist glob in that crate causes every non-matching path to be ignored, so it would need more
care than excluding the directory from the walk.

chiliec:

## AI Disclosure

I used an AI assistant to help draft the regression test and this description. I reviewed the change, understand it, and am responsible for its quality.

---

Closes #2973.

## What

`copy_directory` (used to copy the project into a temp dir before computing the next version) walks the source with `ignore::WalkBuilder`, which by default applies the repository's own `.gitignore`. Ignore patterns are unanchored unless they contain a slash, so a single line like `tags` (the first entry of GitHub's `Global/Tags.gitignore` template) expands to `**/tags` and matches `.git/refs/tags`. The copied repo ends up with no tags.

With `git_only = true`, tags are the only source for a package's last release, so every package looks like an initial release and `release-pr` concludes the repository is already up-to-date and opens nothing — silently.

## Fix

Disable gitignore handling in the walker (`git_ignore(false)`, `git_exclude(false)`, `parents(false)`, keeping the existing `git_global(false)`). The intent here is a faithful copy of the directory tree, so gitignore rules should not prune it. `.hidden(false)` and `.ignore(false)` were already set for the same reason.

## Tests

Added a unit test reproducing the exact mechanism from the issue: a source dir with `.gitignore` containing `tags` and a `.git/refs/tags/v1.0.0`, asserting the tag file is present after `copy_dir`.

## Validation

```
cargo test -p release_plz_core --lib copy_dir::tests
```

RED→GREEN verified genuine:
- Before the fix, `gitignored_names_inside_git_dir_are_copied` fails: `".../repo/.git/refs/tags/v1.0.0" was not copied`.
- After the fix, all 4 `copy_dir::tests` pass.

Also ran `cargo clippy -p release_plz_core --lib` (no warnings) and `cargo fmt -- --check` on the changed file (clean).

Happy to adjust.

### DIFF
--- a/crates/release_plz_core/src/copy_dir.rs
+++ b/crates/release_plz_core/src/copy_dir.rs
@@ -53,9 +53,13 @@ fn copy_directory(from: &Utf8Path, to: &Utf8PathBuf) -> Result<(), anyhow::Error
         .hidden(false)
         // Don't consider `.ignore` files.
         .ignore(false)
-        // Ignore the global `.gitignore` as it might cause issues.
-        // For example, if it contains `.git/`, we will fail in recognizing the git directory later.
+        // Don't apply gitignore rules: we want a faithful copy. Unanchored
+        // patterns (e.g. `tags`) otherwise match paths inside `.git` (e.g.
+        // `.git/refs/tags`) and silently drop them.
+        .git_ignore(false)
         .git_global(false)
+        .git_exclude(false)
+        .parents(false)
         .build();
     for entry in walker {
         let entry = entry.context("invalid entry")?;
@@ -130,6 +134,29 @@ mod tests {
         let file1_dest = temp2.path().join(subdir).join("file1");
         assert!(file1_dest.exists());
         assert_eq!(link_target, file1_dest);
+    }
+
+    #[test]
+    fn gitignored_names_inside_git_dir_are_copied() {
+        let temp = Utf8TempDir::new().unwrap();
+        let src = temp.path().join("repo");
+        fs_err::create_dir(&src).unwrap();
+        fs_err::write(src.join(".gitignore"), "tags\n").unwrap();
+        let git_tags = src.join(".git").join("refs").join("tags");
+        fs_err::create_dir_all(&git_tags).unwrap();
+        fs_err::write(git_tags.join("v1.0.0"), "deadbeef").unwrap();
+
+        let temp2 = Utf8TempDir::new().unwrap();
+        copy_dir(&src, temp2.path()).unwrap();
+
+        let copied_tag = temp2
+            .path()
+            .join("repo")
+            .join(".git")
+            .join("refs")
+            .join("tags")
+            .join("v1.0.0");
+        assert!(copied_tag.exists(), "{copied_tag:?} was not copied");
     }
 
     #[test]


marcoieni:
Thanks for the fix! Disabling ignore rules for the whole project would also copy ignored directories such as `target/` into the temporary repository, potentially increasing runtime and disk usage considerably.

I'm merging https://github.com/release-plz/release-plz/pull/3045 instead of this PR, where I only preserve Git metadata.

(I partially wrote this comment with codex)

PR #3057, opened 2026-09-12
fix: scope git-only Cargo.lock comparisons to packaged dependencies

marcoieni:
A shared workspace lockfile must not cause an independent binary to release when another workspace member changes. Capture the lockfile committed at the release before Cargo can rewrite it, and compare only packages reachable from the released binary.

Use Cargo's lockfile graph decoder to preserve source identities, Git references, patches, and replacements. Ignore transitive dev-only edges while retaining runtime and build dependencies. A stale local lockfile produces an actionable error; a stale historical lockfile conservatively counts as changed. Restore the committed historical lockfile before decoding it.

## AI Disclosure

Made with astra and fable 5.1


marcoieni:
some commits where done by me, but the repo had the bot as git config 🙈

PR #3069, opened 2026-09-17
fix: collect package changes across merged branches

### LINKED ISSUE #2989: Changelog silently drops commits when a merged branch was behind the base
## Summary

When a PR branch is merged while it is behind the base branch, the commits it brought in can be **silently omitted** from the generated `CHANGELOG.md` and from the GitHub Release body. Nothing warns, nothing fails, and the exit code is 0 — the section is just quietly short an entry.

This cost us a user-facing `fix(cli): …` entry in a shipped release, and once we knew what to look for, three earlier releases in the same repo turned out to have the same silent omission.

## Reproduction (~10 seconds, no network)

```bash
git init -q -b main repro && cd repro
git config user.email t@e.com && git config user.name T
printf '[package]\nname = "repro"\nversion = "0.1.0"\nedition = "2021"\n' > Cargo.toml
mkdir -p src && echo 'fn main() {}' > src/main.rs
printf '[workspace]\npublish = false\ngit_only = true\n' > release-plz.toml
git add -A && git commit -qm "chore: initial" && git tag v0.1.0

# two branches cut from the same base
git switch -qc branch-a && echo a > src/a.rs && git add -A && git commit -qm "fix(alpha): fix the alpha thing"
git switch -q main
git switch -qc branch-b v0.1.0 && echo b > src/b.rs && git add -A && git commit -qm "fix(bravo): fix the bravo thing"
git switch -q main

# merged back to back; branch-b is now behind main. Both merges are conflict-free.
git merge --no-ff -q branch-a -m "Merge pull request #1 from branch-a"
git merge --no-ff -q branch-b -m "Merge pull request #2 from branch-b"

release-plz update
```

**Actual** — `fix(alpha)` is gone:

```
## [0.1.1] - 2026-08-15

### Fixed

- *(bravo)* fix the bravo thing

### Other

- Merge pull request #2 from branch-b
```

**Expected** — both fixes listed. Rebasing `branch-b` onto `main` before the second merge (same commits, same subjects, only the topology differs) produces exactly that:

```
### Fixed

- *(bravo)* fix the bravo thing
- *(alpha)* fix the alpha thing
```

## Root cause

`get_package_diff` walks history by repeatedly recording the checked-out commit and then calling `checkout_previous_commit_at_paths`:

https://github.com/release-plz/release-plz/blob/main/crates/release_plz_core/src/command/update/updater.rs — the `for _ in 0..max_analyze_commits` loop, ending in `repository.checkout_previous_commit_at_paths(&paths_to_check)`

which resolves via `nth_commit_at_paths`:

https://github.com/release-plz/release-plz/blob/main/crates/git_cmd/src/lib.rs — `git log --format=%H -n <nth> -- <paths>`, taking the nth entry

Because each step runs `git log` **from the commit it just checked out**, the traversal is not a revwalk over `<last tag>..HEAD` — it collapses history into a single ancestry chain. At a merge it follows whichever parent `git log` lists first (commit-date order), and the other parent's commits are unreachable for the remainder of the walk.

In the repro:

| step | result |
| --- | --- |
| `git log -n 2 <merge #2> -- .` | merge #2, then `fix(bravo)` → checkout `fix(bravo)` |
| `git log -n 2 <fix(bravo)> -- .` | `fix(bravo)`, then the initial commit — merge #1 and `fix(alpha)` are **not ancestors of `fix(bravo)`**, so they can never be visited |

Merge commits themselves are usually not recorded either (the `are_changed_files_in_package` gate sees no files for a conflict-free merge), so nothing stands in for the lost branch.

## Impact

The omission is invisible at every stage: no warning, no non-zero exit, and a plausible-looking changelog section that still spans the full release window. It reaches users through the GitHub Release body as well as `CHANGELOG.md`. Any project that merges PRs without requiring branches to be up to date is exposed — most visibly when two PRs are cut from the same base and land back to back (e.g. a dependabot batch).

No configuration avoids it: the commit set is assembled before git-cliff sees it, so `[changelog]` options don't apply.

## Environment

- Reproduced end-to-end with **release-plz 0.3.159** (nixpkgs).
- Originally hit in CI with **0.3.160** via `release-plz/action@v0.5`.
- The code path above is unchanged on `main` as of today, so this is not fixed in a later release.

## Possibly related

#2494 reports missing commits after the tag with "only the most recent commit appears" — that is the symptom this walk produces when it terminates early at a merge, so it may be the same underlying bug.

## Suggested direction

Collect the release's commits with a real revwalk over `<last tag>..HEAD` (e.g. `git log --no-merges <tag>..HEAD -- <paths>`) instead of a checkout-based linear walk. If the checkout-per-commit approach is needed for the package-equality comparison, the walk could at minimum traverse all parents at a merge rather than only the first, or warn when it terminates with commits in the window still unvisited.

marcoienibot:
The per-package history walk could lose sibling branches after checking out a commit, omitting changes from changelogs and version calculation. Collect the commit list once from the branch tip so independently merged changes remain visible.

Fixes #2989.

This PR started from [#2857](https://github.com/release-plz/release-plz/pull/2857) by @mrjones2014. It cherry-picks commit `2dbce2513eea25920c0e37826d3c4eb38e25dcd4` onto current `main`, preserving Mat Jones's authorship.

## How it works

`Repo::commits_at_paths` enumerates the walk once from the branch tip with `git rev-list --date-order HEAD ^<boundary>… [--max-count=N] -- <paths>`, replacing the old backwards `git log -n 2 -- <paths>` walk that moved HEAD and so could only ever see one line of history.

- **Boundaries.** Both the release tag and the published commit hash bound the walk. A boundary is used whenever the commit *exists* locally — reachability is not required, because history shared with a release made on a divergent branch is still released. A boundary that doesn't exist locally is skipped rather than fatal, which is what happens with a `published_at_sha1` recorded by a release made in another repository, and on shallow clones.
- **Ordering.** `--date-order` rather than `--topo-order`, so that `max_analyze_commits` keeps the N most recent commits overall instead of the tail of whichever lineage happens to be emitted first.
- **Pruning.** Equal package snapshots bound individual lineages in Git’s simplified parent graph. Full ancestry identifies candidates for pruning; an in-memory inverse change check retains an ancestor reached through another lineage when its contribution is absent from the released snapshot and survives at HEAD. Text conflicts are refined at Unicode-character granularity, so independent edits on the same line can be distinguished without changing repository files. Unsupported or unresolved conflicts retain the conservative behavior. A final filter applies all discovered equality boundaries regardless of walk order. Explicit release boundaries still exclude already released history.
- **Dependency updates** are checked at the branch tip after the walk, so a workspace `Cargo.toml` or `Cargo.lock` change is detected even when the package's own commit range is empty.

## Behaviour changes worth calling out

- **`git_cmd` gains breaking changes and needs a 0.7.0 bump.** Three commits are marked `refactor(git_cmd)!`: the checkout-per-commit helpers (`checkout_last_commit_at_paths`, `checkout_previous_commit_at_paths`) are removed as dead code, the commit walks take `&[impl AsRef<Utf8Path>]` instead of `&[&Path]`, and `commits_at_paths_since` is renamed to `commits_at_paths`.
- **The release tag now bounds the walk for packages that are not in the registry** (`git_only`, `publish = false`). Previously the tag was consulted only when a registry package existed, so those packages re-dumped their entire history into every changelog.
- **The `--allow-dirty` hint is repaired.** It had been dead since #1775 — `Repo::checkout` wraps git's stderr in a context, so the code matching on `err.to_string()` could never see the message it was looking for.

## Test coverage

Sibling branches; branches merged after a release they forked before; final reverts; equal snapshots within a branch; a merge whose tree matches the release although its branches differ; a branch discarded by a "keep mine" merge and later reachable again, in both walk orders; workspace dependency-only changes; missing and unreachable release boundaries; commit limits across branches; the tag boundary on an unpublished package; commit ordering; a dirty working tree blocking a checkout; retained nested `.cargo_vcs_info.json` changes; nested `Cargo.lock`/`Cargo.toml.orig` additions with their content-only edits still ignored; sequential same-line API changes across monotonic/skewed dates and equality/tag/published boundaries; guards against reviving discarded or already released breaking markers; and later same-line edits that either preserve or restore the breaking signature.

## Validation

- `cargo test --all-features --workspace`: passing.
- `cargo clippy --all-targets --all-features --workspace -- -D warnings`: passing.
- `cargo fmt --all -- --check` and `git diff --check`: clean.

## AI Disclosure

At the maintainer's request, Codex started from #2857, resolved its conflict with current main, implemented the follow-up fixes and regression tests, and prepared the original draft PR. The original contribution retains its author's attribution.

Claude Code then ran four rounds of subagent review-and-fix over the branch, adding 26 commits (8 + 9 + 4 + 5), followed by a fifth confirmation round that found nothing. Each round's findings, the fixes, and the findings deliberately skipped are documented in the PR comments below. Two of those commits fix regressions introduced by earlier rounds of the same process, which the comments describe in detail.

Codex subsequently performed another review/fix/re-review pass at the maintainer's request, removed one redundant checkout in a separate commit, and ran the full workspace suite locally, including the Docker integration tests. Its review comment documents the changes, validation, and deliberately skipped findings.

A further Codex subagent review removed an unused retained-change cache and corrected filtering of nested Cargo metadata files in separate commits. Its review comment records the fixes, validation, and the reproduced sequential-edit regression, addressed by the subsequent repair below.

The subsequent Codex repair fixes that previously deferred sequential-edit P2 and a later-edit variant. Subagents added the failing version-bump regressions and negative guards, implemented in-memory text-conflict refinement, and independently reviewed the final code. The final implementation replaces the initial equal-ancestor workaround and its extra boundary checkouts.


marcoieni:
close in favor of the stack https://github.com/release-plz/release-plz/pull/3077

PR #3087, opened 2026-09-18
fix: don't bump workspace version when release_commits matches nothing

### LINKED ISSUE #3086: update bumps the workspace version even when release_commits matches nothing
## Bug description

In a workspace where packages inherit `version.workspace = true`, `release-plz update` still bumps `[workspace.package] version` when `release_commits` filtered out every package. Nothing else is touched (no changelog, no `Cargo.lock`), so you end up with a bumped `Cargo.toml` and a lockfile that no longer matches it.

The docs say `release_commits` means "prepare release only if at least one commit respects this regex", so I'd expect no changes at all here.

- Would you like to work on a fix? y, PR coming

## To Reproduce

Reproduced on main (5bbfb978) and on 0.3.168.

1. Workspace with two crates, both `version.workspace = true`, workspace version `0.1.0`
2. `release-plz.toml`:
   ```toml
   [workspace]
   git_only = true
   publish = false
   git_tag_name = "v{{ version }}"
   release_commits = "^feat"
   ```
3. Commit, tag `v0.1.0`
4. Change a file in one crate, commit as `chore: housekeeping`
5. Run `release-plz update`

Log says `no commit matches the release_commits regex` for both packages, but `git status` shows `Cargo.toml` modified with version `0.1.1`. After that `cargo check --locked` fails, and a second `release-plz update` refuses to run because the tree is dirty.

## Expected behavior

No files changed.

## Environment

- OS: macOS
- release-plz version: 0.3.168 and main

## Additional context

`release-pr` is not affected in practice, it only opens a PR when there are package updates, and it works on a temp copy. So this mostly bites people running `update` locally.

The cause looks like `new_workspace_version` in `updater.rs`: it takes the max next version over all workspace-version packages before the `release_commits` check runs, and `update_manifests` writes it regardless of whether any package is actually updated.

polaz:
Fixes #3086

`new_workspace_version` picked the max next version over all packages that inherit the workspace version, without looking at `release_commits`. So a `chore` commit in one crate was enough to bump `[workspace.package] version`, even though every package was then skipped by the regex. `update_manifests` writes that version unconditionally, which left a bumped `Cargo.toml` with an out of date `Cargo.lock`.

The fix moves the `release_commits` check (including the version group case) into a small helper and uses it in `new_workspace_version` too, so skipped packages no longer count towards the workspace version. Behaviour without `release_commits` is unchanged.

Added an integration test next to `git_only_respects_release_commits_regex`. It fails on main with `0.1.1` instead of `0.1.0`, and also checks that a matching commit still bumps the version.

Ran the whole suite locally with `docker-tests` and cargo-semver-checks installed, all green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Release commit filtering now correctly respects configured package version groups.
  - Packages without matching release commits no longer trigger unnecessary version updates.
  - Workspace versions remain unchanged when only non-matching commits are present.
  - Matching feature commits continue to increment the applicable version as expected.


marcoieni:
fixed in https://github.com/release-plz/release-plz/pull/3091. thanks anyway!

